PR #3, opened 2026-07-10
feat(pam): implement rate-limit contract and identity lockout service

klaudia-blazyczek-blurify:
## Summary

Add rate-limit adapter seam to PAM identity service for login brute-force defense. Introduces `RATE_LIMIT_ADAPTER` port; in-memory + Redis implementations; login router enforces lockout with exponential backoff and `lockoutUntil` timestamp in error. Operators can rebind to custom backend (Drizzle, external service) via DI.

## Changes

- **`packages/core/src/contracts/adapters/rate-limit.ts`** — New `RateLimitAdapter` interface with typed options (max attempts, time window, backoff curve)
- **`packages/core/src/pam/identity/*`** — Guard login route with `check()`; compute lockout in service; throw error with `lockoutUntil` timestamp for client countdown
- **`packages/core/src/server/kernel/rate-limiter.ts`** — In-memory backend for dev/test
- **`packages/core/src/server/kernel/redis-rate-limiter.ts`** — Redis backend for production (operator binds in DI)
- **`packages/core/src/pam/identity/__tests__/identity.service.test.ts`** — Unit tests: lockout trigger, expiry, backoff, success clears state

## Acceptance criteria

- [x] Rate-limit adapter interface with typed options (max, windowMs, backoff exponent)
- [x] In-memory + Redis implementations; both tested
- [x] Login router blocks locked-out users before Zod validation
- [x] Failed attempts logged; lockout window computed server-side
- [x] Error response includes `lockoutUntil` ISO timestamp for UI countdown
- [x] Consumer can override default binding via `ctx.provide(RATE_LIMIT_ADAPTER, customImpl)`

## Checklist

- [x] `pnpm verify:drift` is green (no OpenAPI/catalog drift)
- [x] No cross-module direct imports (adapter in `contracts/`, impl in kernel, domain imports only contract)
- [x] No player data, secrets, or env vars in code (Redis config via DI token, not hardcoded)
- [x] AGENTS.md updated for identity module (new port documented)

## Notes

None


agniev-a-hub (dismissed):

packages/core/src/server/kernel/rate-limiter.ts
```diff
@@ -44,6 +44,11 @@ export class InProcessRateLimiter implements RateLimiterAdapter {
     return Promise.resolve({ allowed: false, retryAfterMs: existing.resetAt - now });
   }
 
+  reset(key: string): Promise<void> {
```
zaxovaiko, line 47: AC states "No breaking changes to public PAM contracts or identity routes"; however, you added a new `reset` method to the adapter. Good for now, but be careful when introducing breaking changes for consumers 👍

packages/core/src/server/kernel/rate-limiter.ts
zaxovaiko, line 21: We need to get rid of `InProcessRateLimiter` and rely only on a solution that works consistently across replicas - Redis-backed, so the rate-limit budget is shared cluster-wide instead of per-pod. But let's put it for later.

packages/core/src/contracts/adapters/rate-limit.ts
```diff
@@ -25,6 +25,7 @@ export type RateLimitResult = {
 
 export type RateLimiterAdapter = {
   consume(key: string, opts: RateLimitOptions): Promise<RateLimitResult>;
+  reset(key: string): Promise<void>;
```
zaxovaiko: Also, I think it'd be great to have key as a generic type and create an enum for keys, so that we will know what the framework uses.

packages/core/src/server/kernel/rate-limiter.ts
klaudia-blazyczek-blurify: My bad. Good catch

packages/core/src/server/kernel/rate-limiter.ts
klaudia-blazyczek-blurify: Should I fix it here if I didn't even touch it? I suggest creating separate task for that so that this bug fix can be merged quickly?

agniev-a-hub: Separate task, can take it if needed :)

agniev-a-hub (dismissed):

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -454,6 +455,7 @@ export class IdentityService {
     );
 
     await this.clearLockout(userId);
+    await this.limiter?.reset(`login:${existingUser.email.toLowerCase()}`);
```
zaxovaiko: Let's extract this compound key to a separate function. How could I know later what we expect as a key?

zaxovaiko (approved):

PR #4, opened 2026-07-10
fix(core): respect rememberMe flag for session persistence

klaudia-blazyczek-blurify:
## Summary

Fixed sessions incorrectly persisting after browser close regardless of "Remember me" checkbox state. Sessions now respect the rememberMe flag: unchecked → session-only (cleared on browser close), checked → 30-day persistent.

## Changes

- **core** (`pam/identity` module):
  - `auth.ts`: Removed hardcoded `rememberMeDuration` config that forced all sessions to 30-day persistence
  - `identity.service.ts`: Modified `forwardCookies()` to conditionally set `Max-Age` header based on `rememberMe` flag passed from login handler

## Acceptance criteria

- [x] Without "Remember me" → session-only cookie (no Max-Age), cleared on browser close
- [x] With "Remember me" → persistent cookie (Max-Age=2592000 / 30 days)
- [x] rememberMe flag flows correctly: UI → oRPC → identity service → cookie headers
- [x] No changes needed in BetFeel consumer (already sends rememberMe correctly)

## Checklist

- [x] No cross-module boundary violations
- [x] No new tables
- [x] No secrets or player data added

## Notes

- Manual testing recommended: Check cookie Max-Age in DevTools (F12 → Application → Cookies)
  - Login without "Remember me" → `auth_token` should have NO `Max-Age` / `Expires`
  - Login with "Remember me" → `auth_token` should have `Max-Age: 2592000`
- No breaking changes; purely fixes incorrect behavior


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -80,10 +80,28 @@ function toUser(u: BetterAuthUser) {
...
+function forwardCookies(
+  authResponse: globalThis.Response,
+  resHeaders: Headers,
+  rememberMe?: boolean,
+): void {
   const cookies = authResponse.headers.getSetCookie?.() ?? [];
+  const REMEMBER_ME_DURATION_SECONDS = 30 * 24 * 60 * 60; // 30 days
   for (const cookie of cookies) {
-    resHeaders.append('set-cookie', cookie);
+    const isSessionCookie = cookie.includes('HttpOnly') && cookie.includes('Path=');
```
zaxovaiko: Prefer matching on the actual cookie name prefix [`better-auth.session_token`](https://github.com/better-auth/better-auth/blob/03e0e36a98a21eaf0ed39e384012f3216c954415/packages/better-auth/src/cookies/index.ts#L117) instead of attribute substrings.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -80,10 +80,28 @@ function toUser(u: BetterAuthUser) {
...
 }
 
-function forwardCookies(authResponse: globalThis.Response, resHeaders: Headers): void {
+function forwardCookies(
+  authResponse: globalThis.Response,
+  resHeaders: Headers,
+  rememberMe?: boolean,
+): void {
   const cookies = authResponse.headers.getSetCookie?.() ?? [];
+  const REMEMBER_ME_DURATION_SECONDS = 30 * 24 * 60 * 60; // 30 days
```
zaxovaiko: Infer session duration from `session.expiresIn` instead of hardcoding it here - otherwise this and the better-auth config can drift out of sync

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -250,6 +240,31 @@ export class IdentityService {
     return this.auth.api as unknown as ExtendedAuthApi;
   }
 
+  private forwardCookies(
+    authResponse: globalThis.Response,
+    resHeaders: Headers,
+    rememberMe?: boolean,
+  ): void {
+    const cookies = authResponse.headers.getSetCookie?.() ?? [];
+    const sessionDurationSeconds = this.auth.options.session?.expiresIn ?? 30 * 24 * 60 * 60;
```
agniev-a-hub: Extract calculation of session duration in some CONST_VAR for better exp. You have const value here, but it's closer to config one worth be names as SESSION_DURATION_SECONDS=... and outside of the func body.

packages/core/src/pam/identity/__tests__/identity.service.test.ts
```diff
@@ -13,6 +13,11 @@ const { signInEmailMock, getSessionMock, updateUserMock } = vi.hoisted(() => ({
 vi.mock('@openora/core/server', async (importOriginal) => ({
   ...(await importOriginal<object>()),
   createAuth: vi.fn(() => ({
+    options: {
+      session: {
+        expiresIn: 30 * 24 * 60 * 60,
```
agniev-a-hub: SESSION_DURATION_IN_SECONDS could be also used here, it's not dependancy but config.

agniev-a-hub (dismissed):
Extract session duration and LGTM.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -250,6 +241,32 @@ export class IdentityService {
     return this.auth.api as unknown as ExtendedAuthApi;
   }
 
+  private forwardCookies(
```
zaxovaiko: This whole forwardCookies rewrite duplicates what better-auth already does natively. In the installed 1.6.15, the session cookie's `Max-Age` is `dontRememberMe ? undefined : session.expiresIn` (`better-auth/dist/cookies/index.mjs:126`), driven by the `rememberMe` flag - which we already pass to `signInEmail` at line 337. So when `rememberMe === true` better-auth has already set `Max-Age=expiresIn;` this code strips it and re-appends the identical value. When it's `false/undefined` we pass through - same as the old passthrough. Net effect on the emitted cookie is zero. Suggest reverting `forwardCookies` to the plain passthrough and dropping the string parsing (fragile `split('; ')` + prefix matching on `Set-Cookie`).

zaxovaiko (approved):

PR #12, opened 2026-07-15
feat: error-tracking adapter seam with sentry reference impl

zaxovaiko:
## What

Adds an `ERROR_TRACKING` adapter seam so the runtime can report unhandled errors to a vendor, replaceable via an overlay (PostHog, Rollbar, ...). Ships a Sentry reference impl in core that auto-binds on `SENTRY_DSN`.

- **Contract**: `ErrorTrackingAdapter` port + `ERROR_TRACKING` token (`contracts/adapters/error-tracking.ts`). Fire-and-forget `captureException(error, context?)` - a reporter failure never breaks the calling path.
- **Default + Sentry impl** (`server/kernel/error-tracker.ts`): `NoopErrorTracker` default; `createSentryErrorTracker` auto-binds when `SENTRY_DSN` is set, dynamically importing `@sentry/node` (optional dependency) so the vendor SDK is only loaded when configured. Tracing is gated on `SENTRY_TRACES_SAMPLE_RATE`. Falls back to no-op if the SDK is absent - boot never crashes.
- **Capture wired at the three previously log-only seams**: oRPC `onError` interceptor (carries `userId`/`traceId`, skips expected `ORPCError`s), EventBus subscriber failures, OutboxRelay drain failures.
- Auto-binds in `createApp` before the event-bus/outbox so downstream `c.get(ERROR_TRACKING)` is synchronous; a consumer overlay re-providing the token still wins (last registration wins).
- Docs (`docs/adapters/error-tracking.md`) + regenerated catalog.

## Why

The platform had no error/crash reporting - unhandled errors only hit pino/stdout. This adds reporting while keeping it vendor-neutral so a consumer can swap Sentry for another provider without core changes.

## Acceptance

- Without `SENTRY_DSN`: no-op tracker, `@sentry/node` never loaded, platform unchanged.
- With `SENTRY_DSN`: unhandled errors reported with `trace_id` tag + user id; expected `ORPCError`s are not reported.
- Overlay re-providing `ERROR_TRACKING` overrides the default.

## Verify

`pnpm verify` + `pnpm boundaries` green; catalog/openapi no-drift.


zaxovaiko:
Superseded: this branch got clobbered to an unrelated commit and showed the wrong diff. Reopened with the correct sentry code in a fresh PR off current dev.

PR #13, opened 2026-07-15
feat: vendor-neutral error-tracking adapter seam via logger sink

zaxovaiko:
## What

Adds a vendor-neutral `ERROR_TRACKING` adapter seam so the runtime can report unhandled errors to a vendor (Sentry, PostHog, Rollbar, ...), wired entirely through the logger. Core ships no vendor SDK - the vendor is a consumer overlay.

- **Contract** (`contracts/adapters/error-tracking.ts`): `ErrorTrackingAdapter` port + `ERROR_TRACKING` token. Fire-and-forget `captureException(error, context?)` - a reporter failure never breaks the calling path.
- **Report through the logger** (`server/kernel/error-sink.ts` + `logger.ts`): `createLogger`'s pino hook forwards any error-level log carrying `err` to a process-wide error sink. One `logger.error({ err })` both logs and reports - no separate `captureException` call at each site, and every error log is covered rather than a hand-picked set of seams.
- **Bridged in `createApp`**: after plugins load, the sink is wired to whatever `ERROR_TRACKING` an overlay bound. Unbound -> error logs stay logs-only; the engine imports no vendor SDK and has no `SENTRY_DSN` of its own.
- Docs (`docs/adapters/error-tracking.md`) show the Sentry overlay recipe, mirroring the payment/KYC vendor overlays. Regenerated catalog.

## Why

The platform had no error/crash reporting - unhandled errors only reached pino/stdout. This adds reporting while keeping core vendor-neutral: a consumer picks Sentry, PostHog, or anything else via an overlay, with zero vendor code shipped to shops that don't use it.

## Acceptance

- No overlay bound: error logs stay logs-only, platform unchanged, no vendor SDK loaded.
- Overlay binds `ERROR_TRACKING`: every `logger.error({ err })` is both logged and reported to the vendor.
- Reporter failures are swallowed and never break the calling path.

## Verify

`pnpm verify` green (typecheck + 60 unit tests + lint + module-shape + boundaries); catalog/openapi no-drift. Rebased on `dev`, conflicts resolved, lockfile unchanged.


klaudia-blazyczek-blurify (dismissed):

packages/core/src/server/kernel/logger.ts
```diff
@@ -1,5 +1,45 @@
...
+        if (!err || report === false) {
+          return;
+        }
+
+        const ctx = getCurrentRequestContext();
+        const message = typeof second === 'string' ? second : undefined;
+        reportError(err, {
+          ...(ctx?.userId ? { userId: ctx.userId } : {}),
+          ...(ctx?.traceId ? { traceId: ctx.traceId } : {}),
+          extra: message ? { ...rest, message } : rest,
```
klaudia-blazyczek-blurify, line 40: `rest` is forwarded 1:1 into `ErrorContext.extra`, so any PII/secrets in structured logs will be sent to the external tracker. This conflicts with `.rulesync/rules/conventions.md:219`

klaudia-blazyczek-blurify (changes requested):

packages/core/src/server/kernel/logger.ts
zaxovaiko: `conventions.md:219` is about PR titles, description, and git commits. I'd add info about filtering logs too.

klaudia-blazyczek-blurify (approved):

PR #16, opened 2026-07-16
feat(web): chat navigation implementation

agniev-a-hub:
Implemented items:

  - Chat domain expansion: room navigation, room details, message history, online counts, join/leave flows, and the related
    contract/router/service updates.

  - Chat moderation and access control: private-room membership handling, ban/kick support, room slug conflict handling, and the
    moderation-related race fixes.

  - Chat realtime delivery: Ably-based realtime transport and authorizer integration, plus client hook support for joining
    rooms.

  - Chat schema and data migration work: new/updated Drizzle migrations, snapshots, and room data mapping/cleanup.
  - Identity/PAM work: phone login flow, SMS adapter contract, region guard, and related rate-limit coverage.
  - Core realtime/runtime infrastructure: durable seams for event bus, realtime transport, broker, logger, error reporting, rate
    limiting, and app runtime assertions.

  - Test coverage expansion: new tests across chat, identity, realtime, DB migration, logger, broker, cache, and rate limiter
    behavior.

  - Platform plumbing and docs: contract adapter extensions, audit plugin updates, catalog regeneration, Rulesync/AGENTS
    updates, and ADR/docs additions.

  - Seed and tooling updates: demo-data seeding, DB seed changes, module-shape checks, and generated package metadata updates.


packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -89,13 +169,22 @@ export class ChatService {
...
       } else {
         deliver(message);
       }
     });
+    return () => {
+      unsubscribe();
+      presence?.leave(channel, presenceMemberId, connectionId);
+    };
+  }
+
```
klaudia-blazyczek-blurify, line 187: [BLOCK] There is a race here. The code checks that the user is not banned, then inserts the membership in a separate DB operation. If a ban happens in between, this insert can add the user back. Since `verifyRoomAccess` only checks membership, the banned user can still enter the room. Please make the ban check and insert atomic, or also check the ban when granting access.

extensions/ably/ably-client-authorizer.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  async issueGrant(input: RealtimeClientAuthorizerInput): Promise<RealtimeConnectionGrant> {
+    const capability = Object.fromEntries(
+      input.channels.map((channel) => [channel, ABLY_SUBSCRIBE_CAPABILITIES]),
+    );
+    const tokenRequest = await this.rest.auth.createTokenRequest({
+      // Deliberately ignore clientId from the request: identity and presence are
+      // bound to the authenticated player, not a browser-supplied identifier.
+      clientId: input.userId,
+      ttl: TOKEN_TTL_MS,
```
klaudia-blazyczek-blurify: [BLOCK] This token is valid for one hour and includes every private-room channel returned by `getConnection`. Kicking or banning a member only updates the database; it does not invalidate an existing Ably token. That user can therefore keep receiving private-room messages until the token expires. Please add revocation/disconnect handling, or re-check access before delivery.

packages/core/src/engagement/chat/router/index.ts
```diff
@@ -4,21 +4,41 @@ import {
...
 } from '@openora/core/server';
-import type { RealtimeClientAuthorizer } from '@openora/core/contracts';
+import {
+  makeRateLimitKey,
+  RATE_LIMIT_KEYS,
+  type RateLimiterAdapter,
+  type RealtimeClientAuthorizer,
+} from '@openora/core/contracts';
 import { chatContract } from '../contract/index.js';
 import {
```
klaudia-blazyczek-blurify, line 18: [BLOCK] `listRooms` now calls `getUserId`, so anonymous users get a 401 even when they only want to see public rooms. The chat rules say public rooms should be open to everyone. The same problem affects public room messages, streams and online counts. Please keep auth optional for public rooms and require it only for private rooms.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -1,37 +1,51 @@
...
 import type { RealtimeTransport } from '@openora/core/contracts';
-import { eq, and, isNull, lt, desc, asc, notInArray } from 'drizzle-orm';
+import { eq, and, isNull, lt, desc, asc, notInArray, inArray, count, ne } from 'drizzle-orm';
 import { user } from '@openora/core/pam/schema/identity';
 import type { User } from '@openora/core/pam/schema/identity';
-import { chatRoom, chatMessage, chatUserBlock } from '../schema/index.js';
-import type { ChatRoom, ChatMessage } from '../contract/index.js';
+import {
+  chatRoom,
+  chatMessage,
```
klaudia-blazyczek-blurify, line 21: [WARN] The slug is checked before the insert, but two requests can pass that check at the same time. One can then get a raw PostgreSQL `23505` error, which becomes a 500 instead of the expected 409. The same race exists in `updateRoom`. Please translate the database error into `ChatRoomSlugConflictError`.

klaudia-blazyczek-blurify (changes requested):

agniev-a-hub:
Verified and fixed all four findings:

  - Join/ban race: join and ban now use the same room-scoped advisory transaction lock, with the ban check and membership insert
    performed atomically.

  - Ably access: kick/ban now revoke the user’s Ably tokens immediately via clientId with allowReauthMargin: false. The Ably API
    key must have revocable tokens enabled.

  - Anonymous access: public room lists, history, streams, online counts, and room details no longer require authentication;
    private rooms still do.

  - Slug conflicts: concurrent create/update requests now translate PostgreSQL 23505 errors into ChatRoomSlugConflictError
    (409).

zaxovaiko (commented):
Hm, why did you include Ably here, not in the consumer's project? OSS package must be provider-agnostic.

packages/core/src/engagement/chat/drizzle/migrations/0002_curly_phalanx.sql
zaxovaiko, line 1: I think we can squash all migrations, since names are not useful anyway. That will reduce the number of added lines and snapshots.

package.json
```diff
@@ -67,6 +67,9 @@
     "turbo": "2.9.16",
     "typescript": "6.0.3"
   },
+  "optionalDependencies": {
```
zaxovaiko: Openora should know nothing about end consumers' dependencies, even optional ones.

extensions.config.ts
```diff
@@ -51,4 +51,6 @@ export const extensions = [
   // the Redis cache/rate-limiter (ADR-0028). MESSAGE_BROKER stays in-process (AMQP_URL
   // enables the outbox; a durable broker is still an overlay). Rebind any of these
   // with your own infra plugin (Container last-wins).
+  // Ably is opt-in: without ABLY_API_KEY its plugin leaves the SSE defaults bound.
```
zaxovaiko: It can't be there. Remove it. Ably is used on the end consumer's project; that's not part of the OSS package.

extensions/ably/__tests__/ably.test.ts
zaxovaiko, line 1: Ably is not part of the OSS package. Remove it.

zaxovaiko (changes requested):

agniev-a-hub:
Corrections has been made, but for some reason github don't mark your comments as updated with some changes :(

PR #18, opened 2026-07-16
test: run infra and module tests against real postgres/redis

zaxovaiko:
## Summary

Adds a real-infra test harness (`createTestDb` / `createTestRedis` in `@openora/core/testing`) and moves the kernel drivers plus an initial set of module tests onto real Postgres and Redis. Test-layer only - no production code, contracts, routes or tables change. CI's `verify` job gains `postgres:16-alpine` + `redis:7-alpine` service containers, and Redis leaves the opt-in compose profile so `docker compose up -d` covers local runs.

## Why

Introduced these changes because mock choreography asserts mechanics, not results. A vi-mocked Drizzle chain cannot prove unique-index dedupe, `FOR UPDATE` locking under concurrent writers, atomic conditional updates, or real read-through cache invalidation - it only proves that a query builder was called in a particular order, which survives a genuine regression. Real infra asserts the outcome instead: rows, cache state, consumer-group behaviour.

The goal is that every DB-touching service test asserts against real Postgres/Redis, keeping only external-vendor and cross-module port doubles. This is the first slice; the suite stays fully parallel (ephemeral database per file, per-worker Redis logical DB).

## Worth knowing

Converted here: the harness, the kernel drivers (cache, rate-limiter, Redis Streams broker, event-bus, BullMQ), `cms.reads`, `lobby.cache`, `gaming.service`, `audit.service`, `iam.service`, `identity.service`, `phone-login.service`, and the `identity`/`wallet` rate-limit tests. The conversions bought real coverage that was previously unassertable - assignment unique-index dedupe, the `FOR UPDATE` last-super-admin guard under concurrent unassigns, atomic failed-attempt increments, progressive lockout escalation, and an anti-enumeration shadow record that survives a fresh service instance because the state lives in Redis.

Deferred to follow-ups: compliance, engagement, the remaining pam/identity and pam/tag services, wallet, `outbox/relay`, and the router tests. Staying mocked by design: `server/**` engine tests cannot import a domain schema (engine/domain boundary, ADR-0024/0025), and external-vendor / cross-module port doubles stay mocked everywhere (email, SMS, KYC, PSP, better-auth, `IdentityReader`, `WalletReader`, `EventBus`, `Logger`).

Fixed along the way: the `redis-streams-broker` "close stops the read loop" test raced under parallel load. The behavioural guarantee always held, but the stricter `pending === 0` probe did not - Redis can serve an entry to a blocked `XREADGROUP` before it processes the reader's disconnect, leaving one read-but-unacked entry that is never dispatched and is later reclaimed by `XAUTOCLAIM`. That is the at-least-once contract working as intended, so the assertion was hardened rather than the isolation model.


agniev-a-hub (dismissed):
I believe you ran your agent agains it, so don't think it makes sence to spend 30+ mins here :) 

Conflicts to resolve only.

zaxovaiko:
> I believe you ran your agent agains it, so don't think it makes sence to spend 30+ mins here :)
> 
> Conflicts to resolve only.

Yes, but still would be good to take a look on changes. Two pair of eyes is always better than one

zaxovaiko:
@copilot resolve the merge conflicts in this pull request

PR #20, opened 2026-07-17
feat(pam): automatic wallet-based tags with high_risk resweep and race-safe assignment

marek-chmielowski-blurify:
## Summary

Adds two automatic wallet-based player tags (`withdrawal_review`, `dormant_high_roller`) and a daily `high_risk` resweep, and closes gaps found while building them: tag routes were unguarded, concurrent tag assignment could race, and `AppError`'s callers didn't match its constructor signature.

## Changes

- `pam/tag`: `withdrawal_review` tag assigned on `wallet.withdrawal.requested` (the attempt, before funds move); `dormant_high_roller` assigned by a new daily co-occurrence sweep (inactive + holds `high_roller`), removed on next login (`onUserLogin`) - the sweep itself is assign-only; daily `high_risk` resweep removes the tag once withdrawal frequency drops below threshold (frequency dimension only, amount dimension is never resweepable - see AGENTS.md for the AML rationale).
- `pam/tag`: `playerTag.assignMetadata` records which breach condition(s) - withdrawal amount and/or frequency - triggered a `high_risk` assignment, merging in a later-discovered dimension while the tag stays active. The daily resweep now only auto-clears a tag whose recorded reason is frequency-only; a recorded amount breach, or missing/legacy metadata, is never auto-cleared.
- `pam/tag`: all tag routes (`createTag`/`deleteTag`/`listPlayerTags`/`listAssignableTags`/`assignPlayerTag`/`removePlayerTag`) now gated by `adminGuard` on a new `tag` permission resource (`view`/`create`/`delete`); domain errors mapped via `mapErrors` instead of leaking as raw 500s.
- `pam/tag` schema: partial unique index (`player_tag_active_key` on `(tagId, playerId) WHERE removedAt IS NULL`) backing an `onConflictDoNothing` insert - closes the concurrent-assignment / at-least-once-redelivery race that the app-level pre-check alone couldn't.
- `wallet`: `WalletReader.getWithdrawalCountsInWindow` - batched per-user withdrawal counts for the `high_risk` resweep (one query for all holders, not N).
- `audit`: subscribes to new `tag.created`/`tag.deleted` events so tag catalog admin actions are audited.
- `common/errors`: fixed `ConflictError`/`NotFoundError`/`ValidationError`/`mapDbError` to call `AppError` with the correct `(message, code)` argument order (was `(code, message)`, silently swapped).
- Migration `0002` (squashed from the original `0002`-`0004`) for the `pam/tag` module: new tag keys, partial unique index. Migration `0003` adds the `assignMetadata` jsonb column.

## Acceptance criteria

- [x] `withdrawal_review` tag assigned on withdrawal request when amount crosses the configured threshold, sticky, never auto-removed
- [x] `dormant_high_roller` tag assigned by the daily sweep based on inactivity + `high_roller` co-occurrence; removed on next login (not by the sweep)
- [x] `high_risk` tag auto-removed once withdrawal frequency drops back under threshold, but never when the recorded assignment reason includes an amount breach
- [x] Tag routes require admin permission
- [x] Concurrent/duplicate tag assignment is impossible even under at-least-once event redelivery

## Checklist

- [ ] New data tables carry `tenantId` and are RLS-covered - N/A, no new tables (index-only + jsonb-column migrations)


packages/core/src/pam/tag/drizzle/migrations/0002_silent_skaar.sql
zaxovaiko, line 1: I think we can squash those migrations into a single one to avoid adding more snapshots and files overall

packages/core/src/pam/tag/service/tag-evaluation.service.ts
```diff
@@ -221,4 +268,68 @@ export class TagEvaluationService {
...
+    // One set-based query for all holders' counts, then bound only the per-user
+    // removal writes - the reads are already batched, not per-row.
+    const countsByUser = await this.walletReader.getWithdrawalCountsInWindow(
+      holderUserIds,
+      thresholdDays,
+    );
+
+    await mapConcurrent(holderUserIds, EVAL_CHUNK_SIZE, async (userId) => {
+      const countInWindow = countsByUser.get(userId) ?? 0;
+      if (countInWindow < thresholdCount) {
```
klaudia-blazyczek-blurify, line 572: `high_risk` resweep can auto-remove players who are still risky by amount. `onWithdrawalCompleted` assigns `high_risk` when `amountBreached || countBreached`, but the daily resweep removes every current holder when only the withdrawal count is below `thresholdCount`. With a mixed amount+frequency rule, one large withdrawal can assign the tag, then the next daily sweep clears it because count is low. Store assignment provenance, split the tag dimensions, or re-evaluate a durable amount condition before removal.

packages/core/src/pam/tag/plugin.ts
```diff
@@ -28,6 +28,9 @@ export default definePlugin({
     ctx.events.on('wallet.withdrawal.completed', (p) =>
       tagEvaluationService?.onWithdrawalCompleted(p),
     );
+    ctx.events.on('wallet.withdrawal.requested', (p) =>
```
klaudia-blazyczek-blurify: `withdrawal_review` is not guaranteed to land before auto-payout. Wallet emits `wallet.withdrawal.requested` and immediately awaits `maybeAutoApprove`, while `EventBus.emit()` publishes with `void` and the broker invokes handlers without awaiting them. With auto-withdrawal enabled, a threshold-crossing withdrawal can be approved/completed before this async tag handler finishes. This needs to be synchronous in the withdrawal approval path or auto-approval must check the same rule/tag before paying out.

packages/core/src/pam/tag/drizzle/migrations/0002_silent_skaar.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TYPE "public"."tag_key" ADD VALUE 'dormant_high_roller';--> statement-breakpoint
```
klaudia-blazyczek-blurify, line 1: Upgraded databases get the new enum values but not the required `tag` and `tag_rule` rows. Fresh DB seed covers `dormant_high_roller` and `withdrawal_review`, but migrations only alter the enum; `TagRuleService.upsertTagRule` first looks up an existing tag row and throws if it is missing. Add idempotent migration backfills for both tag catalog rows and default rule rows.

packages/core/src/pam/tag/drizzle/migrations/0004_omniscient_ezekiel_stane.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX "player_tag_active_key" ON "player_tag" USING btree ("tag_id","player_id") WHERE "player_tag"."removed_at" is null;
```
klaudia-blazyczek-blurify: This unique-index migration can fail on existing production data created by the race it is fixing. The old schema only had a non-unique `(player_id, tag_id)` index, and the PR notes the app-level pre-check allowed duplicate active assignments. If any active duplicate `(tag_id, player_id)` rows already exist, Postgres aborts `CREATE UNIQUE INDEX`. Add a dedupe/remediation step or an explicit preflight that reports duplicates before creating the index.

packages/core/src/contracts/adapters/wallet-reader.ts
```diff
@@ -5,6 +5,8 @@ export type WalletReader = {
   getLifetimeDeposit(userId: string): Promise<string>;
   /** Count of completed withdrawals for a player within the last windowDays days. Used for high_risk evaluation. */
   getWithdrawalCountInWindow(userId: string, windowDays: number): Promise<number>;
+  /** Completed-withdrawal counts within the last windowDays days for a batch of players, keyed by userId (absent key = 0). Used for the high_risk daily resweep. */
+  getWithdrawalCountsInWindow(userIds: string[], windowDays: number): Promise<Map<string, number>>;
```
klaudia-blazyczek-blurify: Adding `getWithdrawalCountsInWindow` as a required `WalletReader` method is a breaking port change for any external/operator wallet adapter. Core `WalletReaderService` is updated, but existing downstream bindings implementing the contract will fail at compile/runtime. Consider a backward-compatible optional method with fallback batching via `getWithdrawalCountInWindow`, or version the port.

packages/core/src/pam/tag/drizzle/migrations/0004_omniscient_ezekiel_stane.sql
klaudia-blazyczek-blurify: The migration still creates the unique index directly. If duplicate active rows already exist, the migration will fail before the index is created. Please clean up those duplicates first, or add a check that stops with a clear error.

packages/core/src/pam/tag/service/tag-evaluation.service.ts
klaudia-blazyczek-blurify: The current resweep still removes the tag when the count is below the threshold, even if the tag was added because the amount threshold was exceeded. Please make removal consider both reasons, or store which condition caused the tag to be added.

packages/core/src/pam/tag/drizzle/migrations/0002_silent_skaar.sql
klaudia-blazyczek-blurify: The migration still has no inserts for the new `tag` and `tag_rule` rows. This means an upgraded database can have the enum values but still fail when the rule is updated. Please add safe, repeatable inserts for the two tags and their default rules.

marek-chmielowski-blurify: This is a deliberate, repo-wide convention, not an oversight in this PR - documented in this module's AGENTS.md: "Migrations never seed data (this module included) - the tag/tagRule catalog rows ... are created by seedTag() (seed/index.ts, wired into pnpm seed via tools/db/seed.ts), not baked into a migration, consistent with every other module." Adding INSERTs directly to this migration would both break that cross-module pattern and require hand-editing an already-generated migration file (which is separately disallowed - see db-conventions.md).

seedTag()/seedTagRules() are idempotent (onConflictDoUpdate), specifically so they are safe to re-run against an already-migrated database. AGENTS.md already documents the operational step: "on an already-migrated database, re-run pnpm seed after applying that migration so those two tags exist before the daily sweep/resweep references them by key."

Given every other module in this repo follows the same split (schema/enum changes in migrations, catalog rows via pnpm seed), I would rather keep this PR consistent with that than introduce a one-off exception here. Happy to revisit the repo-wide convention itself in a separate discussion if you feel strongly.

packages/core/src/pam/tag/drizzle/migrations/0004_omniscient_ezekiel_stane.sql
marek-chmielowski-blurify: Fixed in 5f81644 - added a `preSql` option to `runMigrations` (mirroring the existing `extensions` hook - same rationale, raw SQL instead of a bare identifier) so this stays a data-cleanup step outside the generated migration file rather than a hand-edit of `0002_silent_skaar.sql`.

`pam/tag/migrate.ts` now passes a dedupe statement that soft-removes every duplicate active `(tag_id, player_id)` row except the earliest (by `created_at`) before the migrator reaches the unique-index migration - using the table's normal `removed_at`/`removal_reason`/`removal_actor` soft-delete columns so the cleanup itself is audit-visible, not a silent DELETE.

Guarded with `to_regclass('public.player_tag') IS NOT NULL` since `preSql` runs before *every* pending migration including 0000 (which creates the table) - without the guard this would break every fresh install. Verified against a real Postgres instance, not just mocked: built a scratch DB, applied only 0000+0001, manually inserted 3 duplicate active rows for the same (tag, player), ran the real `migrate()` - it kept the oldest row, soft-removed the two duplicates, and the unique index then created cleanly. Re-ran `migrate()` again on that same DB to confirm idempotency (no further changes), and separately confirmed the `to_regclass` guard no-ops correctly against a brand-new empty database.

packages/core/src/pam/tag/drizzle/migrations/0002_silent_skaar.sql
marek-chmielowski-blurify: Done - the tag module migrations are squashed into a single 0002_silent_skaar.sql now, no separate 0003/0004 files.

packages/core/src/pam/tag/service/tag.service.ts
```diff
@@ -52,27 +67,52 @@ export class TagService implements PlayerTags {
...
     }
   }
 
-  public async deleteTag(args: DeleteTagInput) {
+  public async deleteTag(args: DeleteTagInput, actorId: User['id']) {
     try {
       const db = this.drizzle.db;
       await db.delete(tag).where(eq(tag.key, args.key));
+      void this.event.emit('tag.deleted', { key: args.key, actorId });
       return true;
```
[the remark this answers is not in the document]
marek-chmielowski-blurify: Fixed - deleteTag now checks .returning() before emitting: if no row was actually deleted (missing key), it throws TagNotFoundError instead of falling through to the tag.deleted emit. See the .returning().length === 0 guard in TagService.deleteTag.

klaudia-blazyczek-blurify (dismissed):

zaxovaiko (approved):
Giving an approve. Code review will be done a bit later.

PR #21, opened 2026-07-17
fix(repo): fix session cookies for login via phone number (rememberMe option for consumers)

klaudia-blazyczek-blurify:
## Summary

Extends phone-login `verifyOtp` to emit a properly HMAC-signed session cookie on successful
authentication, so browser clients receive a `set-cookie` response header without needing
to handle the raw token themselves. Adds typed error reasons (`otp_cancelled`, `rg_blocked`)
to the contract surface so consumers can branch on structured error data instead of message strings.

## Changes

- **`server/auth`** - new `signSessionCookie` helper: serializes a `better-auth` session token
  into an HMAC-SHA256-signed cookie using the auth context's secret and cookie attributes
  (no hardcoded flags - inherits `httpOnly`, `secure`, `sameSite`, etc. from `better-auth` config)
- **`pam/identity` - service** - `verifyOtp` receives a `resHeaders: Headers` arg and appends
  `set-cookie` after a successful session mint; wires `auth: Auth` dep for cookie config access
- **`pam/identity` - router** - threads `context.resHeaders` into `verifyOtp`
- **`pam/identity` - plugin** - resolves `AUTH_SESSION` and passes `.auth` to `PhoneLoginService`
- **`contracts/schemas/identity`** - adds `PHONE_LOGIN_ERROR_REASONS` enum triple
  (`PhoneLoginErrorReasonSchema` + `PhoneLoginErrorReason`) and attaches `data.reason` to
  `OtpCancelledError` and `rg_blocked` throws
- **Tests** - unit tests for `signSessionCookie` (happy path, `rememberMe` on/off, attribute
  passthrough) and extended `phone-login.service` tests covering the new cookie emission path

## Acceptance criteria

- [ ] `phoneLoginVerify` response sets `set-cookie` with a valid HMAC-signed session token
- [ ] Cookie attributes (`httpOnly`, `secure`, `sameSite`, `path`) are inherited from `better-auth` config, not hardcoded
- [ ] `rememberMe: true` sets `maxAge` derived from session expiry; `rememberMe: false` omits `maxAge`
- [ ] `OtpCancelledError` and RG-blocked error carry `data.reason` matching `PhoneLoginErrorReasonSchema`
- [ ] All new paths covered by unit tests

## Checklist

## Notes

`signSessionCookie` is a pure function exported from `@openora/core/server` - consumers that
roll their own auth flow (non-better-auth) can use it directly if they supply the cookie config.

**Why we bypass `auth.api` for cookie setting:** Phone login mints the session directly into the
DB (bypassing better-auth's plugin chain by design) so that 2FA/TOTP is never triggered for this
flow. Because `auth.api.signIn*` was never called, there is no library-generated `Set-Cookie`
header to forward - the cookie must be constructed manually. We source the signing secret and
cookie attributes from `auth.$context` (better-auth internal, not a public API) to stay in sync
with whatever the operator configures. Follow-up: expose `secret` + `sessionCookieConfig`
through `AuthOptions` so this stops relying on internals.

**Why `rememberMe` tests passed on `dev` but the feature didn't work in the consumer app:** The existing
test asserted `out.session.expiresAt` in the JSON response body, which correctly reflected the
30-day TTL written to the DB session row. What the test did NOT cover was the `Set-Cookie`
header - because on `dev` no cookie was ever emitted by `verifyOtp`. Browser-based clients
derive session lifetime exclusively from `Max-Age`/`Expires` on the cookie, not from the JSON
payload. Result: the DB row had the right expiry, the test was green, but every browser session
expired on tab close (no `Max-Age` = session cookie). This PR fixes that by appending
`set-cookie` with `maxAgeSeconds` derived from `sessionExpiresAt` when `rememberMe` is true.


zaxovaiko (commented):
Hm, I don't think `rememberMe: false` actually sticks here. We sign the session cookie but never set better-auth's paired `dont_remember` cookie, so on the next request `getSession` runs its refresh heuristic against the global `expiresIn: 30d` / `updateAge: 1d` and bumps `expiresAt` back to 30 days - a "don't remember me" phone login ends up persisting a month on shared devices (`phone-login.service.ts:308-320`).

Shouldn't we also mint the `dont_remember` signed cookie (same HMAC scheme, `authContext.authCookies.dontRememberToken`) when `rememberMe` is false, the way the real sign-in path does via `forwardCookies`? That makes `getSession` take the early-return branch instead of refreshing. The `true` path is fine since 30d matches the global config.

Worth a regression test on the false path too - it's the exact thing this PR is fixing.

zaxovaiko (dismissed):

PR #28, opened 2026-07-21
fix(audit): minor bug fixes related to audit 

agniev-a-hub:


packages/core/src/audit/service/audit.service.ts
```diff
@@ -73,6 +73,16 @@ function likePrefix(prefix: string): string {
   return `${prefix.replace(/[\\%_]/g, '\\$&')}%`;
 }
 
+function startOfDayUtc(dateStr: string): Date {
```
zaxovaiko: Maybe unit tests for those? I don't trust dates

agniev-a-hub: Screenshot 2026-07-21 at 18 11 34

It will give invalid date for that one. Even validation makes no sence, since we don't have any aggregate errors for FE display. If blocker IYO will add.

zaxovaiko: Not a blocker, but would be nice to have actual test that verifies the implementation

zaxovaiko (dismissed):
LGTM. Two non-blockers for follow-up: let's tighten `AuditExportFiltersSchema` dates to `z.string().date()` since `/audit/export` is public, and a regression test on the `toDate` day-boundary fix would be good since it's a regulatory surface.

zaxovaiko (approved):

PR #29, opened 2026-07-22
fix(pam): complete BF-317 automatic tagging

marek-chmielowski-blurify:
## Summary

Adds self-exclusion, multi-account and atomic-level player tags, plus completes the BF-317 automatic wallet-based tagging rules. BF-317.

## Changes

- `pam/tag`: self-exclusion, multi-account and atomic-level tag rules; expanded tag-evaluation and tag-service coverage
- `pam/identity`: identity-reader adapter changes supporting multi-account detection
- `pam/player-management`: player service updates for tag-driven state
- `compliance`: KYC service adjustment
- `audit`: plugin wiring for new tag-related audit events
- `contracts`: new `tag` schema slice, new event payloads in `events.ts`, `identity-reader` adapter interface changes
- `testing`: new BF-317 tag e2e test suite

## Acceptance criteria

- [x] Self-exclusion tag applied automatically based on wallet/player activity
- [x] Multi-account tag applied automatically based on identity-reader detection
- [x] Tag rules evaluate atomically at the player level (race-safe)

## Checklist

- [x] New data tables carry `tenantId` and are RLS-covered (`pnpm regen` runs `gen:rls`) - no new tables, migration only adds enum values

## Notes

Targets `feat/automatic-wallet-based-tags` (not `dev`) - this branch builds on top of that in-progress feature branch (open PR #20 into `dev`).

Closes BF-317


packages/core/src/pam/tag/drizzle/migrations/meta/_journal.json
zaxovaiko, line 1: I'm not sure it will work properly. `0003_warn_sil...` migration was already executed, but here you replaced it with a new one. You should add another migration next to the old one instead.

marek-chmielowski-blurify: That migration is from https://github.com/blurifycom/openora/pull/20 pull request. If we merge this one to the other one we are all good. I didnt want to add more migrations to keep it clean.

packages/core/src/pam/tag/service/tag.service.ts
```diff
@@ -327,6 +328,35 @@ export class TagService implements PlayerTags {
     }
   }
 
+  private async _removePlayerTagOnTx(trx: DrizzleTx, args: RemovePlayerTagInput) {
+    const foundTag = await this._findTagByKeyOrThrow(args.tagKey, trx);
+    const active = findOneOrThrow(
+      await trx
```
klaudia-blazyczek-blurify, line 334: **Race condition (Lost update)**

This `select` reads the active tag row without a row lock, and the subsequent `update` doesn't check if the row was already removed. If two `removePlayerTag` requests run concurrently, both will read the same row and try to update it. The second transaction will silently overwrite the `removalReason` and `removedAt` timestamp of the first one.

**Fix:** You can either add a row lock to the select, or add `isNull(playerTag.removedAt)` to the `.where()` clause of the `update()` below and throw an error if no rows are updated.

packages/core/src/pam/player-management/service/player.service.ts
```diff
@@ -227,6 +229,18 @@ export class PlayerService {
       );
     });
 
+    // Emitted AFTER commit (not inside the transaction callback, unlike the
+    // KYC_STATUS_WRITER emit above) - a level change is a best-effort fan-out, not a
+    // regulated single-writer seam, so it follows the standard post-commit event idiom.
```
klaudia-blazyczek-blurify, line 234: **Potential missing audit log on crash**

The `player.level.changed` event is emitted after the database transaction commits. If the application crashes immediately after the commit but before the event is emitted, this state-changing action will never reach the audit log, and the `level` tag won't be updated.

If this is considered a "must-not-lose" event (since it drives the audit log), consider using the outbox pattern so it's guaranteed to be delivered.

PR #30, opened 2026-07-22
fix(audit): fixing audit data missing for various events

agniev-a-hub:
Fixed the issue with audit log missing IP and user agent, since it has not been passed all over the app.


packages/core/src/contracts/schemas/events.ts
zaxovaiko: Looks like `tag.rule.upserted` got missed here - the service is emitting ip/userAgent now (`tag-rule.service.ts:86`), but this schema is still a bare `z.object` while everything else got `.merge(authContextBase)`. Let's do `authContextBase.extend({ tagKey, actorId, after })` to keep it consistent.

What worries me more is that TS didn't catch this at build. It's because `EventBus` has the second `emit(event: string, payload: unknown)` overload - when the typed one fails on the extra fields, resolution just falls through to the loose one and anything passes (missing fields would slip through the same way). 

Can we drop that overload and keep only the generic `emit<K extends DomainEventName>(event: K, payload: DomainEventPayload<K>)`? Then the payload is checked against the real schema and drift like this fails the build instead of shipping. If we genuinely need dynamic event names somewhere, let's move those to an explicit `emitRaw` rather than leaving the escape hatch open for every call site.

packages/core/src/server/kernel/router-utils.ts
```diff
@@ -40,3 +40,15 @@ function resolveAuth(context: unknown): AuthContext {
...
+  ip: string | null;
+  userAgent: string | null;
+} {
+  const fwd = headers['x-forwarded-for'];
+  const first = Array.isArray(fwd) ? fwd[0] : fwd;
+  const real = headers['x-real-ip'];
+  const ip = first?.split(',')[0]?.trim() || (Array.isArray(real) ? real[0] : real) || null;
+  const ua = headers['user-agent'];
+  return { ip: ip ?? null, userAgent: (Array.isArray(ua) ? ua[0] : ua) ?? null };
+}
```
zaxovaiko: `extractClientMeta` inlines its return type, and then the same `{ ip; userAgent }` shape gets re-typed by hand in ~30 places (`wallet.service.ts`, and across `chat/cms/compliance/tag/iam`). Can we please export it once and reuse it? Something like `export type ClientMeta = { ip: string | null; userAgent: string | null }` on `extractClientMeta`, then `meta?: ClientMeta` at the call sites. They've actually already drifted - the function returns required keys but the params are `ip?/userAgent?` optional, which is exactly what a single source of truth avoids.

Also we've got the same extraction logic twice - `extractClientMeta` for `NodeHeaders` here, and a separate `clientIp` + inlined user-agent in `identity.service.ts` (7 times) for Web Headers. Let's keep the `x-forwarded-for` handling in one place - normalize to `NodeHeaders` before calling `extractClientMeta`, or add a Headers overload. Otherwise when we fix the IP spoofing it'll land in one and miss the other.

packages/core/src/compliance/router/index.ts
```diff
@@ -41,26 +47,30 @@ export function createComplianceRouter({
...
+      return compliance.upsertLimit(getUserId(context), input, meta);
+    }),
 
-    deleteLimit: os.deleteLimit.handler(({ input, context }) =>
-      mapErrors({ NOT_FOUND: LimitNotFoundError, FORBIDDEN: LimitOwnershipError }, () =>
-        compliance.removeLimit(input.id, getUserId(context)),
-      ),
-    ),
+    deleteLimit: os.deleteLimit.handler(({ input, context }) => {
+      const meta = extractClientMeta(context.request.headers);
```
zaxovaiko: Hm, maybe we could move this `extractClientMeta` into the context resolve handler just to avoid duplication in every endpoint?

zaxovaiko (changes requested):
Let's tighten type-safety and duplications

packages/core/src/server/auth/admin-guard.ts
```diff
@@ -30,22 +31,26 @@ export class AdminGuard {
     private readonly events?: EventBus,
   ) {}
 
-  async assert(context: unknown): Promise<{ userId: string; role: string }>;
+  async assert(
+    context: unknown,
+  ): Promise<{ userId: string; role: string; ip: string | null; userAgent: string | null }>;
```
klaudia-blazyczek-blurify: The return type of `AdminGuard.assert` was updated here to include `ip: string | null` and `userAgent: string | null`. However, there are multiple test files that mock this function (e.g., `player.router.test.ts`, `wallet-auto-rule.router.test.ts`, `wallet.router.test.ts`, `admin-guard.test.ts`) which still return only `{ userId, role }`. Please update the mocked return values in these test files to include `ip: null, userAgent: null`.

klaudia-blazyczek-blurify (changes requested):

zaxovaiko (approved):

PR #34, opened 2026-07-22
feat(audit): general sort implementation

agniev-a-hub:
General sort implementation


packages/core/src/admin-console/contract/index.ts
```diff
@@ -87,7 +97,13 @@ export const backofficeContract = {
 
   listUsers: oc
     .route({ method: 'GET', path: '/backoffice/users' })
-    .input(PageQuerySchema.extend({ search: z.string().optional() }))
+    .input(
+      PageQuerySchema.extend({
+        search: z.string().optional(),
+        sortBy: AdminUserSortBySchema.optional(),
+        sortOrder: SortOrderSchema.default('asc').optional(),
```
klaudia-blazyczek-blurify: Could we keep the old descending default here? Before this change, `/backoffice/users` sorted by `createdAt` descending. In Zod 4 this default is applied when `sortOrder` is omitted, and the router passes `asc` to the directory, so existing clients now receive the oldest users first.

packages/core/src/pam/identity/service/session.service.ts
```diff
@@ -25,17 +25,36 @@ export class SessionService {
...
+        : sortBy === 'updatedAt'
+          ? session.updatedAt
+          : session.createdAt;
     const [rows, [{ n }]] = await Promise.all([
       db
         .select()
         .from(session)
         .where(where)
-        .orderBy(activeFirst, desc(session.createdAt))
+        .orderBy(activeFirst, dir(col))
```
klaudia-blazyczek-blurify: With `sortBy=expiresAt&sortOrder=asc`, `activeFirst` is still the first ordering key. An active session expiring tomorrow is returned before an expired session from yesterday, so the response is not actually sorted by `expiresAt`. Could we skip `activeFirst` when the client sends sort parameters?

klaudia-blazyczek-blurify (changes requested):

packages/core/src/pam/identity/service/session.service.ts
```diff
@@ -25,17 +25,36 @@ export class SessionService {
...
+    page,
+    limit,
+    sortBy,
+    sortOrder,
+  }: {
+    userId: User['id'];
+    page: number;
+    limit: number;
+    sortBy?: SessionSortBy;
+    sortOrder?: SortOrder;
```
zaxovaiko: I think we can extract common type here, smth like `PaginationOptions<{ userId: User['id'] }, SessionSortBy>`, where second param is `sortBy`

klaudia-blazyczek-blurify (dismissed):

zaxovaiko (approved):

PR #35, opened 2026-07-23
fix(identity): distinguish expired OTP from wrong code on phone login verify

klaudia-blazyczek-blurify:
## Summary

Phone login OTP verification returned the identical error shape for a wrong code and an expired one, making the "code expired" case practically unreachable client-side. Adds a `reason` field so callers can tell the two apart.

## Changes

- `contracts/schemas/identity.ts`: new `PhoneLoginOtpInvalidReasonSchema` (`'expired' | 'wrong_code'`)
- `pam/identity/service/phone-login.service.ts`: `OtpInvalidError` now takes and returns a `reason`; all three throw sites (missing OTP session, expired session, wrong hash) pass the correct one
- `pam/identity/__tests__/phone-login.service.test.ts`: assertions updated to check `reason`, plus a renamed case proving the actual bug (0 attempts remaining is now driven by `reason: 'expired'`, not inferred)
- `pam/identity/AGENTS.md`: documents the new `reason` field on `OtpInvalidError`

## Acceptance criteria

- [x] Wrong code returns `reason: 'wrong_code'`
- [x] Expired code (or missing OTP session) returns `reason: 'expired'`, independent of `attemptsRemaining`

## Checklist

- [x] `pnpm verify` is green (typecheck + lint + boundaries + module-shape + tests) — 598/598 unit tests, 60/60 tool tests
- [x] `pnpm verify:drift` is green (no diff in `docs/openapi.json`/`docs/catalog.json` beyond what's committed)
- [x] New cross-module talk goes through events / command ports / contracts / the `/schema` subpath (no direct module imports) — no new cross-module imports


packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -46,11 +48,11 @@ export function OtpCooldownError(retryAfterMs: number) {
...
 }
 
-/** Wrong or expired code, session still alive. */
-export function OtpInvalidError(attemptsRemaining: number) {
+/** Wrong or expired code, session still alive. `reason` lets the caller tell the two apart. */
+export function OtpInvalidError(attemptsRemaining: number, reason: PhoneLoginOtpInvalidReason) {
   return new ORPCError('UNPROCESSABLE_CONTENT', {
-    message: 'The code is invalid or has expired.',
-    data: { attemptsRemaining },
+    message: reason === 'expired' ? 'The code has expired.' : 'The code is invalid.',
```
zaxovaiko, line 66: Hm, what if you would like to translate errors on the frontend? How would we do that exactly?

klaudia-blazyczek-blurify: Well I didn't think about that bc I just wanted to do a quick fix but I will look into that (those errors were in english before)

klaudia-blazyczek-blurify: Also on FE part consumer app can handle errors by code e.g. UNPROCESSABLE_CONTENT

klaudia-blazyczek-blurify: @zaxovaiko should I do something about it or ignore it?

klaudia-blazyczek-blurify: Now when I thought about it, good point - UNPROCESSABLE_CONTENT only identifies the error category, so it cannot distinguish these cases. Consumers should use the machine-readable data.reason (expired / wrong_code) to map to their own localized UI copy, rather than render the backend message. I’ll keep the server message generic to make that boundary clear.

klaudia-blazyczek-blurify: Audited all ~20 ORPCError call sites. Webhooks (KYC/PSP signature checks) never reach a UI for anyone, left those. `router-utils.ts` + half of `admin-guard.ts` were collapsing real distinctions (not authenticated vs. malformed request vs. not-an-admin vs. missing a specific permission) into identical or string-only signals — added a shared `AuthGuardReasonSchema` (`common.ts`) and promoted resource/action out of the message string into data for the permission-check case. `identity.service.ts `ensureOk passes through an arbitrary upstream (better-auth) message with no fixed cause enum — flagged with a TODO rather than faked. Also found and fixed a live instance of the exact anti-pattern: backoffice was detecting a 403 via `error.message.includes('Forbidden')` — now reads `conventions.md` has the rule so this doesn't get re-litigated per-PR.

zaxovaiko: Let's add an enum (code) and use it for translations.

zaxovaiko (approved):
Seems like phone number and email is leaking in logs, but I will fix it in incoming PR

PR #38, opened 2026-07-25
feat(analytics): add financial analytics and conversion funnel reporting

zaxovaiko:
## Summary

Adds a read-only financial analytics + conversion funnel domain for the backoffice dashboard (BF-219). Deposits/withdrawals by currency and rail, net revenue, bonus cost, a GGR trend, and a registration-to-first-bet conversion funnel, all grouped by currency since the platform has no FX rates.

## Changes

- New top-level domain `packages/core/src/analytics/` (module id `analytics`, published as `@openora/core/analytics`) - owns no tables, reads `wallet`/`pam/identity`/`pam/profile` via their published `/schema` subpaths.
- Two contract slices: `financial` (`GET /analytics/financial/summary`, `GET /analytics/financial/ggr`) and `funnel` (`GET /analytics/funnel/conversion`). All routes guard on the existing `analytics` permission resource and cache reads briefly via the `CACHE` port.
- Shared `DateRangeSchema` / `GranularitySchema` added to `packages/core/src/contracts/schemas/`.
- New composite indexes: `wallet_transaction (status, type, created_at)`, `wallet_transaction (wallet_id, type, status)`, `user (created_at)`.
- `docs/adr/0033-analytics-as-a-read-only-domain.md` records the domain-placement decision and alternatives rejected.
- Unit tests (service, router, contract) plus an integration test in `packages/testing` covering multi-currency splits, zero-filled ranges, and nested funnel stages against a real Postgres + real HTTP app.

## Acceptance criteria

- [x] Financial dashboard: deposits by currency and rail, withdrawals by currency and rail, net revenue, bonus cost, GGR trend over time
- [x] Filterable by date range and currency
- [x] Conversion funnel: registrations - email verified - first deposit - first bet, with drop-off rates at each stage
- [x] Funnel filterable by date range

## Checklist

- [ ] New data tables carry `tenantId` and are RLS-covered - n/a, this domain adds no tables (single-tenant platform, ADR-0026)
- [x] Docs / AGENTS.md updated

## Notes

- Bonus cost reports credited bonus amount only ("bonus credited vs wagered" from the source spec is not achievable - no bonus wagering model exists on the platform).
- GGR is computed from `wallet_transaction` (`bet` minus `win`), not from `game_round.betAmount`/`winAmount` - `casino/gaming` never calls `WALLET_COMMANDS`, so those columns stay unwritten today. Wiring real stake debit/win credit into gameplay is a separate, larger change to the gaming domain, out of scope here (see ADR-0033 consequences).
- The consumer-side backoffice page depends on this contract; that PR follows once this lands and a canary publishes.

Closes BF-219

### DIFF
--- a/packages/core/src/analytics/service/financial-analytics.service.ts
+++ b/packages/core/src/analytics/service/financial-analytics.service.ts
@@ -78,23 +78,26 @@ export class FinancialAnalyticsService {
       baseConditions.push(eq(walletTransaction.currency, currency));
     }
 
-    const [byRail, byCurrency] = await Promise.all([
+    const totalsByRail = (type: 'deposit' | 'withdrawal') =>
       db
         .select({
           currency: walletTransaction.currency,
           rail: walletTransaction.rail,
-          type: walletTransaction.type,
           total: sql<string>`sum(${walletTransaction.amount})`,
         })
         .from(walletTransaction)
         .where(
           and(
             eq(walletTransaction.status, COMPLETED_STATUS),
-            inArray(walletTransaction.type, ['deposit', 'withdrawal']),
+            eq(walletTransaction.type, type),
             ...baseConditions,
           ),
         )
-        .groupBy(walletTransaction.currency, walletTransaction.rail, walletTransaction.type),
+        .groupBy(walletTransaction.currency, walletTransaction.rail);
+
+    const [deposits, withdrawals, byCurrency] = await Promise.all([
+      totalsByRail('deposit'),
+      totalsByRail('withdrawal'),
       db
         .select({
           currency: walletTransaction.currency,
@@ -117,12 +120,8 @@ export class FinancialAnalyticsService {
     ]);
 
     return {
-      deposits: byRail
-        .filter((r) => r.type === 'deposit')
-        .map((r) => ({ currency: r.currency, rail: r.rail, total: r.total })),
-      withdrawals: byRail
-        .filter((r) => r.type === 'withdrawal')
-        .map((r) => ({ currency: r.currency, rail: r.rail, total: r.total })),
+      deposits,
+      withdrawals,
       netRevenue: byCurrency.map((r) => ({ currency: r.currency, total: r.netRevenue })),
       bonusCost: byCurrency.map((r) => ({ currency: r.currency, total: r.bonusCost })),
     };
--- a/packages/core/src/analytics/service/funnel-analytics.service.ts
+++ b/packages/core/src/analytics/service/funnel-analytics.service.ts
@@ -84,18 +84,13 @@ export class FunnelAnalyticsService {
       first_bet: Number(row?.first_bet ?? 0),
     };
 
-    const orderedCounts = [
-      counts.registered,
-      counts.email_verified,
-      counts.first_deposit,
-      counts.first_bet,
-    ];
-
-    return FUNNEL_STAGES.map((stage, index) => ({
-      stage,
-      count: orderedCounts[index] ?? 0,
-      dropOffRate:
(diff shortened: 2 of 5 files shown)


packages/core/src/analytics/service/financial-analytics.service.ts
```diff
@@ -0,0 +1,193 @@
...
+    return {
+      deposits: byRail
+        .filter((r) => r.type === 'deposit')
+        .map((r) => ({ currency: r.currency, rail: r.rail, total: r.total })),
+      withdrawals: byRail
+        .filter((r) => r.type === 'withdrawal')
+        .map((r) => ({ currency: r.currency, rail: r.rail, total: r.total })),
+      netRevenue: byCurrency.map((r) => ({ currency: r.currency, total: r.netRevenue })),
+      bonusCost: byCurrency.map((r) => ({ currency: r.currency, total: r.bonusCost })),
+    };
```
agniev-a-hub, line 127: It's always better to delegate it to SQL. 
Also, some sort of batching would be indeed needed for sync operations if we need them.

packages/core/src/analytics/service/financial-analytics.service.ts
```diff
@@ -0,0 +1,193 @@
...
+      )
+      select b.currency, b.bucket, coalesce(agg.ggr, 0) as ggr
+      from buckets b
+      left join agg on agg.currency = b.currency and agg.bucket = b.bucket
+      order by b.currency, b.bucket
+    `);
+
+    const seriesByCurrency = new Map<string, GgrSeries>();
+    for (const row of result.rows) {
+      const series = seriesByCurrency.get(row.currency) ?? { currency: row.currency, points: [] };
```
agniev-a-hub, line 186: Soon we'll be able to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/getOrInsert#syntax

Not a case for now, only >26 node.

agniev-a-hub (dismissed):
WARN for .map() chain. 
INFO for .hasOrInsert() for future, no blockers.

packages/core/src/contracts/schemas/reporting.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from 'zod';
+import { TimestampSchema } from './common.js';
+
+export const DateRangeSchema = z.object({
+  dateFrom: TimestampSchema.optional(),
+  dateTo: TimestampSchema.optional(),
+});
```
klaudia-blazyczek-blurify, line 7: The `isDateRangeOrdered` function is exported but it is never applied to `DateRangeSchema`. Consider adding `.refine(isDateRangeOrdered, 'dateFrom must be before or equal to dateTo')` so the API returns a 400 Bad Request for invalid date ranges instead of silently returning empty data.

packages/core/src/analytics/service/funnel-analytics.service.ts
```diff
@@ -0,0 +1,101 @@
...
+      first_bet: Number(row?.first_bet ?? 0),
+    };
+
+    const orderedCounts = [
+      counts.registered,
+      counts.email_verified,
+      counts.first_deposit,
+      counts.first_bet,
+    ];
+
```
klaudia-blazyczek-blurify: Mapping this manually by index is a bit brittle. If a new stage is added to `FUNNEL_STAGES` in the future, this array needs to be manually updated as well. Consider looking up the count directly using the stage name from the `counts` object to prevent potential bugs.

packages/core/src/analytics/service/financial-analytics.service.ts
zaxovaiko: True. Missed that while was doing review

PR #41, opened 2026-07-27
feat(chat): chat commands implementation

agniev-a-hub:


packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,671 @@
...
+    // Resolve sender's username for the gift card metadata.
+    const senderSummaries = await this.directory.lookupPlayers([actorId]);
+    const senderSummary = senderSummaries.find((s) => s.userId === actorId);
+    if (!senderSummary) {
+      throw new ChatPlayerNotFoundError(actorId);
+    }
+    const senderUsername = senderSummary.username;
+
+    const { msg, giftId, currency } = await this.drizzle.db.transaction(async (tx) => {
+      const debit = await this.wallet.debit(tx, {
```
klaudia-blazyczek-blurify: This money operation has no idempotency key or database guard. If the client retries after a timeout, it creates another debit and another gift. Please store a request or operation id with a unique constraint and return the original result for retries.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,671 @@
...
+          // Pattern: insert with a temporary value, then set the real foreign reference.
+          messageId: '00000000-0000-0000-0000-000000000000',
+        })
+        .returning();
+
+      if (!giftRow) {
+        throw new InsufficientBalanceError();
+      }
+
+      const systemMsg = await this.systemWriter.postSystemMessage({
```
klaudia-blazyczek-blurify: postSystemMessage publishes immediately even when it receives this transaction. If the later update or the transaction commit fails, users can see a gift message that does not exist and was not charged. Publish only after the transaction commits.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,671 @@
...
+    const recipients = shuffleArray(allOnline.filter((id) => id !== actorId)).slice(
+      0,
+      input.recipientCount,
+    );
+    if (recipients.length === 0) {
+      throw new NoOnlineUsersError();
+    }
+    const { msg, currency } = await this.drizzle.db.transaction(async (tx) => {
+      const splitResult = await tx.execute(
+        sql`SELECT (floor(${input.amount}::numeric / ${recipients.length}))::text AS per_recipient`,
```
klaudia-blazyczek-blurify: This removes the remainder from the payout. For example, 10.99 sent to 10 users debits 10.99 but credits only 10, so 0.99 disappears. Please distribute the exact amount, including the remainder, or debit only the total that is credited.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,671 @@
...
+      avatarUrl: summary.avatarUrl,
+      level: summary.level,
+      joinedAt: summary.createdAt.toISOString(),
+      totalWagered: stats.totalWagered,
+      totalBets: stats.totalBets,
+      currency: summary.currency,
+    };
+  }
+
+  async executeCommand(input: ExecuteInput, actorId: Uuid): Promise<ChatSystemMessage> {
```
marek-chmielowski-blurify, line 247: None of `handleGift`, `handleRain`, or `handleDonate` accept an `idempotencyKey`. Compare to `wallet/contract/index.ts` (`DepositInputSchema`/`WithdrawInputSchema` both have `idempotencyKey: UuidSchema.optional()`) and `wallet.service.ts`'s `deposit()`, which namespaces + hashes the key and short-circuits a replay via a DB guard before ever touching the wallet balance. This module has no equivalent — a retried `POST /chat-command/execute` (client timeout + retry, or a double-tap before the button disables) debits the sender twice. This is exactly the case `db-conventions.md` calls out: "Money / critical paths are transactional and idempotent — a DB guard inside the transaction, not just an idempotencyKey." Suggest adding `idempotencyKey` to the gift/rain/donate branches of `ExecuteInputSchema` and a DB-level guard (e.g. a partial unique index on an idempotency column) mirroring the wallet module's pattern.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,671 @@
...
+      throw new ExceedsLimitError();
+    }
+    if (
+      config?.minAmount !== undefined &&
+      moneyToNumber(input.amount) < moneyToNumber(config.minAmount)
+    ) {
+      throw new BelowMinimumError();
+    }
+
+    const ids = await this.directory.findPlayerIds(input.targetUsername, 20);
```
marek-chmielowski-blurify: `targetUsername` here is already an exact, known value (typed in full or picked from autocomplete) — but it's resolved via `findPlayerIds`, which does `ILIKE '%query%'` against both `user.email` and `player.displayName` with **no ORDER BY** and an independent `LIMIT 20` on each sub-query (see `admin-user-directory.ts:150-163`). `AdminUserDirectory` has no exact-match method at all. Since `%query%` matches the substring anywhere (not just as a prefix), a short or common target username can collide with many unrelated accounts (e.g. `Anna` matches `Joanna_92`, `Marianna`, `anna.smith@...`, `AnnabellePlayer`, ...). If more than 20 rows match on either side and Postgres's unordered scan doesn't happen to include the actual target within the first 20, the exact user is silently dropped before `summaries.find(...)` ever runs, and a real, existing, online player gets `ChatPlayerNotFoundError` on `/donate`. Same pattern at line 607 in `handleBlockAction` for `/block` and `/ignore`. Since the caller already knows the exact username, suggest an exact-match lookup (`WHERE lower(display_name) = lower(\)`) instead of reusing the capped fuzzy search meant for autocomplete.

packages/core/src/engagement/chat-commands/contract/index.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  execute: oc
+    .route({ method: 'POST', path: '/chat-command/execute' })
+    .input(
+      z.discriminatedUnion('type', [
+        z.object({
+          type: z.literal('gift'),
+          amount: MoneyAmountSchema,
+          roomId: UuidSchema,
+          idempotencyKey: UuidSchema.optional(),
```
klaudia-blazyczek-blurify: The key is optional, so callers can still send a money command without any idempotency guard. A timeout followed by a normal retry will then debit again. Please require an idempotency key for every gift, rain, and donate request.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,845 @@
...
+          eq(chatCommandIdempotency.actorId, actorId),
+          eq(chatCommandIdempotency.commandType, commandType),
+          eq(chatCommandIdempotency.idempotencyKey, idempotencyKey),
+        ),
+      )
+      .limit(1);
+    if (!row) {
+      return null;
+    }
+    if (moneyToNumber(row.amount) !== moneyToNumber(amount)) {
```
klaudia-blazyczek-blurify: This checks only the amount. The same key and amount with a different room, recipient count, or donate target is treated as a replay and returns the old result. Store and compare a fingerprint of the complete command input.

packages/core/src/engagement/chat-commands/router/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  return os.router({
+    listCommands: os.listCommands.handler(() => svc.listCommands()),
+
+    execute: os.execute.handler(({ input, context }) => {
+      const actorId = getUserId(context);
+      return mapErrors(
+        {
+          NOT_FOUND: [CommandDisabledError, ChatPlayerNotFoundError],
+          CONFLICT: [
```
klaudia-blazyczek-blurify, line 38: The new idempotency errors are not mapped here. A reused key or a concurrent replay will become a 500 response instead of the intended conflict response. Please add both error classes to CONFLICT.

packages/core/src/pam/identity/admin-user-directory.ts
```diff
@@ -140,19 +144,47 @@ export class DrizzleAdminUserDirectory implements AdminUserDirectory {
...
+        email: user.email,
+        language: user.language,
+        avatarUrl: user.image,
+        createdAt: player.createdAt,
+        level: player.level,
+        currency: player.currency,
+      })
+      .from(player)
+      .innerJoin(user, eq(player.userId, user.id))
+      .where(ilike(player.displayName, username))
```
klaudia-blazyczek-blurify: This is not an exact lookup: ILIKE treats percent and underscore in the supplied username as wildcards. A donate command can then select an arbitrary matching player. Use case-insensitive equality, or escape the pattern before querying.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,853 @@
...
+
+    const configMax = config?.maxRecipients ?? 50;
+    if (input.recipientCount > configMax) {
+      throw new ExceedsLimitError();
+    }
+    const amountUnits = Math.floor(moneyToNumber(input.amount));
+    if (input.recipientCount > amountUnits) {
+      throw new TooManyRecipientsError();
+    }
+    const allOnline = await this.transport.getOnlineUserIds(chatChannel(input.roomId));
```
zaxovaiko, line 632: Replay lookup happens after online-user discovery and recipient randomization. Retrying a completed rain after recipients leave returns NoOnlineUsersError instead of the stored result. Please check idempotency before presence lookup and randomization.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,853 @@
...
+      const idempotencyRowId = await this.guardCommandIdempotency(
+        tx,
+        'rain',
+        actorId,
+        input.idempotencyKey,
+        input.amount,
+        fingerprint,
+      );
+
+      const splitResult = await tx.execute(
```
zaxovaiko: Fractional rain still destroys value: 10.99 divided among 10 recipients becomes 1 per recipient and debits only 10. Please handle the remainder explicitly, either by deterministic allocation or by rejecting/refunding it.

packages/core/src/engagement/chat-commands/router/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            SelfModerationActionError,
+            ChatCommandIdempotencyKeyReuseError,
+            ConcurrentCommandReplayError,
+          ],
+        },
+        () => svc.executeCommand(input, actorId),
+      );
+    }),
+
+    getGift: os.getGift.handler(({ input }) =>
```
zaxovaiko: Gift reads and claims only use the gift UUID; room visibility and membership are not enforced. Possession of a private-room gift identifier therefore exposes its data and permits claiming it. Please require authentication and membership for the gift's room.

packages/core/src/engagement/chat/plugin.ts
```diff
@@ -1,19 +1,39 @@
...
   id: 'chat',
+  dependsOn: ['identity'],
   register(ctx) {
+    ctx.provide(
+      CHAT_SERVICE,
+      (c) =>
+        new ChatService(
+          c.get(DRIZZLE),
+          c.get(EVENT_BUS),
+          c.get(CHAT_REALTIME_TRANSPORT),
```
zaxovaiko, line 31: The chat-scoped realtime tokens are consumed but never bound by the chat plugin or runtime defaults. Resolving chat routes can therefore fail at boot unless every consumer supplies both bindings. Please bind the documented defaults or fail with an explicit composition-root requirement.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,853 @@
...
+    const summaries = await this.directory.lookupPlayers(filteredIds);
+    return summaries.map((s) => ({
+      userId: s.userId,
+      username: s.username,
+      avatarUrl: s.avatarUrl,
+      level: s.level,
+    }));
+  }
+
+  async getPlayerProfile(userId: Uuid): Promise<PlayerProfileCard> {
```
zaxovaiko: Any authenticated user can retrieve another player's join date, currency, wager total, and bet count. Please enforce profile visibility or redact these fields to an explicitly public subset.

packages/core/src/contracts/adapters/wallet-commands.ts
```diff
@@ -10,7 +10,7 @@ export type WalletDebitArgs = {
 };
 
 export type WalletDebitOutcome =
-  | { ok: true; newBalance: string }
+  | { ok: true; newBalance: string; currency: string }
   | { ok: false; available: string };
 
 export type WalletCreditArgs = {
```
zaxovaiko, line 16: credit() accepts no expected currency, allowing a USD debit to nominally credit an EUR wallet. Please require matching currencies or make explicit FX/settlement part of the wallet command contract.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,853 @@
...
+        .where(eq(chatCommandIdempotency.id, idempotencyRowId));
+
+      return { msg: systemMsg, giftId: giftRow.id, currency: debit.currency };
+    });
+
+    // The caller now owns the commit boundary: postSystemMessage was passed `tx` above so
+    // it did not auto-publish - publish only now that this transaction has committed.
+    void this.transport.publish(chatChannel(input.roomId), msg);
+
+    await this.audit.record({
```
zaxovaiko: Audit persistence occurs after the money transaction. If audit.record() fails, the transfer commits unaudited and an idempotent replay returns before repairing the audit record. Please make audit persistence durable and idempotent through an outbox or transactional audit path.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -0,0 +1,853 @@
...
+      avatarUrl: summary.avatarUrl,
+      level: summary.level,
+      joinedAt: summary.createdAt.toISOString(),
+      totalWagered: stats.totalWagered,
+      totalBets: stats.totalBets,
+      currency: summary.currency,
+    };
+  }
+
+  async executeCommand(input: ExecuteInput, actorId: Uuid): Promise<ChatSystemMessage> {
```
zaxovaiko, line 247: Private-room membership is never checked. An authenticated user can target any private room by UUID and move money or publish commands there. Please authorize the caller through a chat-owned room-access seam before reading online users, moving money, or publishing.

packages/core/src/engagement/chat-commands/drizzle/migrations/0001_overjoyed_robin_chapel.sql
zaxovaiko, line 1: Let's squash migrations too

zaxovaiko (approved):

agniev-a-hub:
@zaxovaiko Could you please merge? I have pending req from Klaudia that is fixed, but I guess I need action from her. And she is unavailable.

PR #42, opened 2026-07-28
feat(casino): add admin game and player activity reporting

marek-chmielowski-blurify:
## Summary
- Adds admin-facing reporting for game performance (`AdminGameReporting`) and player activity (`admin-player-activity` adapter) across `casino/gaming`, `iam`, and `admin-console`
- New contracts/schemas (`admin-game-reporting.ts`, `admin-player-activity.ts`, `game.ts`) and a Drizzle migration for the supporting queries
- Wires the new adapters through `admin-console`'s router/service so the backoffice can consume game performance + player activity data (consumed by the betfeel `feat/bf-218-game-player-analytics` MR)
- Rebased onto `dev` to pick up the intervening stake-debit and cooling-off work in `casino/gaming`; resolved conflicts in `plugin.ts` (merged DI wiring), `contracts/schemas/common.ts` (both additions kept), and `gaming.service.test.ts` (unified the `GAME_ROW` fixture with the new `gameType` field)
- Regenerated `docs/catalog.json` to reflect current routes

## Test plan
- [x] `pnpm verify` (types, lint, format, boundaries, module-shape, deprecations, unit tests, tool tests) — 17/17 tasks green
- [x] `pnpm verify:drift` — catalog/openapi regeneration clean

PR #44, opened 2026-07-29
fix: resolve PR #42 review comments

marek-chmielowski-blurify:
## Summary

Resolves all 10 unresolved review comments left on #42 (merged without addressing them).

- Removed superfluous comments that only restated what `AGENTS.md`/naming already documents (`game.ts`, `admin-player-activity.ts`)
- Deduped `GAME_PERFORMANCE_SORT_FIELDS` between the contract port and admin-console (import from `@openora/core/contracts` instead of redeclaring)
- Added a `currency` filter to the game-performance report instead of summing bet/win amounts across currencies
- Rewrote DAU/WAU/MAU to a single `LEFT JOIN` + `FILTER` instead of three correlated subqueries per generated day
- Fixed a retention off-by-one: the registration-day session no longer counts as a return; the window now starts the day after registration
- Replaced hardcoded snake_case SQL identifiers with Drizzle column references in the retention query
- Added a `TODO` documenting that `winAmount` is never set yet (`endRound` has no way to report a win), so revenue currently always equals volume
- Rewrote the `DrizzleAdminGameReporting`, `DrizzleAdminPlayerActivity`, and `BackofficeService.getGamePerformance` test suites against real Postgres instead of mocked query builders/ports

## Test plan

- [x] `pnpm verify` (types, lint, boundaries, shape, deprecations, unit tests, drift) - green
- [x] New/rewritten real-Postgres tests cover: DAU/WAU/MAU correctness, retention window fix, currency filter, zero-activity edge cases

### DIFF
--- a/packages/core/src/admin-console/service/backoffice.service.ts
+++ b/packages/core/src/admin-console/service/backoffice.service.ts
@@ -168,6 +168,7 @@ export class BackofficeService {
       dateFrom: filter.dateFrom ? new Date(filter.dateFrom) : undefined,
       dateTo: filter.dateTo ? new Date(filter.dateTo) : undefined,
       gameType: filter.gameType,
+      currency: filter.currency,
       sortBy: filter.sortBy,
       sortDir: filter.sortDir,
     });
--- a/packages/core/src/casino/gaming/admin-reporting.ts
+++ b/packages/core/src/casino/gaming/admin-reporting.ts
@@ -14,20 +14,24 @@ export class DrizzleAdminGameReporting implements AdminGameReporting {
   async listGamePerformance(filter: GamePerformanceFilter): Promise<GamePerformanceRow[]> {
     const db = this.drizzle.db;
 
-    // status/date filters scope which ROUNDS count towards a game's metrics, so they
-    // live in the join condition, not WHERE - a game with zero completed rounds in
-    // range still appears with all-zero metrics rather than being dropped from the
+    // status/date/currency filters scope which ROUNDS count towards a game's metrics,
+    // so they live in the join condition, not WHERE - a game with zero completed rounds
+    // in range still appears with all-zero metrics rather than being dropped from the
     // report. `gameType` narrows which GAMES appear at all, so it stays in WHERE.
     const joinConditions = [
       eq(gameRound.gameId, game.id),
       eq(gameRound.status, 'completed'),
       filter.dateFrom ? gte(gameRound.startedAt, filter.dateFrom) : undefined,
       filter.dateTo ? lte(gameRound.startedAt, filter.dateTo) : undefined,
+      filter.currency ? eq(gameRound.currency, filter.currency) : undefined,
     ].filter(Boolean);
     const where = filter.gameType ? eq(game.gameType, filter.gameType) : undefined;
 
     const volume = sql<string>`coalesce(sum(${gameRound.betAmount}), 0)`;
     // GGR - can be negative when a game pays out more than it takes in over the range.
+    // TODO: winAmount is never set today - endRound has no way to report a win, so it
+    // stays at its schema default('0') and revenue always equals volume until round
+    // settlement actually computes payouts.
     const revenue = sql<string>`coalesce(sum(${gameRound.betAmount}) - sum(${gameRound.winAmount}), 0)`;
     const uniquePlayers = sql<number>`count(distinct ${gameRound.userId})`;
     const roundsPlayed = sql<number>`count(${gameRound.id})`;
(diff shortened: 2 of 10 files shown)


packages/core/src/admin-console/service/backoffice.service.ts
```diff
@@ -168,6 +168,7 @@ export class BackofficeService {
       dateFrom: filter.dateFrom ? new Date(filter.dateFrom) : undefined,
       dateTo: filter.dateTo ? new Date(filter.dateTo) : undefined,
       gameType: filter.gameType,
+      currency: filter.currency,
```
agniev-a-hub, line 171: Somewhere you check for undefined, somewhere you don't. Make sure it's not affecting the filtering logic.

packages/core/src/casino/gaming/admin-reporting.ts
```diff
@@ -14,20 +14,24 @@ export class DrizzleAdminGameReporting implements AdminGameReporting {
...
       filter.dateTo ? lte(gameRound.startedAt, filter.dateTo) : undefined,
+      filter.currency ? eq(gameRound.currency, filter.currency) : undefined,
     ].filter(Boolean);
     const where = filter.gameType ? eq(game.gameType, filter.gameType) : undefined;
 
     const volume = sql<string>`coalesce(sum(${gameRound.betAmount}), 0)`;
     // GGR - can be negative when a game pays out more than it takes in over the range.
+    // TODO: winAmount is never set today - endRound has no way to report a win, so it
+    // stays at its schema default('0') and revenue always equals volume until round
+    // settlement actually computes payouts.
```
agniev-a-hub: If it's another ticket's scope - just remove. TODO overall is bad :(

agniev-a-hub (dismissed):
Filtering is WARN.

packages/core/src/iam/adapters/admin-player-activity.ts
```diff
@@ -94,10 +102,10 @@ export class DrizzleAdminPlayerActivity implements AdminPlayerActivity {
       returned_users as (
         select distinct cu.user_id
         from cohort_users cu
-        inner join ${session} s
-          on s.user_id = cu.user_id
-          and s.updated_at >= cu.registered_at
-          and s.updated_at <= cu.registered_at + interval '1 day' * ${windowDays}
+        inner join ${session}
+          on ${session.userId} = cu.user_id
```
klaudia-blazyczek-blurify, line 105: Because `cu.registered_at` is a timestamp, `cu.registered_at + interval '1 day'` strictly requires 24 hours to pass. If a user registers at 11:00 PM and plays again at 9:00 AM the next calendar day, they won't be counted as retained. It is usually better to base retention on calendar days using `cu.cohort_date + interval '1 day'`.

packages/core/src/casino/gaming/admin-reporting.ts
```diff
@@ -14,20 +14,24 @@ export class DrizzleAdminGameReporting implements AdminGameReporting {
...
+    // status/date/currency filters scope which ROUNDS count towards a game's metrics,
+    // so they live in the join condition, not WHERE - a game with zero completed rounds
+    // in range still appears with all-zero metrics rather than being dropped from the
     // report. `gameType` narrows which GAMES appear at all, so it stays in WHERE.
     const joinConditions = [
       eq(gameRound.gameId, game.id),
       eq(gameRound.status, 'completed'),
       filter.dateFrom ? gte(gameRound.startedAt, filter.dateFrom) : undefined,
       filter.dateTo ? lte(gameRound.startedAt, filter.dateTo) : undefined,
+      filter.currency ? eq(gameRound.currency, filter.currency) : undefined,
```
klaudia-blazyczek-blurify, line 28: If `currency` is not provided in the filter, this query will sum `betAmount` and `winAmount` across all currencies into a single mixed-currency aggregate. Should the currency filter be required, or should the query group by currency if it's missing?

packages/core/src/casino/gaming/admin-reporting.ts
zaxovaiko: Why is TODO bad? 😄

packages/core/src/admin-console/service/backoffice.service.ts
marek-chmielowski-blurify: filter.currency is already `string | undefined` (same shape as dateFrom/dateTo), and it's checked downstream in the join condition (`filter.currency ? eq(gameRound.currency, filter.currency) : undefined`), so passing it through unconditionally here is safe - an undefined currency correctly results in no currency filter being applied. Leaving as-is.

packages/core/src/casino/gaming/admin-reporting.ts
marek-chmielowski-blurify: Intentional - when no currency filter is applied the report intentionally sums across currencies without FX conversion, and the operator-facing UI shows a disclaimer in that case so the number isn't misread as a converted total. Added a one-line comment in the code documenting this as a design decision.

packages/core/src/iam/adapters/admin-player-activity.ts
marek-chmielowski-blurify: Good catch - fixed. The join condition now uses cu.cohort_date + interval '1 day' instead of cu.registered_at + interval '1 day', so retention is based on calendar days rather than a strict 24h window. Also removed the now-unused registered_at column from the CTE and added a regression test covering exactly this case (registers 23:00, returns 09:00 the next calendar day).

klaudia-blazyczek-blurify (approved):

PR #45, opened 2026-07-30
feat(wallet): add runtime auto-withdrawal thresholds

marek-chmielowski-blurify:
## Summary

Moves global fiat and crypto auto-withdrawal thresholds from static platform configuration into a DB-backed wallet singleton. Adds Super Admin read/update routes, audit recording, safe seed defaults, migration coverage, and unit/e2e tests.

## Why

Introduced these changes so operators can update auto-withdrawal thresholds at runtime without a redeploy while preserving fail-closed payout behavior and a complete audit trail.

## Worth knowing

The new `wallet_auto_withdrawal_config` row is a global singleton for the current single-tenant runtime, so it intentionally has no `tenantId` or RLS policy. Both seeded thresholds default to zero, preventing an upgrade from silently enabling auto-approval; missing config also fails closed.

- [x] `pnpm check:drift` is green (catalog / OpenAPI not stale) - run `pnpm regen` if not
- [ ] New data tables carry `tenantId` and are RLS-covered (`pnpm regen` runs `gen:rls`)

Closes BF-211


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1047,14 +1062,54 @@ export class WalletService {
     if (rule) {
       return { value: rule.threshold, source: 'per-player' };
```
klaudia-blazyczek-blurify, line 1067: Blocking: when the singleton row is missing, this early return skips `getAutoWithdrawalConfig()`. A player with an override can still have a withdrawal auto-approved, so an unseeded install is not fail-closed. Check that the global row exists before accepting the per-player threshold.

packages/core/src/wallet/router/index.ts
```diff
@@ -173,6 +174,42 @@ export function createWalletRouter(
...
+        const {
+          userId: adminId,
+          ip,
+          userAgent,
+        } = await adminGuard.assert(context, 'auto-withdrawal-config', 'update');
+        // Non-throwing read: the row may legitimately be missing on an unseeded
+        // install, and `set` below self-heals it via upsert - a missing `before`
+        // must not block that self-heal.
+        const before = await wallet.getAutoWithdrawalConfigOrNull();
+        const config = await wallet.setAutoWithdrawalConfig(adminId, input);
```
klaudia-blazyczek-blurify: Blocking: this commits the threshold update before the audit write. If `audit.record()` fails, the API returns an error but the threshold remains changed without the required audit entry. Make the update and audit record atomic, or roll the update back when auditing fails.

packages/core/src/wallet/contract/index.ts
```diff
@@ -133,6 +133,21 @@ export const SetAutoWithdrawalRuleInputSchema = z.object({
...
+  fiatThreshold: MoneyAmountSchema,
+  cryptoThreshold: MoneyAmountSchema,
+  updatedBy: UuidSchema.nullable(),
+  updatedAt: TimestampSchema,
+  createdAt: TimestampSchema,
+});
+export type WalletAutoWithdrawalConfig = z.infer<typeof WalletAutoWithdrawalConfigSchema>;
+
+export const SetWalletAutoWithdrawalConfigInputSchema = z.object({
+  fiatThreshold: MoneyAmountSchema,
```
klaudia-blazyczek-blurify: This accepts unlimited integer digits, but the database column is `numeric(18,8)` and only allows 10 digits before the decimal. For example, `10000000000` passes this schema and then causes a database 500. Limit the input precision so invalid values return a 4xx response.

packages/core/src/wallet/service/wallet.service.ts
klaudia-blazyczek-blurify: I know this return already existed, but this PR adds the new fail-closed requirement based on the DB singleton. Because `getAutoWithdrawalConfig()` is called only after the early return, that requirement is skipped for players with a per-player rule.

klaudia-blazyczek-blurify (dismissed):

packages/core/src/wallet/service/wallet.service.ts
marek-chmielowski-blurify: Fixed in 56ec711: `resolveAutoThreshold` now reads the global `wallet_auto_withdrawal_config` singleton (via `getAutoWithdrawalConfig()`, which throws when the row is missing) BEFORE checking for a per-player `auto_withdrawal_rule` override. An unseeded install now fails closed for every player, including those with an override, not just the ones falling through to the global threshold. Added a regression test covering this exact scenario.

packages/core/src/wallet/router/index.ts
marek-chmielowski-blurify: Fixed in 56ec711: the before-read, upsert, and `AUDIT_WRITER.recordInTransaction` call now all run inside one `db.transaction` in `WalletService.setAutoWithdrawalConfig`. If the audit write throws, the whole transaction (including the threshold upsert) rolls back, so the config can never change without its audit entry. The router now just calls the service method and passes ip/userAgent through as `meta`; the atomicity moved into the service, matching the pattern already used in `chat-commands.service.ts`.

packages/core/src/wallet/contract/index.ts
marek-chmielowski-blurify: Fixed in 56ec711: added a module-local `WalletAutoWithdrawalThresholdSchema` (a `.refine()` on `MoneyAmountSchema`) that caps the integer part at 10 digits, matching `wallet_auto_withdrawal_config`'s `decimal(18,8)` column (18 total digits - 8 scale = 10 integer digits). `SetWalletAutoWithdrawalConfigInputSchema.fiatThreshold/cryptoThreshold` now use it, so `10000000000` is rejected as a 4xx at the contract boundary instead of overflowing the DB. Kept this local to wallet's contract rather than tightening the shared `MoneyAmountSchema` itself, since that schema also backs `decimal(18,2)` columns elsewhere (compliance, pam) with a different, wider integer-digit budget (16, not 10) - a blanket 10-digit cap there would be incorrect for those. Added a regression test.

zaxovaiko (approved):

PR #47, opened 2026-07-31
fix(iam): audit super-admin and grant-escalation denials

klaudia-blazyczek-blurify:
## Summary

`IamService.assertSuperAdmin` and the no-escalation check in `setRolePermissions` now emit `identity.user.unauthorized_access` before throwing, so these service-level denials are audited the same way `AdminGuard` denials always have been. BF-412.

## Why

Two service-level FORBIDDEN paths (`NotSuperAdminError`, `GrantEscalationError`) throw before ever reaching `AdminGuard.assert()`, so they bypassed the guard's audit emission and produced no `identity.user.unauthorized_access` row - even though the request genuinely reached the backend and was genuinely denied.

## Alternatives considered

An earlier version of this PR also added `AdminGuard.recordDeniedAccess` / an iam `reportAccessDenied` route so a frontend page guard could self-report a denial after blocking navigation client-side. Dropped per review: a client-side redirect never sends the underlying request, so it isn't a reliable audit signal - only a real backend request that hits `AdminGuard.assert()` (or one of the two service-level checks above) is. The frontend continues to redirect + show a toast for inaccessible pages, without a separate audit entry.

## Risks

None. Changeset downgraded to `patch` - no new public API, pure bug fix.


marek-chmielowski-blurify (dismissed):

marek-chmielowski-blurify (dismissed):

agniev-a-hub (dismissed):
As per slack

packages/core/src/server/auth/admin-guard.ts
```diff
@@ -105,43 +114,100 @@ export class AdminGuard {
...
+    }
+
+    this.emitUnauthorized(
+      caller.userId,
+      caller.role,
+      resource,
+      missingAction,
+      caller.ip,
+      caller.userAgent,
+    );
```
zaxovaiko: I strongly suggest to use named parameters

packages/core/src/server/auth/admin-guard.ts
```diff
@@ -105,43 +114,100 @@ export class AdminGuard {
...
+      (action) => !this.checkGrant(grants, userRole, resource, action),
+    );
+    if (!missingAction) {
+      return { recorded: false };
+    }
+
+    if (this.rateLimiter) {
+      const key = makeRateLimitKey(
+        RATE_LIMIT_KEYS.ACCESS_DENIED_REPORT,
+        `${caller.userId}:${resource}:${level}`,
```
zaxovaiko: Let's extract the key to some function (as we did before)

packages/core/src/iam/contract/index.ts
```diff
@@ -189,6 +193,11 @@ export const iamContract = {
   getMyPermissions: oc
     .route({ method: 'GET', path: '/iam/my-permissions' })
     .output(EffectivePermissionsSchema),
+
+  reportAccessDenied: oc
```
zaxovaiko: What is the purpose of this endpoint?

zaxovaiko (dismissed):

PR #60, opened 2026-08-07
fix(chat): adding metadata to system messages and more

agniev-a-hub:


packages/core/src/contracts/adapters/chat-system-writer.ts
```diff
@@ -10,6 +15,19 @@ export type ChatSystemWriter = {
     metadata: CommandMetadata;
     tx?: unknown;
   }): Promise<SystemChatMessage>;
+  /** @deprecated Use postSystemMessage; retained for adapter compatibility. */
```
marek-chmielowski-blurify: You added a method thats already deprecated?

marek-chmielowski-blurify (changes requested):

packages/core/src/contracts/adapters/chat-system-writer.ts
agniev-a-hub: Thank you for hint, removed it. Usually you save backwards compatabylity for WIP features, but this particular one is redundant.

marek-chmielowski-blurify (approved):

PR #64, opened 2026-08-11
fix(audit): resolve player resource ids

marek-chmielowski-blurify:
## Summary

Fix audit records so player-scoped events store the PAM player ID as resourceId. Update integration and E2E coverage for KYC, responsible gambling, and player-level audit flows. BF-335.

## Why

Player audit events previously persisted the identity user ID while declaring resourceType as player. Admin log and export filters using the actual player ID therefore missed those records.

## Alternatives considered

Keep using the identity user ID as resourceId, but that conflicts with the declared player resource type and the identifiers used by player-facing consumers.

## Risks

No schema migration. When an event references a user without a materialized player row, resourceId remains null.

### DIFF
--- a/packages/core/src/audit/plugin.ts
+++ b/packages/core/src/audit/plugin.ts
@@ -10,7 +10,11 @@ function isRecord(v: unknown): v is Record<string, unknown> {
   return typeof v === 'object' && v !== null;
 }
 
-function mapEventToRecord(topic: string, p: Record<string, unknown>): RecordInput {
+export async function mapEventToRecord(
+  topic: string,
+  p: Record<string, unknown>,
+  resolvePlayerId: (userId: string | null) => Promise<string | null>,
+): Promise<RecordInput> {
   const result = /\.(failed|rejected|declined)$/.test(topic) ? 'failure' : 'success';
 
   const base: RecordInput = {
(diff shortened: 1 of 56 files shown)


packages/core/src/audit/plugin.ts
```diff
@@ -48,7 +52,7 @@ function mapEventToRecord(topic: string, p: Record<string, unknown>): RecordInpu
       actorType: 'player',
       actorId: str(p['userId']),
       resourceType: 'player',
-      resourceId: str(p['userId']),
+      resourceId: await resolvePlayerId(str(p['userId'])),
```
agniev-a-hub: The whole idea of making one more DB call from audit to later record to the audit is strange... Please elaborate why we don't pass that data on the service layer of callee through event body? 
Audit module himself must be standalone.

agniev-a-hub (changes requested):
I got it now. 

You want to write the different data - instead of userId to write playerId. 

As I explained offline (or on meet, don't remember), user is our core entity representing a person. 

User is used for chat / payment / auth etc. Audit is low-level data storage. 

To implement this fix, the idea was to GET playerId from BE and then redirect. Another option - as we have it now - to just paste userId in the search bar. 

But writing playerId as default level entity for audit is strange... Please elaborate on necessarity.

packages/core/src/audit/plugin.ts
marek-chmielowski-blurify: To be discussed offline.

agniev-a-hub (approved):

PR #65, opened 2026-08-12
feat(social): add send friend request

marek-chmielowski-blurify:
## Summary

Adds a new `engagement/social` module with `sendFriendRequest` and `getRelationships` routes, the backend foundation for the Friends & Social epic's friend-request flow.

## Why

Players need to be able to send a friend request to another player, see it reflected as "Pending" immediately, have the recipient notified in-app, and be blocked from re-sending duplicates or requesting a player who has blocked them. This is the first slice of the Friends & Social epic; accept/decline, friends list, and block/ignore management land in separate follow-up tickets.

## Alternatives considered

- Reusing the existing `chat_user_block` table for the "recipient has blocked sender" gate - rejected, it's a chat-only message mute (confirmed by its own code comments), not the social-graph-level block this AC needs. A new `social_user_block` table ships instead, read-only in this PR.
- Treating a mutual/simultaneous request (both players request each other) as a conflict error - product chose auto-accept instead, so that case flips the existing pending row straight to accepted.

## Risks

- `social_user_block` has no write endpoint yet - it's schema-only until the block/ignore management ticket lands, so the "blocked" gate can never actually trigger until then.
- Friendship rows for suspended/closed/deactivated players are filtered on read rather than cascade-deleted, per this repo's "never hard-delete" convention - this differs from the literal wording in the design doc, which described a cascade delete.
- No rate limit on `sendFriendRequest` yet; worth a follow-up if spam becomes a concern.


packages/core/src/engagement/social/drizzle/migrations/0000_fine_jamie_braddock.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TYPE "public"."friendship_status" AS ENUM('pending', 'accepted');--> statement-breakpoint
+CREATE TABLE "friendship" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"requester_id" uuid NOT NULL,
+	"addressee_id" uuid NOT NULL,
+	"status" "friendship_status" DEFAULT 'pending' NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"responded_at" timestamp with time zone
```
agniev-a-hub, line 9: I'd do here NULL or Date for accepted_at and refused_at instead of STATUS. Also, you don't have a scenario when it's not accepted. 

accepted_at NULL and refused_at Date means it's refused. Both NULL - pending. 

Please adjust.

packages/core/src/engagement/social/drizzle/migrations/0000_fine_jamie_braddock.sql
```diff
@@ -0,0 +1,24 @@
...
+);
+--> statement-breakpoint
+CREATE TABLE "social_user_block" (
```
agniev-a-hub, line 12: Image

We have two tables here - chat_user_block and chat_user_ignore. Naming is not the best, and to be changed to common user_ignore and user_block, but table reference will stay same. Please use those tables, since it's same functionality as chat now has - block and ignore.

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    // Both directions checked in one query: at most the two rows for this pair exist.
+    const blocks = await this.drizzle.db
+      .select({ blockerId: socialUserBlock.blockerId, blockedId: socialUserBlock.blockedId })
+      .from(socialUserBlock)
+      .where(and(isNull(socialUserBlock.removedAt), pairBlockCondition(callerId, targetUserId)));
+    if (blocks.some((b) => b.blockerId === targetUserId && b.blockedId === callerId)) {
+      // Target has blocked the caller - never disclosed, same error as "doesn't exist".
+      throw new FriendRequestTargetNotFoundError(targetUserId);
```
agniev-a-hub: Strange, please verify if it's direct requirement by Confluence. 

It's not right to not let user know why we don't let him to send that request. He sees user, tries to invite him -> doesn't exist.

agniev-a-hub (approved):

packages/core/src/engagement/social/drizzle/migrations/0000_fine_jamie_braddock.sql
marek-chmielowski-blurify: I will change this in the next pr. It was not a good idea to include this code here. I will remove all blocking/unblocking from this pr and indlude in the next one.

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: This wont happen, because sender should not see the button to add them in the first place. No button -> no request. Treat this code as safety net for illegal operations

> "Add Friend" option not visible if the recipient has blocked the sender

agniev-a-hub: It's not the point of current client functionality, but API design. 

openora is something not only to be used by many consumers. And some of those might want to have that.

packages/core/src/engagement/social/drizzle/migrations/0000_fine_jamie_braddock.sql
agniev-a-hub: Super, ty.

agniev-a-hub (dismissed):
After code transfer - feel free to merge.

agniev-a-hub (approved):

PR #66, opened 2026-08-13
feat(chat): chat moderation implementation

agniev-a-hub:


packages/core/src/engagement/chat/service/chat.service.ts
marek-chmielowski-blurify, line 1: This service is already very long. How about we move this to a separate service for moderation only?

marek-chmielowski-blurify (dismissed):

marek-chmielowski-blurify (dismissed):

packages/core/src/contracts/adapters/realtime.ts
```diff
@@ -43,13 +43,21 @@ export type RealtimeTransport = {
...
    */
   publish<T>(channel: string, event: T): void | Promise<void>;
+  /**
+   * Fan out a deletion/tombstone event to the channel. The transport does not
+   * remove persisted data; callers must complete the database mutation first.
+   * SSE and managed transports deliver this through their normal live stream.
+   */
+  remove<T>(channel: string, event: T): void | Promise<void>;
   /**
    * Subscribe a handler to a channel. Returns an unsubscribe fn the caller MUST
```
klaudia-blazyczek-blurify: Please keep this backward compatible. Existing realtime adapters do not implement `remove`, so this makes the public port breaking. Make it optional with a fallback, or use `publish` for the tombstone.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -0,0 +1,231 @@
...
+        .from(chatRoomMember)
+        .where(and(eq(chatRoomMember.roomId, roomId), eq(chatRoomMember.userId, moderatorId)))
+        .limit(1);
+      if (!moderator) {
+        throw new ChatRoomNotMemberError(roomId);
+      }
+      if (moderator.role !== 'moderator' && moderator.role !== 'owner') {
+        throw new ChatRoomNotModeratorError(roomId);
+      }
+      const removed = await t
```
klaudia-blazyczek-blurify, line 209: A moderator can remove an owner or another moderator here because the target role is never checked. Please allow a non-owner moderator to remove only a `member`, as the ban and mute paths do.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -0,0 +1,231 @@
...
+      });
+      this.events.emit('chat.room.member.removed', {
+        roomId,
+        userId,
+        removedBy: moderatorId,
+        ip: ip ?? null,
+        userAgent: userAgent ?? null,
+      });
+    }
+    await this.transport?.revokeClient?.(userId);
```
klaudia-blazyczek-blurify: The default transport does not implement `revokeClient`, so this removed member’s open private-room stream stays subscribed until disconnect. Please revoke the room channel with `revokeClientFromChannel`.

packages/core/src/engagement/chat/service/chat-ban.service.ts
```diff
@@ -0,0 +1,138 @@
...
+    durationSeconds: number | null;
+    reason: string;
+    actorId: Uuid;
+  } & ClientMeta) {
+    const scope =
+      roomId === '__global' || roomId === '__all_public' || roomId === '__all' ? roomId : 'room';
+    const concreteRoomId = scope === 'room' ? roomId : null;
+    const expiresAt =
+      durationSeconds === null ? null : new Date(Date.now() + durationSeconds * 1000);
+    const [created] = await this.drizzle.db
```
klaudia-blazyczek-blurify: Please validate the room before inserting. A missing UUID becomes an unhandled foreign-key error, and a private room bypasses the public-room guard used by mutes.

packages/core/src/engagement/chat/service/chat-ban.service.ts
```diff
@@ -0,0 +1,138 @@
...
+      .insert(chatPlatformBan)
+      .values({ userId, bannedBy: actorId, roomId: concreteRoomId, scope, reason, expiresAt })
+      .onConflictDoNothing()
```
klaudia-blazyczek-blurify: An expired ban still has `liftedAt = null`, so this returns success after a conflict but does not apply the new ban. Please close expired rows transactionally before inserting; scope-specific bans also need a scope-aware key.

packages/core/src/engagement/chat/service/chat-room-mute.service.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+    await this.assertModerator(roomId, moderatorId, userId);
+    const [active] = await this.drizzle.db
+      .select({ id: chatRoomMute.id })
+      .from(chatRoomMute)
+      .where(
+        and(
+          eq(chatRoomMute.roomId, roomId),
+          eq(chatRoomMute.userId, userId),
+          isNull(chatRoomMute.liftedAt),
```
klaudia-blazyczek-blurify: An expired mute is still found because this query checks only `liftedAt`. Re-muting returns success without creating an effective mute. Please treat expired rows as inactive and replace them atomically.

packages/core/src/engagement/chat/service/chat-mute.service.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export class ChatMuteService {
+  constructor(
+    private readonly drizzle: DrizzleService,
+    private readonly audit: AuditWritePort,
+  ) {}
+
+  async assertCanSend(userId: Uuid, roomId: Uuid | null, isPublic = true) {
+    const now = new Date();
+    if (roomId !== null) {
```
klaudia-blazyczek-blurify, line 50: Global chat passes `roomId: null`, so this branch never checks its configuration. Players can still send messages after global read-only mode is enabled. Please resolve and check the global room configuration too.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
klaudia-blazyczek-blurify: At least I think it should work like that, but we can decide on the approach

packages/core/src/engagement/chat/service/chat-mute.service.ts
klaudia-blazyczek-blurify: Unless we don't plan on a read-only mode for global chat

klaudia-blazyczek-blurify (changes requested):

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
agniev-a-hub: Addressed

packages/core/src/engagement/chat/service/chat-mute.service.ts
agniev-a-hub: Yup, user which is not logged in cannot use chat features or send messages. In case of ban - we don't give access to the user with respecting info sent from the BE;

agniev-a-hub:
@klaudia-blazyczek-blurify Issuees have been addressed, please check :)

packages/core/src/engagement/chat-commands/schema/index.ts
```diff
@@ -1,13 +1,4 @@
...
-  jsonb,
-  timestamp,
-  uuid,
-  unique,
-  decimal,
-} from 'drizzle-orm/pg-core';
+import { pgTable, text, boolean, jsonb, timestamp, uuid, unique } from 'drizzle-orm/pg-core';
 import type { CommandConfig } from '../contract/index.js';
 
 export const chatCommandConfig = pgTable(
```
zaxovaiko, line 4: Hm, I thought we'd keep it as non-shared functionality. Is there any good reason why it's here, in openora?

agniev-a-hub: Brought offline.

klaudia-blazyczek-blurify (approved):

PR #68, opened 2026-08-17
feat(social): list and remove friends, dissolve friendship on block

marek-chmielowski-blurify:
## Summary

BF-427: adds the Friends tab backend surface - `listFriends` (online/offline status + the caller's own ignore state) and `removeFriend`, plus automatic friendship dissolution when either party blocks the other.

## Why

The Friends tab needs a paginated friends list with presence and a way to remove a friend. Blocking a user (existing chat feature) must also dissolve any active friendship per BF-427's AC, and player online/offline status had no writer for `player.lastSeenAt` yet.

- `social.listFriends` (GET `/social/friends`): accepted, non-removed friendships with displayName, derived online/offline status, and the caller's own `isIgnored` flag (drives the Ignore/Unignore menu state).
- `social.removeFriend` (DELETE `/social/friends/{targetUserId}`): soft-dissolves an active friendship, emits `social.friendship.removed`.
- `social/plugin.ts` subscribes to `chat.user.blocked` and dissolves any active friendship for the pair via the same soft-delete path.
- New `PLAYER_ACTIVITY_TRACKER` port, bound in player-management: throttled (1/min) writer of `player.lastSeenAt` called fire-and-forget from the request auth middleware - first writer of that column.
- `friendship` schema: `removedAt` column, partial unique pair index (`removedAt IS NULL`) so re-friending after a removal works, plus a check that a removal requires a prior acceptance.
- Audit mapping for `social.friendship.removed`.

Avatar, rank badge, and an "Away" presence state from the AC are intentionally out of scope - no avatar or rank/loyalty module exists yet in the platform; only online/offline is derived for now.

## Alternatives considered

None.

## Risks

None. Additive migration (new nullable column + index swap, no data migration). `chat.user.blocked` handling never throws into the event bus (errors are logged only). `notifications` module does not subscribe to `social.friendship.removed` or `chat.user.blocked`, so no notification is sent on remove/block, per AC. Verified with `pnpm verify` (typecheck, lint, format, boundaries, module-shape, unit + integration tests, catalog drift) - all green.


packages/core/src/engagement/social/plugin.ts
```diff
@@ -1,19 +1,43 @@
...
+    // reuses the chat module's existing block infra rather than duplicating it. Never
+    // throws into the event bus dispatch: a failure here is logged, not propagated.
+    ctx.events.on('chat.user.blocked', (payload) => {
+      const parsed = domainEventSchemas['chat.user.blocked'].safeParse(payload);
+      if (!parsed.success || !socialRef) {
+        return;
+      }
+      const p = parsed.data;
+      socialRef
+        .dissolveFriendshipOnBlock(p.blockerId, p.blockedId)
```
klaudia-blazyczek-blurify: Please make this required privacy update durable. This handler only logs a failure, so a crash or database error after a successful block can leave the friendship active and return the blocked user from `/social/friends`. Use an atomic or retryable path.

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -332,6 +349,148 @@ export class SocialService {
...
+          isNotNull(friendship.acceptedAt),
+        ),
+      )
+      .returning();
+    const removed = rows[0];
+    if (!removed) {
+      return undefined;
+    }
+
+    this.events.emit('social.friendship.removed', {
```
klaudia-blazyczek-blurify: Please persist the removal event with the friendship update. `emit()` is best-effort, so a publish failure after this update removes the row without the audit event. Use `emitInTransaction()` with the transaction and outbox.

packages/core/src/engagement/social/plugin.ts
klaudia-blazyczek-blurify: This event can be delayed until after the users unblock and become friends again. It then removes the new friendship. Please make the block and dissolve operation synchronous and transactional instead of acting on a later event.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/pam/player-management/service/player-activity-tracker.service.ts
```diff
@@ -0,0 +1,20 @@
+import { DrizzleService } from '@openora/core/server';
+import type { Player, PlayerActivityTracker } from '@openora/core/contracts';
+import { and, eq, isNull, lt, or, sql } from 'drizzle-orm';
+import { player } from '@openora/core/pam/schema/profile';
+
+export class DrizzlePlayerActivityTracker implements PlayerActivityTracker {
```
zaxovaiko: Do we really need this additional layer?

zaxovaiko:
You have too many unnecessary comments in the code. I'd clean up the code, make it readable, and reduce the tokens agents use.

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -332,6 +339,144 @@ export class SocialService {
...
+      otherUserIds.length > 0
+        ? await Promise.all([
+            this.drizzle.db
+              .select({
+                userId: player.userId,
+                displayName: player.displayName,
+                lastSeenAt: player.lastSeenAt,
+              })
+              .from(player)
+              .where(inArray(player.userId, otherUserIds)),
```
klaudia-blazyczek-blurify, line 448: Please exclude `suspended` and `closed` players here, as `getRelationships()` does. Otherwise an existing friend can still see their profile and presence after the account becomes unavailable.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: Fixed

packages/core/src/pam/player-management/service/player-activity-tracker.service.ts
marek-chmielowski-blurify: It depends on how we define online/offline. lastSeenAt tracks activity across the whole app, not just chat. If chat presence is enough for the Friends tab, we can reuse the existing chat presence mechanism. Do we need chat-only or app-wide activity?

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: Fixed in 1798be11 - `listFriends` now excludes players with `status: 'suspended' | 'closed'`, matching `getRelationships()`. Added a regression test for both statuses.

marek-chmielowski-blurify: fixed

packages/core/src/pam/player-management/service/player-activity-tracker.service.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(private readonly drizzle: DrizzleService) {}
+
+  async touchLastSeen(userId: Player['userId']): Promise<void> {
+    await this.drizzle.db
+      .update(player)
+      .set({ lastSeenAt: new Date() })
```
klaudia-blazyczek-blurify: It's a minor detail, but you are mixing application time (`new Date()`) with database time (`sql\`now() - interval '1 minute'\``). It might be slightly safer to use `sql\`now()\`` for both, or `new Date()` for both to avoid clock drift issues between the app server and the database.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -522,8 +529,14 @@ export class ChatService {
         .offset(pageToOffset(page, limit)),
       this.drizzle.db.select({ n: count() }).from(chatUserBlock).where(where),
     ]);
+    const usernameByUserId = await this.usernamesFor(rows.map((r) => r.blockedId));
```
klaudia-blazyczek-blurify: Consider adding an early return or skipping the query if `rows.length === 0`. It could save an unnecessary database query.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -650,8 +672,14 @@ export class ChatService {
         .offset(pageToOffset(page, limit)),
       this.drizzle.db.select({ n: count() }).from(chatUserIgnore).where(where),
     ]);
+    const usernameByUserId = await this.usernamesFor(rows.map((r) => r.ignoredId));
```
klaudia-blazyczek-blurify: Similarly here, consider checking if `rows.length` is zero to skip the `usernamesFor` call.

klaudia-blazyczek-blurify (dismissed):

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -332,6 +339,148 @@ export class SocialService {
...
+          isNotNull(friendship.acceptedAt),
+        ),
+      )
+      .returning();
+    const removed = rows[0];
+    if (!removed) {
+      return undefined;
+    }
+
+    this.events.emit('social.friendship.removed', {
```
klaudia-blazyczek-blurify: You are emitting a domain event (`this.events.emit`) directly inside a database transaction (`tx` passed from `chat.service.ts`). If the transaction rolls back after this point, the event will have already been broadcasted, leading to "ghost" events in other modules

packages/core/src/pam/player-management/service/player-activity-tracker.service.ts
zaxovaiko: I meant that we don't really need this additional `DrizzlePlayerActivityTracker` 😄

agniev-a-hub (approved):

agniev-a-hub (dismissed):
emitting event is the only one blocking issue from my side, but once you fix it GTGT IMO

zaxovaiko (approved):

PR #69, opened 2026-08-17
feat(pam): expose and search by player id on the admin directory

klaudia-blazyczek-blurify:
## Summary

Adds `playerId` to the admin user-directory port and to the Transaction Monitor / Withdrawal Queue admin list contracts, and fixes the player search (`findPlayerIds`) to also match an exact `playerId`/`userId`, not just email/displayName.

## Why

PAM is playerId-native while wallet and admin-console are userId-native, so the same person shows a different UUID depending on which admin list is open, and pasting a UUID into the player search box previously returned nothing because `findPlayerIds` only matched `user.email` and `player.displayName` substrings.

## Alternatives considered

Bolting an id-resolution overlay onto the downstream consumer instead of fixing the port here - rejected because both list endpoints already join through the same `AdminUserDirectory.lookupPlayers`/`findPlayerIds` port, so extending the port adds zero new queries and keeps every consumer playerId-consistent at the source rather than duplicating resolution logic per app.

Substring (`ILIKE '%term%'`) matching for the new id search branches, mirroring the existing email/displayName behaviour - rejected because a UUID substring match is not meaningful (a fragment of one UUID can coincidentally appear in another) and would force a sequential text-cast scan; used exact match instead, consistent with the existing precedent in `PlayerService.list`.

## Risks

`AdminPlayerSummary.playerId` is a required field on a port type (`AdminUserDirectory`) that downstream repos can implement, so any external implementation missing it will fail to typecheck until updated - shipped as a `minor` changeset rather than `patch` to signal this.

No schema or migration changes (the column already existed; this only adds it to existing select projections). No change to existing email/displayName search behaviour - the two new id branches are additive.

### DIFF
--- a/packages/core/src/pam/identity/admin-user-directory.ts
+++ b/packages/core/src/pam/identity/admin-user-directory.ts
@@ -2,7 +2,7 @@ import type { AdminUserDirectory, AdminUserListOptions, ClientMeta } from '@open
 import { KycStatusSchema, normalizeKycStatus } from '@openora/core/contracts';
 import { DrizzleService, pageToOffset } from '@openora/core/server';
 import type { EventBus } from '@openora/core/server';
-import { asc, count, desc, eq, ilike, inArray, sql } from 'drizzle-orm';
+import { asc, count, desc, eq, ilike, inArray, or, sql } from 'drizzle-orm';
 import { user } from './schema/index.js';
 // Read-only cross-domain read of the player/profile table via the public /schema
 // subpath (allowed per ADR-0020) so back-office lists can label players by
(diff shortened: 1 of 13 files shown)


packages/core/src/pam/identity/admin-user-directory.ts
```diff
@@ -176,7 +178,7 @@ export class DrizzleAdminUserDirectory implements AdminUserDirectory {
   // crutch - it is effectively unreachable for a real admin search term.
   async findPlayerIds(query: string, limit = 1000) {
     const term = `%${query}%`;
-    const [byEmail, byName] = await Promise.all([
+    const [byEmail, byName, byPlayerId, byUserId] = await Promise.all([
```
zaxovaiko: Seems like we need to do the aggregation on db level

packages/core/src/pam/identity/admin-user-directory.ts
```diff
@@ -172,29 +174,24 @@ export class DrizzleAdminUserDirectory implements AdminUserDirectory {
...
-    return [...ids].slice(0, limit);
+    const rows = await this.drizzle.db
+      .selectDistinct({ id: user.id })
+      .from(user)
+      .leftJoin(player, eq(player.userId, user.id))
+      .where(
+        or(
+          ilike(user.email, term),
+          ilike(player.displayName, term),
+          ilike(sql`${player.id}::text`, query),
```
marek-chmielowski-blurify: Casting to ::text here means this branch cant use the PK index, so it's a seq scan. And if all you need here is the exact match on id, maybe just compare it directly by =.

packages/core/src/pam/identity/admin-user-directory.ts
```diff
@@ -174,45 +174,24 @@ export class DrizzleAdminUserDirectory implements AdminUserDirectory {
   }
 
   // Substring search is index-backed by the pg_trgm GIN indexes on user.email and
-  // player.display_name; the 1000 cap is a safety bound on the id set, not a perf
-  // crutch - it is effectively unreachable for a real admin search term.
+  // player.display_name; the id branches are exact matches on the ::text cast, not
+  // substring, so a UUID fragment can't coincidentally match. One query, deduped and
+  // capped by Postgres via DISTINCT + LIMIT, rather than four round trips merged in JS.
```
zaxovaiko: Let's not add obvious comments. I know that agents add them, but let's maybe adjust instructions or something

zaxovaiko (approved):

PR #70, opened 2026-08-17
feat(wallet): add wallet_balance table for per-currency balances

zaxovaiko:
## Summary
- Adds `wallet_balance (walletId, currency, amount)` with a unique index on `(walletId, currency)` and an FK to `wallet`.
- Purely additive: nothing reads or writes it yet, so behaviour is unchanged.

## Why
BF-213: the wallet is crypto-first and the requirements assume a multi-currency balance ("underlying balances and transactions are always stored in their original currency"), but `wallet` has a single `balance` column with a single `currency`. A player depositing BTC and then ETH today has both amounts added into the same number, which mixes units.

Bottom of a 4-PR stack. This one only creates the table so the follow-ups stay small: #71 moves the reads and writes onto it, #72 lifts the single-currency restriction, #73 is the deposit-address side.

## Alternatives considered
Keeping `wallet.balance` and adding a `currency` discriminator to it - that is the same single-row problem with extra steps. A player needs several balances at once, so it has to be a separate table with the pair unique.

## Risks
No backfill is included. There is no production data yet, so a wallet created before this PR simply has no `wallet_balance` row until #71 credits it. If that changes before rollout it needs a one-off data script, kept separate from the code change.


packages/core/src/wallet/schema/index.ts
```diff
@@ -42,6 +42,24 @@ export const wallet = pgTable('wallet', {
...
 
+export const walletBalance = pgTable(
+  'wallet_balance',
+  {
+    id: uuid().primaryKey().defaultRandom(),
+    walletId: uuid()
+      .notNull()
+      .references(() => wallet.id),
+    currency: text().notNull(),
+    amount: decimal({ precision: 18, scale: 8 }).notNull().default('0'),
```
klaudia-blazyczek-blurify: This new per-currency balance must preserve 18-decimal assets such as ETH. `numeric(18, 8)` cannot store `0.000000000000000001`, so a valid balance cannot round-trip. Use a crypto-safe precision/scale and regenerate the migration; update the money contract before this table is written.

marek-chmielowski-blurify (dismissed):

packages/core/src/wallet/schema/index.ts
```diff
@@ -42,6 +42,24 @@ export const wallet = pgTable('wallet', {
     .$onUpdateFn(() => new Date()),
 });
 
+export const walletBalance = pgTable(
+  'wallet_balance',
+  {
+    id: uuid().primaryKey().defaultRandom(),
```
agniev-a-hub, line 50: True, I was not fan of that UuidSchema anyway :)

packages/core/src/wallet/schema/index.ts
```diff
@@ -42,6 +42,24 @@ export const wallet = pgTable('wallet', {
...
+    walletId: uuid()
+      .notNull()
+      .references(() => wallet.id),
+    currency: text().notNull(),
```
agniev-a-hub, line 54: Heads up for currency as text - intended?

damianrzepka (approved):

damianrzepka (approved):

packages/core/src/wallet/schema/index.ts
zaxovaiko: Yup. Currencies are quite dynamic values in our case, so it's better to keep them as a string for now, not an enum.

PR #71, opened 2026-08-17
refactor(wallet): store balances in wallet_balance keyed by currency

zaxovaiko:
## Summary
- Adds `creditWalletBalance`, `debitWalletBalance` and `readWalletBalance` to `wallet.service.ts` and moves all seven balance mutation sites onto them: deposit, withdrawal hold, PSP-failure refund, admin-reject refund, webhook deposit credit, and `WalletCommandsService.debit`/`credit`.
- `debitWalletBalance` keeps the same atomic guard the old code had, now as `WHERE walletId AND currency AND amount >= x` - a lost race still updates zero rows and reports the shortfall.
- `creditWalletBalance` is an upsert, so the balance row is created on first credit and the arithmetic stays in Postgres numeric.
- Test seeds and the demo seeder now write `wallet_balance`; `wallet.balance` is still present but no longer read or written by anything.

## Why
Follow-up to #70. Behaviour is deliberately unchanged here - `CurrencyMismatchError` still rejects a deposit or withdrawal in a currency other than the wallet's, and no test assertion about balances changed. Splitting the storage move from the product change keeps each PR revertible on its own; lifting the single-currency restriction is #72.

## Alternatives considered
Dual-writing `wallet.balance` alongside `wallet_balance` during a transition. Rejected - two sources of truth for money is exactly the thing worth avoiding, and there is no production data that needs the safety net.

## Risks
`wallet.balance` is now dead weight: written as `'0'` on wallet creation and never updated. Anything outside this repo reading that column directly would silently see zero. Grepped the monorepo and only the wallet service touched it. It is dropped in #72.

### DIFF
--- a/packages/core/src/wallet/service/wallet.service.ts
+++ b/packages/core/src/wallet/service/wallet.service.ts
@@ -37,6 +37,7 @@ import { eq, asc, desc, sql, and, gte, lte, count, inArray } from 'drizzle-orm';
 import { createHash } from 'node:crypto';
 import {
   wallet,
+  walletBalance,
   walletTransaction,
   autoWithdrawalRule,
   walletAutoWithdrawalConfig,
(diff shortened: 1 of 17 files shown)


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -115,6 +116,56 @@ export function railFor(currency: string, cryptoCurrencies?: readonly string[]):
...
+      },
+    })
+    .returning({ amount: walletBalance.amount });
+}
+
+export function debitWalletBalance(
+  txn: DrizzleDb,
+  walletId: Wallet['id'],
+  currency: string,
+  amount: string,
```
klaudia-blazyczek-blurify, line 148: `currency` accepts any letter case, but this lookup is case-sensitive. A request with `usd` for a wallet stored as `USD` will miss the balance on debit or create a second balance row on credit. Normalize the balance key or use the wallet currency consistently.

klaudia-blazyczek-blurify (changes requested):

marek-chmielowski-blurify (dismissed):

klaudia-blazyczek-blurify (dismissed):

damianrzepka (dismissed):

damianrzepka (approved):

PR #72, opened 2026-08-17
feat(wallet): multi-currency balances with a per-player active currency

zaxovaiko:
## Summary
- Drops the now-unused `wallet.balance` column; `wallet.currency` becomes the player's active currency rather than the only currency they can hold.
- Removes `CurrencyMismatchError` from the deposit and withdrawal paths. A deposit in a new currency credits its own `wallet_balance` row and leaves the others untouched; a withdrawal in a currency with no balance now fails with `InsufficientBalanceError`, which is the honest error and fails closed.
- Adds `GET /wallet/balances` returning `{ activeCurrency, balances: [{ currency, balance }] }` and `PUT /wallet/active-currency`.
- `GET /wallet/balance` is unchanged and still returns the active currency's balance, so existing clients keep working.

## Why
BF-213: per Confluence, balances are multi-currency and stored in their original currency, and cross-asset conversion is a separate out-of-scope `Swap` feature. The single-currency guard was blocking that outright - a player could only ever hold one asset.

Keeping balances per asset also means we never owe EUR while holding BTC, so there is no unhedged position between deposit and withdrawal.

## Alternatives considered
Renaming `wallet.currency` to `activeCurrency`. Drizzle would emit a drop plus add rather than a rename, and the column already means "the currency this player plays in", so it stayed as is.

Changing `GET /wallet/balance` to return the array directly. That breaks every existing caller for no gain, so the array lives on a new route instead.

## Risks
`setActiveCurrency` does not validate the currency against a supported list or require an existing balance - a player can switch to a currency they hold nothing in, and the next bet simply fails on balance. Worth tightening once there is a canonical supported-currency list.

`WalletCommandsService.credit` still rejects a currency other than the active one, so gameplay stays single-currency by design. That matches "a player plays in one currency at a time".


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -341,6 +341,35 @@ export class WalletService {
     };
   }
 
+  async getBalances(userId: User['id']) {
+    const [record] = await this.drizzle.db.select().from(wallet).where(eq(wallet.userId, userId));
+
+    if (!record) {
+      return { activeCurrency: 'USD', balances: [] };
```
klaudia-blazyczek-blurify: Returning a hardcoded `USD` here could be inconsistent with other methods (like `setActiveCurrency`) that throw a `WalletNotFoundError`. Consider throwing an error, or at least adding a comment explaining why returning an empty array is better here.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -341,6 +341,35 @@ export class WalletService {
...
+      .select({ currency: walletBalance.currency, balance: walletBalance.amount })
+      .from(walletBalance)
+      .where(eq(walletBalance.walletId, record.id))
+      .orderBy(walletBalance.currency);
+
+    return { activeCurrency: record.currency, balances: rows };
+  }
+
+  async setActiveCurrency(userId: User['id'], currency: string) {
+    const updated = await this.drizzle.db
```
klaudia-blazyczek-blurify, line 376: It would be good to add a TODO comment here about validating the `currency` against a canonical list of supported currencies, just so it's not forgotten (as mentioned in the PR description).

klaudia-blazyczek-blurify (changes requested):

klaudia-blazyczek-blurify (dismissed):

damianrzepka (dismissed):

damianrzepka (approved):

PR #73, opened 2026-08-17
feat(wallet): network- and tag-aware deposit addresses, async webhook verifier

zaxovaiko:
## Summary
- `wallet_deposit_address` gains `network` and `tag`, and the deposit-address contract gains `network` on input and `network`/`tag` on output.
- Replaces the two old indexes with four partial unique ones: `(userId, currency, network)` and `(userId, currency)` split on whether `network` is null, and `(address, tag)` and `(address)` split on whether `tag` is null.
- Adds `wallet_provider_vault (userId, providerName, vaultAccountId)`, unique on `(userId, providerName)` and on `(providerName, vaultAccountId)`.
- Widens `PaymentWebhookVerifier.verify` to `boolean | Promise<boolean>` and awaits it in the webhook handler. `PaymentAdapter.issueDepositAddress` takes an optional `network` and may return a `tag`; `PaymentWebhookEvent` carries an optional `tag`.

## Why
BF-213: a unique deposit address per player only works the same way for UTXO assets. XRP shares one address across all players and attributes by destination tag, and account-based assets (ETH, USDT, USDC, BNB, TRX, SOL) cannot issue a second address on a shared wallet at all - Fireblocks' own SDK says `createVaultAccountAssetAddress` is "for UTXO or Tag/Memo based assets ONLY". Those need a vault account per player, which is what `wallet_provider_vault` tracks.

`address` was previously a plain index and `findDepositAddressByAddress` returned whichever row came back first, so a shared address would credit an arbitrary player. The partial unique index makes that impossible at the DB level rather than by convention.

The sync `verify()` was the single thing blocking the JWKS webhook migration, since resolving the signing key needs a network fetch. The handler was already `async`, so this is a one-line port change.

## Alternatives considered
`nullsNotDistinct()` on a single unique index per constraint would be tidier, but drizzle 0.45 does not expose it, hence the pairs of partial indexes.

Storing `vaultAccountId` on `wallet_deposit_address` instead of its own table. One vault holds all of a player's account-based assets, so it would be repeated per row and could drift - it belongs one level up.

## Risks
`wallet_provider_vault` is created but nothing writes to it yet. The vault resolver lives in the consumer overlay and lands separately.

No `deposit-options` route here (the currency and network pairs with their minimums). The data lives in overlay config, so it goes with the adapter change rather than in core.


packages/core/src/contracts/adapters/payment.ts
```diff
@@ -17,6 +17,7 @@ export type PaymentWebhookEvent =
   | {
       kind: 'deposit';
       address: string;
+      tag?: string;
```
klaudia-blazyczek-blurify, line 26: Please add `network` to deposit webhook events. The address is now stored per network, but webhook lookup still uses only `address` and `tag`. The same address can exist on two chains, so this either rejects a valid address or cannot match the deposit safely.

packages/core/src/wallet/drizzle/migrations/0009_confused_exodus.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE UNIQUE INDEX "wallet_provider_vault_user_id_provider_name_idx" ON "wallet_provider_vault" USING btree ("user_id","provider_name");--> statement-breakpoint
+CREATE UNIQUE INDEX "wallet_provider_vault_provider_name_vault_account_id_idx" ON "wallet_provider_vault" USING btree ("provider_name","vault_account_id");--> statement-breakpoint
+CREATE UNIQUE INDEX "wallet_deposit_address_user_id_currency_network_idx" ON "wallet_deposit_address" USING btree ("user_id","currency","network") WHERE "wallet_deposit_address"."network" IS NOT NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "wallet_deposit_address_address_tag_idx" ON "wallet_deposit_address" USING btree ("address","tag") WHERE "wallet_deposit_address"."tag" IS NOT NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "wallet_deposit_address_user_id_currency_idx" ON "wallet_deposit_address" USING btree ("user_id","currency") WHERE "wallet_deposit_address"."network" IS NULL;--> statement-breakpoint
+CREATE UNIQUE INDEX "wallet_deposit_address_address_idx" ON "wallet_deposit_address" USING btree ("address") WHERE "wallet_deposit_address"."tag" IS NULL;
```
klaudia-blazyczek-blurify: This migration can fail on a valid existing database. The old `address` index was not unique, so existing rows may share an address. After both new columns start as `NULL`, this unique index rejects those rows. Please add a safe data migration or preflight plan before creating the index.

packages/core/src/contracts/adapters/payment.ts
zaxovaiko: Good catch

klaudia-blazyczek-blurify (dismissed):

klaudia-blazyczek-blurify (dismissed):

klaudia-blazyczek-blurify (dismissed):

damianrzepka (dismissed):

damianrzepka (approved):

PR #78, opened 2026-08-18
feat(social): accept/decline/cancel friend requests + Requests tab li…

marek-chmielowski-blurify:


package.json
```diff
@@ -20,7 +20,7 @@
     "check:format": "oxfmt --check",
     "check:boundaries": "depcruise apps packages --config .dependency-cruiser.cjs",
     "check:shape": "tsx tools/lint/verify-module-shape.ts",
-    "check:deprecations": "depretec --fail-on-found --exclude packages/core/src/pam/identity/adapters/identity-reader.service.ts",
+    "check:deprecations": "depretec --fail-on-found --exclude packages/core/src/pam/identity/adapters/identity-reader.service.ts --exclude packages/core/src/engagement/social/service/social.service.ts",
```
zaxovaiko: ? 😄

packages/core/src/engagement/social/contract/index.ts
```diff
@@ -68,6 +68,31 @@ export const RemoveFriendInputSchema = z.object({
...
+export const ListFriendRequestsInputSchema = z.object({
+  ...PageQuerySchema.shape,
+  direction: FriendRequestDirectionSchema,
+});
+export type ListFriendRequestsInput = z.infer<typeof ListFriendRequestsInputSchema>;
+
+export const FriendRequestEntrySchema = z.object({
+  friendshipId: UuidSchema,
+  userId: UuidSchema,
+  displayName: z.string(),
```
zaxovaiko: Do we need it? Maybe we should resolve the entire user entity instead?

package.json
zaxovaiko: To be honest, I'd check what's wrong there and remove both excludes.

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -481,6 +474,234 @@ export class SocialService {
...
+  async declineFriendRequest(callerId: Uuid, friendshipId: Uuid): Promise<void> {
+    const rows = await this.drizzle.db
+      .update(friendship)
+      .set({ refusedAt: new Date() })
+      .where(
+        and(
+          eq(friendship.id, friendshipId),
+          eq(friendship.addresseeId, callerId),
+          isNull(friendship.acceptedAt),
+          isNull(friendship.refusedAt),
```
zaxovaiko, line 551: Hm, decline sets `refusedAt`, but the partial unique index is only `WHERE removed_at IS NULL` - so the pair stays locked forever and the requester can never send again. However, we agree in scenario 3 that "either player can send a new request in the future", and it treats decline and cancel as the same outcome. Let's set `removedAt` here too, like `cancelFriendRequest` does.

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -481,6 +474,234 @@ export class SocialService {
 
     return { items, total: Number(countRows[0]?.n ?? 0), page, limit };
   }
+
+  async acceptFriendRequest(callerId: Uuid, friendshipId: Uuid): Promise<Friendship> {
```
zaxovaiko, line 473: Don't we need to check if user is blocked here?

packages/core/src/engagement/social/service/social.service.ts
```diff
@@ -481,6 +474,234 @@ export class SocialService {
...
+    }
+
+    this.events.emit('social.friend_request.cancelled', {
+      friendshipId: updated.id,
+      requesterId: updated.requesterId,
+      addresseeId: updated.addresseeId,
+    });
+  }
+
+  async listFriendRequests(
```
zaxovaiko, line 594: We're not filtering suspended/closed accounts here, but `listFriends` does. Confluence also says a blocked player "cannot appear in the blocker's social surfaces"

packages/core/src/engagement/social/service/social.service.ts
zaxovaiko: Can be simplified & shortened a bit

zaxovaiko: Related to the decline change - after a decline this returns refused with `canSendRequest: false` permanently, so the "Add Friend" button stays dead on both sides. Should be set to `true`

agniev-a-hub (commented):
I have nothing to add to the guys review.

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: Fixed

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: Added check

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: fixed

packages/core/src/engagement/social/service/social.service.ts
marek-chmielowski-blurify: done

marek-chmielowski-blurify: fixed.

zaxovaiko (dismissed):
lgtm

zaxovaiko (approved):

PR #81, opened 2026-08-19
feat(admin-console): return assigned admin roles on user list/get/update

marek-chmielowski-blurify:
## Summary

`listUsers`/`getUser`/`updateUser` now return each admin user's `admin_role_assignment` entries (role id + name) inline, batched, via a new `assignedRoles` field.

## Why

A consumer's admin-users table needs each row's assigned roles, but `admin_role_assignment` wasn't reachable from `listUsers` - forcing a bespoke per-user round trip downstream.

## Alternatives considered

Folding this into the existing `AdminUserDirectory` port (owned by `identity`) - rejected: `admin_role_assignment`/`admin_role` are owned by `iam`, not `identity`, so that would cross a module boundary the boundary lint (`no-cross-domain`, `pnpm check:boundaries`) forbids. Instead this adds a dedicated `ADMIN_ROLE_ASSIGNMENT_DIRECTORY` port owned by `iam`, following the same pattern as the existing `ADMIN_PLAYER_ACTIVITY`/`ADMIN_WALLET_REPORTING`/`ADMIN_GAME_REPORTING` read ports.

## Risks

None. Additive only - no new routes, no migration, no event changes; `assignedRoles` is always present (empty array for zero assignments), so existing callers are unaffected.


packages/core/src/iam/adapters/admin-role-assignment-directory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { adminRole, adminRoleAssignment } from '../schema/index.js';
+
+export class DrizzleAdminRoleAssignmentDirectory implements AdminRoleAssignmentDirectory {
+  constructor(private readonly drizzle: DrizzleService) {}
+
+  async listByUserIds(userIds: readonly string[]): Promise<AdminRoleAssignmentSummary[]> {
+    if (userIds.length === 0) {
+      return [];
+    }
+    return this.drizzle.db
```
klaudia-blazyczek-blurify, line 16: Please add a real Postgres test for this query. The current tests mock this port, so they cannot catch a broken join or incorrect role mapping here.

klaudia-blazyczek-blurify: But also it is not a blocker

klaudia-blazyczek-blurify (approved):

packages/core/src/iam/adapters/admin-role-assignment-directory.ts
```diff
@@ -0,0 +1,26 @@
+import type {
+  AdminRoleAssignmentDirectory,
+  AdminRoleAssignmentSummary,
+} from '@openora/core/contracts';
+import { DrizzleService } from '@openora/core/server';
+import { eq, inArray } from 'drizzle-orm';
+import { adminRole, adminRoleAssignment } from '../schema/index.js';
+
+export class DrizzleAdminRoleAssignmentDirectory implements AdminRoleAssignmentDirectory {
```
zaxovaiko, line 9: I don't like the idea of adding an extra wrapper around `AdminRoleAssignmentDirectory` and especially naming it with a `Drizzle` prefix. I feel like it's kinda locks us to Drizzle, and secondly creates an additional unnecessary layer

zaxovaiko (approved):

packages/core/src/iam/adapters/admin-role-assignment-directory.ts
marek-chmielowski-blurify: The purpose of naming this with a Drizzle prefix is actually the opposite of locking us to Drizzle. Swapping orms only means writing a new implementation for the different ORM. That's why this one is named Drizzle, and for other orms the class name would just change and we'd swap the implementation. The consumer never depends on DrizzleAdminRoleAssignmentDirectory directly.

zaxovaiko: The idea is to avoid making too many manual changes, first of all. Second, I don't like using OOP in JS overall 😄. Third, I don't see any real benefit in adding a new wrapper around `AdminRoleAssignmentDirectory` just to add one extra generic function. Let's keep it for now, but remove later in other PRs

PR #83, opened 2026-08-19
feat(identity): add player registration (BF-379)

damianrzepka:
## Summary

BF-379. Adds public player registration - `username` plus terms/age acceptance, an email verification gate before the first sign-in, and a rate-limited username availability endpoint. Replaces `player.displayName` with the identity-owned `username` as the public player identity everywhere.

## Why

We had no self-service registration path at all. Public player identity leaned on `player.display_name`, which isn't unique, isn't validated, and drifted between chat, social and the back office - two players could show up under the same name. Registration also has to capture affirmative terms and age consent, and nothing recorded it.

## Alternatives considered

- Keeping the consent fields on `user`. They're player data, not auth data, so `terms_version`, `terms_accepted_at`, `age_accepted_at`, `registration_ip` and `registration_user_agent` live on `player`. Identity writes them at registration through a new `PLAYER_PROVISIONING` command port that `profile` implements, so identity still never imports `profile`.
- Recording consent from the `identity.user.registered` event instead. `emit()` is best-effort, so a dropped handler would silently lose the consent record. `messaging-and-microservices` says a mutation that has to happen now takes a synchronous command port.
- Moving `username` to `player` as well. better-auth writes it atomically inside its own sign-up transaction via `additionalFields`, and the `lower(username)` unique index has to sit on the same table - splitting it would cost that atomicity and break the chat/social resolution path.
- Backfilling legacy names from `profile.migrate` `postSql`. That only ran when `profile` itself had a pending migration, so existing installs would never have been backfilled. The backfill is part of the tracked identity migration now and `postSql` is gone.

## Risks

- Breaking: sign-in requires a verified address, unconditionally. Nothing has ever set `user.email_verified = true` outside the verification flow itself, so any install carrying accounts from before this change must backfill that column or those accounts are locked out - admins included. Acceptable here because nothing runs this in production yet; a config switch to soften the upgrade was considered and dropped, since an opt-in security default is how the original finding ships unfixed.
- Breaking: `PlayerSchema.displayName` becomes `username` - this also moves the `player.list` sort key and the social `FriendListEntry` / `FriendRequestEntry` payloads.
- Breaking: admin `player.update` no longer takes `email`. Changing a player address needs the verified email-change flow, which is for later.
- Breaking: `identity.register` returns `{ status: 'check-email' }` and sets no session cookie. Registration stays disabled unless the operator configures `registration.termsVersion` and `registration.webUrl`.
- Migration `0007` runs a full-table `UPDATE` plus a non-`CONCURRENTLY` unique index build on `user`, all in one transaction (the migration runner wraps each file, and `CREATE INDEX CONCURRENTLY` is illegal inside a transaction block). On a large `user` table that blocks writes to a hot table for the duration - worth a low-traffic window. Making the runner support non-transactional migrations is deferred.
- The username backfill gives the first holder of a name the plain handle and suffixes later holders with their user id, so migrated names are stable but some of them are ugly. Those players can change them.
- `/identity/email/verify` used to bucket every session-less caller under one shared `anonymous` rate-limit key and only emitted `identity.email.verified` when the caller already had a session. Both were latent before this PR because registration auto-signed-in; the verification link makes the unauthenticated path routine, so both are fixed here.
- Client IP still comes from `x-real-ip` - `extractClientMeta` deliberately ignores `x-forwarded-for` without a trusted proxy boundary. The geo gate and the per-IP registration limit are only as strong as that boundary is at the edge.


packages/core/src/contracts/schemas/identity.ts
```diff
@@ -10,10 +10,21 @@ export type Theme = z.infer<typeof ThemeSchema>;
 // BCP 47 upper bound - the longest real-world tags stay well under this.
 export const LanguageSchema = z.string().max(35);
 
+// The username is the stable, public player identifier. It deliberately has no
+// display-name fallback: registrations and the backfill must establish it before
+// any player-facing surface can resolve an account.
```
zaxovaiko: We can remove comments (applies to all files)

packages/core/src/engagement/social/__tests__/social.router.int.test.ts
```diff
@@ -24,17 +26,31 @@ function build() {
 }
 
 async function seedPlayer(overrides: Partial<typeof player.$inferInsert> = {}) {
```
zaxovaiko: I see this function repeats multiple times in different files. Could we extract it as a helper somewhere?

packages/core/src/pam/identity/drizzle/migrations/0006_far_mephisto.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "user" ADD COLUMN "username" text;--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "terms_version" text;--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "terms_accepted_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "age_accepted_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "registration_ip" text;--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "registration_user_agent" text;--> statement-breakpoint
```
zaxovaiko: Those fileds must go to `players` table

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -123,6 +126,17 @@ type AuthCall<B> = (opts: {
 }) => Promise<globalThis.Response>;
 
 type ExtendedAuthApi = {
+  signUpEmail: AuthCall<{
```
zaxovaiko: Try to reuse already existing types, don't declare new ones

packages/core/src/server/auth/auth.ts
```diff
@@ -78,25 +80,46 @@ export function createAuth(options: AuthOptions): BetterAuthType {
...
+        termsVersion: { type: 'string', required: false, input: true },
+        termsAcceptedAt: { type: 'date', required: false, input: true },
+        ageAcceptedAt: { type: 'date', required: false, input: true },
+        registrationIp: { type: 'string', required: false, input: true },
+        registrationUserAgent: { type: 'string', required: false, input: true },
         theme: { type: 'string', required: false, input: true, defaultValue: 'system' },
         language: { type: 'string', required: false, input: true, defaultValue: 'en' },
       },
     },
     emailAndPassword: {
```
klaudia-blazyczek-blurify, line 88: Sending a verification email does not block login. `IdentityService.login()` can still sign in a new account before the email link is used. Please require email verification and add this flow to the integration tests.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -366,28 +394,85 @@ export class IdentityService {
   }
 
   async register(input: RegisterInput, reqHeaders: NodeHeaders, resHeaders: Headers) {
+    const registration = this.platformConfig?.registration;
+    if (!registration) {
+      throw new ORPCError('FORBIDDEN', { message: 'Registration is unavailable' });
+    }
+    const { ip, userAgent } = extractClientMeta(reqHeaders);
```
klaudia-blazyczek-blurify, line 383: This geo gate trusts `X-Forwarded-For` through `extractClientMeta()`. A client from a blocked country can send an allowed IP and bypass both geo checks and the IP limit. Please use an IP from a trusted proxy boundary.

packages/core/src/pam/player-management/router/index.ts
```diff
@@ -29,41 +33,51 @@ export function createPlayerRouter(
...
+        { NOT_FOUND: PlayerNotFoundError, CONFLICT: DuplicateUsernameError },
+        async () => {
+          const before = await player.get(input.playerId);
+          const updated = await player.update(
+            input.playerId,
+            {
+              username: input.username,
+              status: input.status,
+              level: input.level,
+              email: input.email,
```
klaudia-blazyczek-blurify: Forwarding `email` here lets anyone with `player:update` replace a player email. They can then use password reset to take over the account, bypassing the verified identity email-change flow. Please keep this in a separately authorised and verified workflow.

packages/core/src/pam/profile/migrate.ts
```diff
@@ -12,6 +12,66 @@ export function migrate(databaseUrl?: string) {
     migrationsSchema: 'drizzle',
     // player.display_name GIN trgm index (substring search) needs the pg_trgm extension.
     extensions: ['pg_trgm'],
+    // The generated identity migration adds nullable user.username + its partial
+    // unique index. Profile runs after identity in the shipped plugin order, so
+    // the legacy display name is available here to backfill players safely.
+    postSql: [
```
klaudia-blazyczek-blurify: This backfill will not run on an existing installation. `postSql` only runs when the profile module has pending migrations, but this PR adds no profile migration. Existing player users will keep `username = NULL`. Please add a tracked migration step for this backfill.

packages/core/src/pam/profile/migrate.ts
```diff
@@ -12,6 +12,66 @@ export function migrate(databaseUrl?: string) {
...
+              WHEN normalized_username IS NULL OR length(normalized_username) < 3
+                THEN left('player_' || coalesce(normalized_username, ''), 20)
+              ELSE normalized_username
+            END AS base_username
+          FROM normalized
+        ), ranked AS (
+          SELECT
+            id,
+            base_username,
+            row_number() OVER (PARTITION BY base_username ORDER BY id) AS collision_suffix
```
klaudia-blazyczek-blurify: The suffixes are not unique across different base names. Legacy names `foo`, `foo`, and `foo_2` produce `foo`, `foo_2`, and `foo_2`, so the new unique index aborts the migration. Please allocate globally unique deterministic names and add a migration test.

packages/core/src/engagement/social/contract/index.ts
```diff
@@ -56,7 +56,7 @@ export type Relationship = z.infer<typeof RelationshipSchema>;
 export const FriendListEntrySchema = z.object({
   userId: UuidSchema,
   friendshipId: UuidSchema,
-  displayName: z.string(),
+  username: z.string(),
```
klaudia-blazyczek-blurify, line 59: `listFriendRequests` still returns the legacy `displayName` through `FriendRequestEntrySchema`. Pending requests and accepted friendships will show different public identities. Please move this path to `username` as well, including its contract and tests.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -366,28 +394,85 @@ export class IdentityService {
...
-      playerId: await this.identityReader.getPlayerIdByUserIdSafe(body.user.id),
-      ip,
-      userAgent,
-    });
-    return { user: toUser(body.user) };
+  async usernameAvailable(username: string) {
+    const [existing] = await this.drizzle.db
+      .select({ id: user.id })
+      .from(user)
+      .where(eq(user.username, username))
```
klaudia-blazyczek-blurify: `user_username_unique` indexes `lower(username)`, but this predicate uses `username = ...`, so PostgreSQL cannot use that index. This public endpoint will scan the whole user table on each request. Query `lower(user.username)` or add a matching index, and rate-limit the endpoint.

packages/core/src/contracts/schemas/identity.ts
zaxovaiko: Done in `14f5133f` - stripped the explanatory blocks this PR added, here and in the other files. I kept the `extractClientMeta` warning in `router-utils.ts` though, since it documents a security footgun (what happens if you trust `x-forwarded-for` without a proxy boundary) that isn't derivable from the signature. Shout if you want that one gone too.

packages/core/src/engagement/social/__tests__/social.router.int.test.ts
zaxovaiko: Done in `14f5133f` - extracted `seedPlayerWithUser` into `packages/core/src/testing/seed-player.ts`, and the three social suites now just delegate to it. Same story on the e2e side: the 15-line `/identity/register` block was copy-pasted across 8 files, so that's now `registerPlayer` in `packages/testing/src/register.ts`. There are still ~13 other `seedPlayer` variants in modules this PR doesn't touch - but that's for later.

packages/core/src/pam/identity/service/identity.service.ts
zaxovaiko: Done in `14f5133f` - `signUpEmail` now infers from the table instead of a hand-written body:
```ts
signUpEmail: AuthCall<
  Pick<typeof user.$inferInsert, 'email' | 'name' | 'username'> & { password: string }
>;
```
The consent fields dropped out of it entirely once they moved to `player` (see the migration thread).

packages/core/src/pam/identity/drizzle/migrations/0006_far_mephisto.sql
zaxovaiko: Done in `14f5133f` - `terms_version`, `terms_accepted_at`, `age_accepted_at`, `registration_ip` and `registration_user_agent` are on `player` now (new `profile` migration `0002`). Since the `player` row is created lazily, identity writes them at registration through a new `PLAYER_PROVISIONING` command port that `profile` implements - so identity still never imports `profile`.

I left `username` on `user` - better-auth writes it atomically inside its own sign-up transaction via `additionalFields`, and the `lower(username)` unique index has to sit on the same table. Moving it would split registration into two writes and break the chat/social resolution path. Happy to revisit if you'd rather have it on `player`.

Also renumbered to `0007` - `dev` landed its own `0006` in the meantime.

packages/core/src/server/auth/auth.ts
zaxovaiko: Good catch. Fixed in `14f5133f` - `emailAndPassword.requireEmailVerification: true`, which better-auth enforces in its sign-in route, so no hand-rolled gate needed. Covered by a new `registration.e2e.test.ts`: register, assert login fails, verify, assert login succeeds and sets a cookie. Seeded fixtures are now created `emailVerified: true`, and the shared `registerPlayer` test helper verifies by default (`verifyEmail: false` when a test wants the unverified state).

packages/core/src/pam/player-management/router/index.ts
zaxovaiko: Agreed - dropped in `14f5133f`. `email` is gone from the `player.update` contract, service and audit payload, along with the now-dead `DuplicateEmailError`. It was accepted-but-ignored on `dev` anyway, so nothing regresses. A verified email-change flow is for later.

packages/core/src/pam/profile/migrate.ts
zaxovaiko: You're right, it would never have run. Fixed in `14f5133f` - the backfill moved into the identity migration `0007` itself, between the `ADD COLUMN` and the `CREATE UNIQUE INDEX`, so it's tracked and runs exactly once on existing installs. `postSql` is deleted from `runMigrations` entirely - it only existed for this.

packages/core/src/pam/profile/migrate.ts
zaxovaiko: Also right. The suffix loop is gone in `14f5133f`. The first holder of a base name keeps it and every later holder falls back to `left(base, 7) || '_' || left(user_id, 12)`, so allocation is unique by construction in a single set-based statement - no per-row probing against the growing set of taken handles.

Worth noting: my first attempt was the obvious `WHILE EXISTS ... suffix + 1` loop, and it wedged for minutes on a 2.6k-row table because a shared base name makes it O(n^2). The `CREATE UNIQUE INDEX` runs right after in the same transaction, so if the allocation ever were wrong the migration fails loudly instead of corrupting.

packages/core/src/engagement/social/contract/index.ts
zaxovaiko: Done in `14f5133f` - `FriendRequestEntrySchema` uses `username` now, and `listFriendRequests` joins `user` for it like the friends list already did. Tests updated.

packages/core/src/pam/identity/service/identity.service.ts
zaxovaiko: Half agree. `extractClientMeta` doesn't read `x-forwarded-for` at all unless you pass `{ trustForwarded: true }`, and registration doesn't - so the specific XFF bypass isn't there. It does still take `x-real-ip` from the request, which is client-controlled without a proxy boundary, so the underlying point stands.

That's the repo-wide convention across all 18 call sites though, documented on the helper, so I don't want to change the boundary in this PR - it's an app/deployment concern (strip and replace the header at the edge). Flagged it under Risks in the description. Happy to open a separate ticket for a real trusted-proxy config if you'd rather not leave it.

Fixed the test helper while in here - it was setting `x-forwarded-for` to vary the client IP, which was silently ignored, so every registration shared one rate-limit bucket. It uses `x-real-ip` now.

packages/core/src/pam/identity/service/identity.service.ts
zaxovaiko: Done in `14f5133f` - all three lookups (`usernameAvailable` plus the two in `register`) query `lower(user.username)` now, so they hit `user_username_unique`. `UsernameSchema` already lowercases on parse, so the comparison side is normalised.

Rate limit added too, but keyed on the caller IP rather than the handle (`30/min`) - the abuse shape is one client enumerating many usernames, not one username being probed repeatedly, so a per-username bucket wouldn't have stopped anything.

packages/core/src/server/auth/auth.ts
zaxovaiko: Changed my mind on this one in `440cf2e7` - it's now unconditional rather than a config switch.

I'd first made it opt-in via `registration.requireEmailVerification` (default off) to avoid locking out accounts created before `email_verified` was tracked, since nothing has ever set that column outside the verification flow. But an opt-in security default just means your finding ships unfixed for anyone who doesn't read the config, so the knob is gone and the gate is always on. We can take the breaking change because nothing runs this in production yet - any install carrying older accounts has to backfill `user.email_verified = true`, which is called out in the changeset and the description.

Covered end to end in `registration.e2e.test.ts`: register, assert login is refused, follow the real emailed link, assert login succeeds and sets a cookie. The verification step goes through `/identity/email/verify` with the token pulled out of the actual sent email, not a direct DB write.

zaxovaiko (approved):

PR #84, opened 2026-08-19
feat(identity): rolling window lockout

agniev-a-hub:


packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -330,6 +345,36 @@ export class PhoneLoginService {
...
+    if (hasFailedLoginWindowExpired(account.lastFailedLoginAt, account.lastLockoutAt, nowMs)) {
+      await this.drizzle.db
+        .update(user)
+        .set({
+          failedLoginAttempts: 0,
+          lockoutUntil: null,
+          lockoutCount: 0,
+          lastLockoutAt: null,
+          lastFailedLoginAt: null,
+        })
```
klaudia-blazyczek-blurify, line 362: Mutating `account` in place after a DB reset breaks the immutability rule (conventions.md: "derive new objects, don't mutate"). Replace with `account = { ...account, failedLoginAttempts: 0, lockoutUntil: null, lockoutCount: 0, lastLockoutAt: null, lastFailedLoginAt: null }`.

packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -330,6 +345,36 @@ export class PhoneLoginService {
...
+      account.lockoutUntil = null;
+      account.lockoutCount = 0;
+      account.lastLockoutAt = null;
+      account.lastFailedLoginAt = null;
+    } else if (account.lockoutUntil && account.lockoutUntil.getTime() <= nowMs) {
+      await this.drizzle.db
+        .update(user)
+        .set({ failedLoginAttempts: 0, lockoutUntil: null })
+        .where(eq(user.id, account.id));
+      account.failedLoginAttempts = 0;
```
klaudia-blazyczek-blurify: Same mutation issue for the expired-lock branch: `account.failedLoginAttempts = 0; account.lockoutUntil = null` mutates the object directly. Use a spread reassignment.

packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -393,6 +438,10 @@ export class PhoneLoginService {
         createdAt: new Date(),
         updatedAt: new Date(),
       });
+      await t
```
klaudia-blazyczek-blurify, line 443: On successful OTP verify, only `failedLoginAttempts` and `lockoutUntil` are cleared inside the transaction. `lockoutCount`, `lastLockoutAt`, and `lastFailedLoginAt` are left dirty. The next failed login will read a stale `lockoutCount` and immediately escalate the tier. Should call `resetLockoutTier` (or inline the full reset) to clear all five fields.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -456,9 +431,37 @@ export class IdentityService {
       existingUser = { ...existingUser, failedLoginAttempts: 0, lockoutUntil: null };
```
klaudia-blazyczek-blurify: When `lockoutUntil` has elapsed, only `failedLoginAttempts` and `lockoutUntil` are cleared here (via `existingUser` spread), but the DB row is not updated - the clean-up write is done at line ~427 via `clearLockout`. Immediately after, the window-expiry check at line ~436 may fire a *second* DB update on the same request if `lastFailedLoginAt` is old enough. That is two UPDATE round-trips for a single login attempt. Consider merging both conditions into one write.

packages/core/src/pam/identity/service/lockout-policy.service.ts
```diff
@@ -0,0 +1,75 @@
...
+export const LOCKOUT_WINDOW_MS = 24 * 60 * 60 * 1000;
+export const LOCKOUT_TIER_DURATIONS_MS = [60_000, 5 * 60_000, 15 * 60_000] as const;
+
+export function isActiveLockout(lockoutUntil: Date | null | undefined, nowMs = Date.now()) {
+  return lockoutUntil !== null && lockoutUntil !== undefined && lockoutUntil.getTime() > nowMs;
+}
+
+export function hasFailedLoginWindowExpired(
+  lastFailedLoginAt: Date | null | undefined,
+  fallbackLastLockoutAt?: Date | null,
```
klaudia-blazyczek-blurify: `isActiveLockout` checks both `!== null` and `!== undefined` separately. Since the parameter is typed `Date | null | undefined`, prefer `lockoutUntil != null` (loose inequality) which covers both in one check, which is the idiomatic TS pattern for nullish guards.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -587,7 +595,12 @@ export class IdentityService {
       }
       // Only a genuine credential rejection counts toward lockout - a transient DB/network
       // error must never lock the account out.
```
klaudia-blazyczek-blurify, line 604: `error.data as { code?: string } | undefined` is an unchecked `as` cast used to silence the compiler (conventions.md: "No `as` casts to silence the compiler - fix the root cause"). `ORPCError.data` is already typed; narrow it properly or use `isObject(error.data) && error.data['code'] === 'ACCOUNT_LOCKED'`.

klaudia-blazyczek-blurify:
**PR title** `Feat/bf 454/rolling window lockout` doesn't follow the conventional commit format (conventions.md). The squash-merge commit on `dev` is derived directly from the MR title — an uppercase or non-conventional title lands a failing commitlint check on `dev`.

Expected something like: `feat(identity): rolling window lockout`

klaudia-blazyczek-blurify (changes requested):

agniev-a-hub:
@klaudia-blazyczek-blurify corrections applied, ty

klaudia-blazyczek-blurify (approved):

PR #85, opened 2026-08-20
feat(wallet): add manual balance adjustments

zaxovaiko:
## Summary

Adds super-admin manual wallet credits and debits with durable idempotency, atomic audit records, and admin transaction history metadata.

## Why

Operations need a controlled correction path alongside reconciliation for exceptional balance changes.

## Alternatives considered

Reconciliation alone does not cover deliberate operational corrections. Reusing payment flows would incorrectly involve payment providers.

## Risks

- Adds `manual_credit` and `manual_debit` enum values through a generated migration.
- Full verification exposed existing flaky rate-limit and async tag tests. Schema, wallet integration, and HTTP E2E tests pass independently.

### DIFF
--- a/packages/core/src/contracts/schemas/common.ts
+++ b/packages/core/src/contracts/schemas/common.ts
@@ -23,12 +23,13 @@ export type ClientMeta = z.infer<typeof ClientMetaSchema>;
 // and the contract can never drift into truncating each other.
 export const MONEY_SCALE = 18;
 export const MONEY_PRECISION = 38;
+const MONEY_INTEGER_DIGITS = MONEY_PRECISION - MONEY_SCALE;
 
 export const MoneyAmountSchema = z
   .string()
   .regex(
-    new RegExp(`^\\d+(\\.\\d{1,${MONEY_SCALE}})?$`),
-    `must be a non-negative decimal string with at most ${MONEY_SCALE} decimal places`,
+    new RegExp(`^\\d{1,${MONEY_INTEGER_DIGITS}}(\\.\\d{1,${MONEY_SCALE}})?$`),
+    `must be a non-negative decimal string with at most ${MONEY_INTEGER_DIGITS} integer and ${MONEY_SCALE} decimal places`,
   );
 export type MoneyAmount = z.infer<typeof MoneyAmountSchema>;
 
@@ -47,7 +48,7 @@ export type AuthGuardReason = z.infer<typeof AuthGuardReasonSchema>;
 export const SignedMoneyAmountSchema = z
   .string()
   .regex(
-    new RegExp(`^-?\\d+(\\.\\d{1,${MONEY_SCALE}})?$`),
-    `must be a decimal string (optionally negative) with at most ${MONEY_SCALE} decimal places`,
+    new RegExp(`^-?\\d{1,${MONEY_INTEGER_DIGITS}}(\\.\\d{1,${MONEY_SCALE}})?$`),
+    `must be a decimal string (optionally negative) with at most ${MONEY_INTEGER_DIGITS} integer and ${MONEY_SCALE} decimal places`,
   );
 export type SignedMoneyAmount = z.infer<typeof SignedMoneyAmountSchema>;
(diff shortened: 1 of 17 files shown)


packages/core/src/contracts/schemas/common.ts
```diff
@@ -23,12 +23,13 @@ export type ClientMeta = z.infer<typeof ClientMetaSchema>;
...
 export const MONEY_SCALE = 18;
 export const MONEY_PRECISION = 38;
+const MONEY_INTEGER_DIGITS = MONEY_PRECISION - MONEY_SCALE;
 
 export const MoneyAmountSchema = z
   .string()
   .regex(
-    new RegExp(`^\\d+(\\.\\d{1,${MONEY_SCALE}})?$`),
-    `must be a non-negative decimal string with at most ${MONEY_SCALE} decimal places`,
+    new RegExp(`^\\d{1,${MONEY_INTEGER_DIGITS}}(\\.\\d{1,${MONEY_SCALE}})?$`),
```
klaudia-blazyczek-blurify, line 31: This rejects valid `numeric(38,18)` values with leading zeroes, such as `000000000000000000001`. PostgreSQL accepts that value, so this is a breaking validation change unrelated to the limit. Validate numeric precision instead of input string length.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -502,6 +505,125 @@ export class WalletService {
...
+      if (existing) {
+        this.assertManualAdjustmentReplayMatches(existing, {
+          adminId,
+          direction,
+          amount,
+          currency,
+          reason,
+        });
+        return { transactionId: existing.id, status: existing.status };
+      }
```
klaudia-blazyczek-blurify, line 566: `balanceBefore` is read without locking the balance row. A deposit can commit after this read and before the adjustment, so the immutable audit record can store the wrong before balance. Lock and read the balance with the adjustment, or derive it from the returned balance.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -502,6 +505,125 @@ export class WalletService {
     return { transactionId, status: 'completed' };
   }
 
+  async manualAdjust({
```
marek-chmielowski-blurify, line 508: `manualAdjust()` updates the wallet balance and writes an audit row, but unlike every other balance-mutating method on `WalletService` (deposit, withdraw, etc.) it never emits a domain event.

That means admin-issued manual credits/debits are invisible to `compliance/service/rg-monitoring.service.ts` and `pam/tag/service/tag-evaluation.service.ts`, both of which only react to `wallet.deposit.completed`. A manual credit to a self-excluded or deposit-limited player's wallet would silently bypass RG deposit-limit checks and deposit-velocity risk tagging that a real deposit of the same amount would trigger.

This also seems to contradict the class's own documented contract a few lines up (around line 305-311): *"Domain events emit only AFTER the owning transaction commits, and only for the non-replay branch of an idempotent call"* — which implies every non-replay balance change should emit one.

Worth emitting a dedicated event (e.g. `wallet.manualAdjustment.completed`) after commit, and wiring RG/tagging to consume it, or explicitly documenting why manual adjustments are exempt.

zaxovaiko:
Folded into #92 - it replays this commit, so merging here would add a second wallet migration `0010` alongside the one #92 generates.

All three review points are fixed there in `602228ea`:

- `MoneyAmountSchema` was bounding the string, not the value, so `000000000000000000001` got rejected even though Postgres stores it fine in `numeric(38,18)`. Leading zeros are consumed before the digit budget now.
- `balanceBefore` is derived from the balance the update returns, instead of a separate read. `manualAdjust` locks the wallet row but `creditDepositByAddress` does not take that lock, so a deposit could commit in between and the audit row would keep a `before` that never existed.
- `manualAdjust` emits `wallet.manual_adjustment.created`. Gave it its own topic instead of reusing `wallet.deposit.completed` - a correction is not a deposit, and it would overstate deposits and GGR.

Closing in favour of #92, review there please.

PR #86, opened 2026-08-20
fix(chat): allow standalone digit 5

jakubfilinger-b:
BF-436

### DIFF
--- a/packages/core/src/engagement/chat/moderation/profanity.ts
+++ b/packages/core/src/engagement/chat/moderation/profanity.ts
@@ -14,6 +14,9 @@ const profanity = new Profanity({
   unicodeWordBoundaries: true,
 });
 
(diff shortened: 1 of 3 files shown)


packages/core/src/engagement/chat/moderation/profanity.ts
```diff
@@ -14,6 +14,9 @@ const profanity = new Profanity({
   unicodeWordBoundaries: true,
 });
 
+// @2toad/profanity 3.3.0 incorrectly lists "5" as Portuguese profanity.
+profanity.removeWords(['5']);
```
zaxovaiko, line 18: https://github.com/2Toad/Profanity/blob/19d6da192d61bd14544b7e417ea2613a2ba8357f/src/data/profane-words.ts#L2866 indeed

zaxovaiko (approved):

PR #88, opened 2026-08-20
feat(wallet): operator-editable asset catalog and custody adapter seam

zaxovaiko:
## Summary

- Adds `wallet_asset`: the `(currency, network)` pairs the platform accepts, each with its own minimum deposit, minimum withdrawal, withdrawal fee, and independent deposit/withdrawal enable flags.
- Adds a public `GET /wallet/assets` serving the enabled pairs to the player currency and network pickers, plus an admin CRUD group behind a new `wallet-asset` permission resource.
- Adds `WALLET_ASSET_CATALOG` so a payment adapter can read the catalog without importing wallet internals. The wallet plugin binds a DB-backed default, so an operator gets a working catalog with no wiring.
- Extends `PaymentAdapter` with `supportsAsset`, plus optional `listSweepableBalances`, `sweepToPool`, `listTransactions` and `getWithdrawalStatus` for a custody vendor that pools per-player balances.

## Why

Which currencies a platform accepts was operator config living in code and env vars, so adding one meant a deploy. The limits were also per-currency, which is wrong the moment a currency exists on more than one chain: USDT costs cents to move on one network and dollars on another, so a single currency-wide floor is wrong on at least one of them.

Deposit and withdrawal enable separately because they genuinely come up at different times - an operator can accept deposits for an asset before the withdrawal-side float behind it exists.

`providerAssetId` is the only vendor-touching column and the wallet module treats it as an opaque string, so a custody vendor's identifier scheme stays entirely in that vendor's adapter. The four optional `PaymentAdapter` methods are the seam a sweeping and reconciliation job will sit on later - this PR only defines them, nothing calls them yet.

## Alternatives considered

A second DI token for the custody methods, instead of optional methods on `PaymentAdapter`. Optional methods match the existing `issueDepositAddress`/`parseWebhook` idiom, and mean `MockPaymentAdapter` and every PSP adapter need no change at all.

Keeping the currency and network sets as enums validated at boot. That catches a bad value earlier, but it is exactly what makes adding a currency a deploy. `supportsAsset` recovers the safety: the catalog write path asks the bound adapter first, so an admin cannot save a pair the vendor cannot serve, and the failure lands on the admin at config time instead of on a player at deposit time.

## Risks

The `wallet_asset` migration is additive and nothing existing reads the table, so current deposit and withdrawal behaviour is unchanged.

Deleting a pair is blocked while players still hold that currency. `wallet_balance` is keyed by currency alone, with no `network` column, so that guard is necessarily currency-wide: removing one network of a currency someone holds is blocked even if their balance arrived over a different one. It fails safe, and widening it would mean putting `network` on `wallet_balance`, which is a bigger change than this PR should carry.

`wallet-asset` is a new permission resource. The built-in `admin` role gets it here, but any role whose levels are already stored will have no level for it and so no access until it is granted.

The `(currency, network)` key is deliberately immutable on update - renaming is a delete plus a create, so a vendor reference cannot be rewritten out from under a pending transaction.


packages/core/src/wallet/router/index.ts
```diff
@@ -202,6 +206,53 @@ export function createWalletRouter(
       }),
     },
 
+    listAssets: os.listAssets.handler(() => wallet.listEnabledWalletAssets()),
```
klaudia-blazyczek-blurify, line 209: This only filters the picker response. The deposit-address route still accepts any pair, and the deposit and withdrawal flows never read this catalog. A player can call the API directly after an asset is disabled or removed, bypassing the flags, limits, and fee. Enforce the configured asset in every server-side flow; a withdrawal also needs to carry and persist its network.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1289,6 +1327,143 @@ export class WalletService {
...
+  async updateWalletAsset(
+    adminId: User['id'],
+    { currency, network, ...changes }: UpdateWalletAssetInput,
+    meta?: ClientMeta,
+  ): Promise<WalletAsset> {
+    if (changes.providerAssetId !== undefined) {
+      this.assertAdapterSupports(currency, network);
+    }
+    return this.drizzle.db.transaction(async (txn) => {
+      const [before] = await txn
```
klaudia-blazyczek-blurify, line 1397: Lock this row, or use a versioned update, before reading it. Two admins can both read the same old state and update it in sequence, so the second audit record misses the first change. The audit trail needs the real before and after state.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1289,6 +1327,143 @@ export class WalletService {
...
+      // necessarily currency-wide: removing one network of a currency players still hold
+      // is blocked even if their balance arrived over another network. Fails safe.
+      const [held] = await txn
+        .select({ n: count() })
+        .from(walletBalance)
+        .where(and(eq(walletBalance.currency, currency), sql`${walletBalance.amount} > 0`));
+      if ((held?.n ?? 0) > 0) {
+        throw new WalletAssetInUseError();
+      }
+      await txn
```
klaudia-blazyczek-blurify, line 1450: Check that the delete actually removed a row before writing the audit record and returning true. Two concurrent deletes can both read the row; after the first commits, the second deletes nothing but still records another deletion and reports success.

klaudia-blazyczek-blurify (changes requested):

zaxovaiko:
Superseded by #92.

Every commit here is in that branch - the asset catalog and the custody adapter seam were cherry-picked onto it unchanged, and the tests came with them. Closing this one rather than rebasing it, because the reason to fold it in was structural: this PR and #85 both generated a wallet migration `0010`, and #89 stacked an `0011` on top. Renumbering by hand would leave drizzle hashing bytes that no longer match what was applied. #92 regenerates a single `0010` from the final schema instead, so the collision cannot happen.

No review comments were left here, so nothing is lost by moving. The work continues in #92 -> #95 -> #96.

PR #89, opened 2026-08-20
feat(wallet): settle withdrawals on an explicit network

zaxovaiko:
## Summary

- `wallet_transaction` gains a `network` column and a `(currency, network)` index, so reconciliation and reporting are per chain rather than per currency.
- A withdrawal now resolves its settlement network from the asset catalog before any funds are held, and pins it on the row. `POST /wallet/withdraw` takes an optional `network`.
- The per-chain `minWithdrawal` from the catalog is enforced on that resolved network.
- `approveWithdrawal` passes the pinned `network` to the adapter, and a webhook deposit records the chain the vendor reported, falling back to the one the address was issued on.

## Why

Stacked on #88, which added the catalog but left nothing reading it. This is the half that makes the config do something.

A currency does not identify a chain. USDT settles on ERC20, TRC20 and BEP20 with different addresses, fees and minimums, so a payout that carries only a currency is ambiguous, and an adapter receiving it has to guess. Guessing wrong sends a player's funds over a chain their receiving wallet may not support. Storing only the currency on the transaction has the same problem after the fact: two USDT withdrawals on different chains are indistinguishable in the ledger.

## Alternatives considered

Inferring the network from the destination address. That works for chains with distinctive address formats and fails exactly where it matters most, since ERC20 and BEP20 share the EVM address format - the two chains a USDT payout is most likely to be confused between.

Defaulting to a "primary" network per currency instead of rejecting an ambiguous request. A silent default is the failure mode this is meant to prevent, and it would be indistinguishable from a correct payout until the player reports missing funds.

## Risks

The column is nullable and backfills as `NULL`, so existing rows keep working. `NULL` stays correct for the fiat rail and for internal types like bet, win and bonus that never touch a chain.

Behaviour only changes for a currency that has catalog rows. A currency with none passes through unchecked, which keeps a fiat-only operator working exactly as before. A currency whose rows are all withdrawal-disabled now fails closed, which is a deliberate reversal of the previous silent-allow.

An operator who populates the catalog with several networks for one currency turns previously-accepted `POST /wallet/withdraw` calls that omit `network` into a 4xx. That is the intended behaviour, but it is a client-visible change for anyone who configures a multi-chain currency.

Minimum deposit is stored but still unenforced. A deposit arrives by webhook after the funds have already moved, so rejecting one is not possible; deciding what to do with a below-minimum deposit is its own piece of work.

### DIFF
--- a/packages/core/src/wallet/schema/index.ts
+++ b/packages/core/src/wallet/schema/index.ts
@@ -72,6 +72,10 @@ export const walletTransaction = pgTable(
     type: walletTransactionTypeEnum().notNull(),
     amount: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull(),
     currency: text().notNull(),
+    // The chain this moved on. A currency does not identify one, so reconciliation and
+    // reporting are per (currency, network). Null on a fiat rail and on any internal
+    // transaction type (bet/win/bonus) that never touches a chain.
+    network: text(),
     status: walletTransactionStatusEnum()
       .$type<WalletTransactionStatus>()
       .notNull()
@@ -104,6 +108,7 @@ export const walletTransaction = pgTable(
     index('wallet_transaction_status_idx').on(t.status),
     index('wallet_transaction_rail_idx').on(t.rail),
     index('wallet_transaction_currency_idx').on(t.currency),
+    index('wallet_transaction_currency_network_idx').on(t.currency, t.network),
     index('wallet_transaction_tx_hash_idx').on(t.txHash),
     index('wallet_transaction_status_type_created_at_idx').on(t.status, t.type, t.createdAt),
     index('wallet_transaction_wallet_id_type_status_idx').on(t.walletId, t.type, t.status),
--- a/packages/core/src/wallet/service/wallet.service.ts
+++ b/packages/core/src/wallet/service/wallet.service.ts
@@ -117,6 +117,28 @@ export const WalletAssetInUseError = makeConflictError(
   'Players still hold a balance in this currency',
 );
 
+export const AmbiguousNetworkError = createDomainError(
+  'AmbiguousNetworkError',
+  (currency, networks) =>
+    `Currency ${currency} settles on several networks (${networks}) - a network is required`,
+);
+
+export const UnsupportedNetworkError = createDomainError(
+  'UnsupportedNetworkError',
+  (currency, network) => `Currency ${currency} is not settled on network ${network}`,
+);
+
+export const WithdrawalDisabledError = createDomainError(
+  'WithdrawalDisabledError',
+  (currency) => `Withdrawals are disabled for ${currency} on every configured network`,
+);
+
+export const BelowMinimumWithdrawalError = createDomainError(
+  'BelowMinimumWithdrawalError',
+  (amount, minimum, currency, network) =>
+    `Withdrawal of ${amount} ${currency} on ${network} is below the ${minimum} ${currency} minimum`,
+);
+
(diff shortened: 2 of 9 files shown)


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -145,6 +167,69 @@ const DEFAULT_WALLET_CURRENCY = 'USD';
...
+export function assertAboveMinimumWithdrawal(
+  assets: readonly WithdrawalAsset[],
+  amount: string,
+  currency: string,
+  network: string | null,
+): void {
+  const asset = assets.find(
+    (candidate) => candidate.withdrawalEnabled && candidate.network.toUpperCase() === network,
+  );
+  if (asset && moneyToNumber(amount) < moneyToNumber(asset.minWithdrawal)) {
```
klaudia-blazyczek-blurify, line 228: Please use an exact decimal comparison here. `moneyToNumber` intentionally loses precision and is only for coarse heuristics, so two amounts that differ past the 15th digit can be treated as equal and a below-minimum withdrawal can pass. Compare exactly and add a boundary test.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -665,6 +765,9 @@ export class WalletService {
     if (this.resolveRail(currency) === 'crypto' && !destinationAddress) {
```
klaudia-blazyczek-blurify, line 765: Please check an existing idempotency key before applying the current asset rules. If a pending ERC20 withdrawal is retried after that asset is disabled or its minimum increases, this now throws instead of returning the original transaction. A money replay must return its original result.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -705,6 +808,7 @@ export class WalletService {
             currency,
```
klaudia-blazyczek-blurify, line 808: Please include `network` in the idempotency comparison. A first request for USDT on ERC20 and a retry with the same key for BEP20 have the same amount and currency, so the retry is accepted but returns the ERC20 transaction. The network changes the payment destination and must cause an idempotency-key-reuse error.

packages/core/src/wallet/schema/index.ts
```diff
@@ -72,6 +72,10 @@ export const walletTransaction = pgTable(
     type: walletTransactionTypeEnum().notNull(),
     amount: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull(),
     currency: text().notNull(),
+    // The chain this moved on. A currency does not identify one, so reconciliation and
+    // reporting are per (currency, network). Null on a fiat rail and on any internal
```
klaudia-blazyczek-blurify, line 76: Please extend the admin reporting contract and implementation with `network`. The new column and index are not returned or filterable by `AdminWalletReporting`, so back-office reconciliation and reporting are still only per currency.

klaudia-blazyczek-blurify (changes requested):

zaxovaiko:
Superseded by #92.

The network-settlement commit is on that branch unchanged - it applied cleanly, since this PR was stacked on #88 and both share the same base.

One thing did change in the move: `assertAboveMinimumWithdrawal` compared `moneyToNumber(amount)` against `moneyToNumber(asset.minWithdrawal)`. That is a float comparison gating a real player withdrawal, and `MONEY_SCALE` is 18 while a double carries about 15 significant digits. #92 replaces it with an exact `moneyCompare` over integer minor units.

Closing rather than rebasing, for the migration reason described in #88. The work continues in #92 -> #95 -> #96.

PR #93, opened 2026-08-21
fix(chat): mute fix applied to the globla chat

agniev-a-hub:


packages/core/src/engagement/social-transfers/service/social-transfers.service.ts
```diff
@@ -922,6 +928,9 @@ export class SocialTransfersService implements GiftCommands, RainCommands {
     if (target.userId === actorId) {
       throw new DonateSelfError();
     }
+    if (await this.blockWriter.isBlockedBetween(actorId, target.userId)) {
```
klaudia-blazyczek-blurify: This check runs before the transfer transaction. If either user blocks the other after it returns false, the donation is still debited and credited. Check and lock the pair inside the transfer transaction so a block cannot race a transfer.

packages/core/src/contracts/adapters/chat-block-writer.ts
```diff
@@ -8,6 +8,7 @@ export type ChatBlockWriter = {
    * blocked/ignored players out of player search results.
    */
   getExcludedUserIds(viewerId: string): Promise<string[]>;
+  isBlockedBetween(userA: string, userB: string): Promise<boolean>;
```
klaudia-blazyczek-blurify: Please use `Uuid` for these parameters. They are user IDs on a public port, and `string` accepts invalid values.

klaudia-blazyczek-blurify (changes requested):

klaudia-blazyczek-blurify (dismissed):

damianrzepka (approved):

PR #94, opened 2026-08-21
fix(chat): provision and protect the __global chat room

jakubfilinger-b:
Without a persisted __global room, getConnection() never grants the chat:global realtime capability to authenticated viewers, so presence.enter() is rejected and the online count silently stays 0.

- migration seeds the __global room (idempotent, ON CONFLICT DO NOTHING)
- deleteRoom now rejects deleting the global room (ChatRoomProtectedError)
- regression test covers the migration seed, listRooms, and the delete guard

### DIFF
--- a/packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
+++ b/packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
@@ -1,0 +1,7 @@
+-- The `__global` room is a system invariant every deployment needs: without it,
+-- `/chat/connection` never grants the `chat:global` realtime capability, presence.enter()
+-- is rejected, and the online count silently stays 0. ON CONFLICT makes this safe to
+-- apply even if a room with this slug was already created out-of-band (eg by the demo seed).
+INSERT INTO "chat_room" (name, slug, is_public)
+VALUES ('Global', '__global', true)
+ON CONFLICT ("slug") DO NOTHING;
--- a/packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
+++ b/packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
@@ -1,7 +1,8 @@
 -- The `__global` room is a system invariant every deployment needs: without it,
 -- `/chat/connection` never grants the `chat:global` realtime capability, presence.enter()
 -- is rejected, and the online count silently stays 0. ON CONFLICT makes this safe to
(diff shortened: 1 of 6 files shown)


packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
```diff
@@ -0,0 +1,7 @@
+-- The `__global` room is a system invariant every deployment needs: without it,
+-- `/chat/connection` never grants the `chat:global` realtime capability, presence.enter()
+-- is rejected, and the online count silently stays 0. ON CONFLICT makes this safe to
+-- apply even if a room with this slug was already created out-of-band (eg by the demo seed).
+INSERT INTO "chat_room" (name, slug, is_public)
+VALUES ('Global', '__global', true)
+ON CONFLICT ("slug") DO NOTHING;
```
klaudia-blazyczek-blurify: A previously soft-deleted `__global` room still owns this unique slug. `ON CONFLICT DO NOTHING` leaves it deleted, so global chat is still unavailable. Restore the room on conflict.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -1601,6 +1606,17 @@ export class ChatService {
   }
 
   async deleteRoom(id: ChatRoom['id'], actorId?: User['id'], meta?: ClientMeta) {
+    const existing = findOneOrThrow(
+      await this.drizzle.db
+        .select({ slug: chatRoom.slug })
+        .from(chatRoom)
+        .where(and(eq(chatRoom.id, id), isNull(chatRoom.deletedAt)))
+        .limit(1),
+      new ChatRoomNotFoundError(id),
```
klaudia-blazyczek-blurify, line 1618: An admin can still rename the `__global` room through `updateRoom`. Then `getConnection` cannot find it and does not grant global chat access. Protect its slug from updates too.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/engagement/chat/drizzle/migrations/0007_seed_global_chat_room.sql
jakubfilinger-b: fixed

packages/core/src/engagement/chat/service/chat.service.ts
jakubfilinger-b: fixed

klaudia-blazyczek-blurify (approved):

zaxovaiko (approved):

PR #102, opened 2026-08-22
feat(identity): verify registration email by otp and sign in on verify (BF-379)

zaxovaiko:
## Summary

Registration now mails a six-digit code instead of a magic link, and verifying that code is what mints the session. Email verification enforcement moves behind `registration.requireEmailVerification` (default off). BF-379.

## Why

Two acceptance criteria contradicted what shipped: "successful registration creates a session and redirects to the homepage", and "unverified players are not restricted while the KYC toggle is off".

They cannot both be satisfied at sign-up. better-auth only returns its indistinguishable duplicate-email response when `autoSignIn` is off or verification is required, so minting the session at sign-up would have traded away the "registering with an already used email does not reveal whether the account exists" criterion. Moving the session to the verification step satisfies all three: sign-up stays sessionless and indistinguishable, and the player is signed in the moment they enter the code.

The verification gate on sign-in was unconditional, which locked out every unverified account. It is now operator config, defaulting off, so unverified players stay unrestricted until KYC is turned on.

The RG and suspension gate was duplicated across password login and phone login. The new verification path mints a session too, so it needed the same gate; rather than add a third copy, the check moved into one shared implementation in `rg-guard.service.ts` that all three call. A gate only some login paths enforce is not a gate.

## Alternatives considered

- `autoSignIn: true` at sign-up. Simplest reading of the ticket, but it turns off better-auth's generic duplicate-email response and hands back account enumeration on the registration form.
- Keeping verification mandatory and treating the ticket as wrong. Rejected: the product decision is explicit that unverified players are unrestricted while KYC is off.
- `emailOTP({ sendVerificationOnSignUp: true })` to mail the code. Rejected: that hook fires off the sign-up response, which on the duplicate-email path carries a synthetic user with the real owner's address. It would mail a live code to someone who never registered, and verifying that code signs its bearer in. The code is therefore sent explicitly from `register()`, only in the genuinely-new-user branch.
- Leaving the link flow in place alongside the code. Rejected as dead surface once nothing issues links.

## Risks

Breaking changes for downstream operators:

- `verifyEmail` takes `{ email, otp }` instead of `{ token }` and returns `{ user, session }` (or `{ twoFactorRedirect: true }`) instead of `{ success }`.
- `sendEmailVerification` takes `{ email }` and is unauthenticated, since the player has no session until the code is verified. It always answers success and only mails an account that exists and is still unverified.
- The `verifyEmail` email template renders an `otp` instead of a `url`/`token`; custom renderers need updating.
- `registration.webUrl` is removed and `registration.requireEmailVerification` added. `RegistrationConfigSchema` is `.strict()`, so an operator config still carrying `webUrl` now fails validation at boot rather than being ignored.

Operational:

- With the gate off by default, an unverified account is fully functional: someone can register an address they do not control, play and cash out, and the real owner can no longer register that address. This follows the product decision that unverified players are unrestricted while KYC is off, and is flagged rather than overridden. Operators wanting the stricter behaviour set `registration.requireEmailVerification: true`.
- A 2FA-enrolled account is verified but deliberately not signed in by the code: better-auth mints that session through `createSession`, which its twoFactor plugin does not hook, so the code alone would otherwise bypass the second factor. Such accounts get `twoFactorRedirect` and sign in through `login`.
- The verify and resend rate-limit budgets fail closed, so a limiter outage rejects verification attempts rather than opening a brute-force window on a six-digit code.


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1178,68 +1166,113 @@ export class IdentityService {
...
-      return null;
+    // better-auth mints this session with `createSession`, which its twoFactor plugin only
+    // hooks on the sign-in routes - so an enrolled account would get a full session from
+    // the emailed code alone, bypassing its second factor. Verification still stands; the
+    // session does not, and the player signs in through `login` to face the challenge.
+    if (account?.twoFactorEnabled) {
+      await this.drizzle.db
+        .update(session)
+        .set({ expiresAt: new Date() })
+        .where(eq(session.userId, body.user.id));
```
damianrzepka: This expires every session on the account, not just the one better-auth minted a few lines up — the filter is eq(session.userId, …), while the comment above describes killing the session that createSession just issued.

The trigger path is opened by this same PR: with requireEmailVerification now defaulting to false, an unverified player can sign in with their password (covered by the new lets an unverified player sign in while the verification gate is off test), enable 2FA from that session (nothing gates enable2fa on emailVerified), and only then enter the code from the email. Verification succeeds, and they get silently logged out of every device including the one they're on — useVerifyEmail invalidates me right after.

body.token is already read at line 1223, so eq(session.token, body.token) (guarded for a null token) scopes it correctly. The RG/backoffice branch should keep revoking everything — that one is deliberate.

The verifies but does not sign in a 2FA-enrolled account test won't catch this: it registers a fresh account that only ever has one session. Worth extending it with a second pre-existing session that must survive.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1178,68 +1166,113 @@ export class IdentityService {
...
-      );
-      await this.api.sendVerificationEmail({ body: { email }, headers, asResponse: true });
+  /**
+   * Resends the registration code. Unauthenticated by design - the player has no session
+   * until the code is verified - so it answers SUCCESS for every address and only mails a
+   * code to an account that exists and is still unverified. Verified accounts are skipped
+   * deliberately: `verifyEmail` signs the code's bearer in, so resending to a verified
+   * address would turn this into passwordless sign-in for anyone who can read that inbox.
+   */
+  async sendEmailVerification(input: ResendEmailVerificationInput, reqHeaders: NodeHeaders) {
```
damianrzepka, line 1245: This route has no test anywhere in the repo (grepped for sendEmailVerification and email/verify/send). The docstring above states the three properties that make an unauthenticated mail-sending endpoint safe: uniform SUCCESS so it can't be used to enumerate addresses, mail sent only to an account that exists and is still unverified, and the email+IP rate limits. The middle one is load-bearing — since verifyEmail signs the code's bearer in, resending to a verified address turns this into passwordless sign-in for anyone who can read that inbox — and a regression there would be completely silent.

Three cases in registration.e2e.test.ts would cover it: unknown address → 200 and zero mails; verified address → 200 and zero mails; unverified address → exactly one mail.

packages/core/src/contracts/schemas/platform-config.ts
```diff
@@ -124,8 +124,11 @@ export const RegistrationConfigSchema = z
   .object({
     /** Version recorded beside the player's affirmative terms acceptance. */
     termsVersion: z.string().min(1),
-    /** Public consumer origin used in verification-email links. */
-    webUrl: z.url(),
+    /**
+     * Blocks sign-in until the player has verified their address. Off by default:
+     * unverified players stay unrestricted while the KYC toggle is off.
+     */
+    requireEmailVerification: z.boolean().default(false),
```
damianrzepka, line 131: registration.webUrl disappears from a schema that is .strict(), so every operator who has it in their platform config today fails at boot after upgrading. No changeset is in the PR, and CONTRIBUTING.md:82 requires one for any change to published behavior — this one also changes the /identity/email/verify contract (token → email + otp, plus the new output shape) and turns /identity/email/verify/send from a session route into one taking a required email.

I'd argue for major with a one-line migration note rather than minor, given the strict-schema boot failure.

damianrzepka (approved):

PR #103, opened 2026-08-23
feat(wallet): rollover tracking for chat gifts and rain bonus credits

marek-chmielowski-blurify:
BF-326: gift/rain credits create rollover-tracked bonus credits; casino bets advance credits oldest-first and complete them at the configured target. Withdrawals and other player debits cannot spend the proportional locked share until the requirement is met.

## Summary

Gift and rain credits snapshot the Super-Admin multiplier at creation. Bets advance rollover progress transactionally, and completion emits the existing notification/audit flow. Active rollover credits are returned by default; completed history is available with the status filter.

## Risks

Rollover progress is intentionally forward-only: refunds and voided rounds do not reverse previously earned progress in this change and require a separate settlement policy. The multiplier is capped at 100x to prevent accidental permanently locked balances.


marek-chmielowski-blurify:
Will clean from unnecessary comments.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -309,6 +321,54 @@ export async function readWalletBalance(
...
+  return txn
+    .update(walletBalance)
+    .set({ amount: sql`${walletBalance.amount} - ${amount}::numeric` })
+    .where(
+      and(
+        eq(walletBalance.walletId, walletId),
+        eq(walletBalance.currency, balanceKey(currency)),
+        gte(
+          sql`${walletBalance.amount} - COALESCE((
+            SELECT SUM(${walletBonusCredit.rolloverRequired} - ${walletBonusCredit.rolloverProgress})
```
zaxovaiko: Hm, this locks `rolloverRequired - rolloverProgress`, which is the wagering requirement, not the bonus itself. With the seeded `1` multiplier the two are equal so every test passes, but set the multiplier to `10` and a 100 gift locks 1000 - the player can no longer withdraw their own deposit.

Shouldn't we lock the unfinished share of `creditedAmount` instead, something like `creditedAmount * (1 - progress / required)`? Same applies to `readLockedBonusAmount` just below.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -91,6 +105,15 @@ export class WalletCommandsService implements WalletCommands {
 
     await this.writeLedgerRow(txn, row, type, amount);
 
+    if (type === 'bet') {
```
zaxovaiko, line 118: Only `withdraw` goes through `debitWithdrawableBalance` - `debit` here still calls `debitWalletBalance`, so gift, rain and tip walk straight past the lock. A player gifts the locked balance to an alt and withdraws from there, and the alt even gets a fresh credit out of it.

Let's put the guard here instead, for every type except `bet` and `loss` - one place, all callers, rather than patching each debit path.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -120,6 +143,161 @@ export class WalletCommandsService implements WalletCommands {
...
+    const multiplier = await this.resolveRolloverMultiplier(txn);
+
+    const [creditRow] = await txn
+      .insert(walletBonusCredit)
+      .values({
+        walletId,
+        userId,
+        currency,
+        sourceType,
+        creditedAmount: amount,
```
zaxovaiko, line 202: This stores `wallet.currency` raw, but `readLockedBonusAmount` and `debitWithdrawableBalance` both match on `balanceKey(currency)` (uppercased). The `wallet` row is inserted with whatever the caller passed, unlike `wallet_balance`. Let's store `balanceKey(currency)` here so a lowercase currency can't silently disable the lock.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -120,6 +143,161 @@ export class WalletCommandsService implements WalletCommands {
...
+        rolloverMultiplier: multiplier,
+        rolloverRequired: creditRow.rolloverRequired,
+      },
+    });
+  }
+
+  private async applyBonusRolloverProgress(
```
zaxovaiko, line 229: Nothing walks `rolloverProgress` back on a refund or a voided round, so the rollover stays earned. Is that intentional for now? Fine either way, just worth calling out in the PR description if so.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -120,6 +143,161 @@ export class WalletCommandsService implements WalletCommands {
...
+          eq(walletBonusCredit.status, 'active'),
+        ),
+      )
+      .orderBy(asc(walletBonusCredit.createdAt), asc(walletBonusCredit.id));
+
+    const completed: CompletedBonusCredit[] = [];
+    let remaining = amount;
+
+    for (const credit of activeCredits) {
+      if (Number(remaining) <= 0) {
```
zaxovaiko: Float compare on money. Safe enough as a break guard, but let's use `moneyToNumber` to stay consistent with the rest of the wallet.

packages/core/src/wallet/__tests__/bonus-rollover-concurrency.int.test.ts
```diff
@@ -0,0 +1,283 @@
...
+      return true;
+    }
+    if (cause instanceof Error && cause.message.includes('deadlock detected')) {
+      return true;
+    }
+  }
+  return false;
+}
+
+async function withDeadlockRetry<T>(fn: () => Promise<T>, retries = 6): Promise<T> {
```
zaxovaiko: `withDeadlockRetry` tells me we genuinely hit `40P01` between the bet and withdraw paths, but nothing retries in the runtime - a player just gets a 500. Can we align the lock order between the two paths, or move the retry to the service boundary? A test-only retry hides it rather than fixing it.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1645,6 +1731,67 @@ export class WalletService {
...
+  ): Promise<BonusRolloverConfig> {
+    return this.drizzle.db.transaction(async (txn) => {
+      const [before] = await txn
+        .select()
+        .from(walletBonusRolloverConfig)
+        .where(eq(walletBonusRolloverConfig.singletonKey, 'global'));
+      const rows = await txn
+        .insert(walletBonusRolloverConfig)
+        .values({ singletonKey: 'global', multiplier, updatedBy: adminId })
+        .onConflictDoUpdate({
```
zaxovaiko, line 1838: `$onUpdateFn` doesn't apply to `onConflictDoUpdate.set`, so `updatedAt` keeps its insert value forever and we hand that back in the DTO. Let's set it explicitly here. `walletAutoWithdrawalConfig` has the same bug, but that's for later.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1645,6 +1731,67 @@ export class WalletService {
     });
   }
 
+  async getBonusRolloverStatus(userId: User['id']): Promise<{ credits: BonusCredit[] }> {
+    const rows = await this.drizzle.db
+      .select()
+      .from(walletBonusCredit)
+      .where(eq(walletBonusCredit.userId, userId))
+      .orderBy(desc(walletBonusCredit.createdAt))
+      .limit(50);
```
zaxovaiko, line 1805: Returns the last 50 credits of any status, no filter, no paging - a player with some history can't see their active ones. Shouldn't we filter to `active`, or take a status param?

packages/core/src/wallet/contract/index.ts
```diff
@@ -219,6 +219,43 @@ export const SetWalletAutoWithdrawalConfigInputSchema = z.object({
...
+});
+
+export const BonusRolloverConfigSchema = z.object({
+  id: UuidSchema,
+  multiplier: MoneyAmountSchema,
+  updatedAt: TimestampSchema,
+});
+export type BonusRolloverConfig = z.infer<typeof BonusRolloverConfigSchema>;
+
+export const SetBonusRolloverConfigInputSchema = z.object({
```
zaxovaiko, line 283: Positive but unbounded - a typo'd `100000` makes every bonus permanently unwithdrawable. Maybe worth a sane upper cap here.

zaxovaiko (changes requested):
Nice work on the tests here - the repeated-trial concurrency probes against real PG are exactly the right level of paranoia for this, and keeping all the progress arithmetic in Postgres numeric instead of JS is good.

Two blockers though. The withdrawal lock holds the wagering requirement rather than the bonus principal, so anything above a `1` multiplier locks the player's own money, and the lock only covers `withdraw` - gift and rain still drain it. Both are exploitable with the feature working exactly as written. The rest is minor.

packages/core/src/wallet/service/wallet.service.ts
marek-chmielowski-blurify: Resolved: the lock now uses the unfinished proportional share of creditedAmount, so a 100-unit gift at 10x locks at most 100 units, not 1000; readLockedBonusAmount uses the same formula.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Resolved: every real debit except bet goes through debitWithdrawableBalance, so gifts, rain and tips cannot bypass the lock.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Resolved: bonus credits now store balanceKey(currency), matching all lock queries and preventing lowercase currencies from bypassing the lock.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Intentional for this scope: rollover progress is forward-only, so refunds and voids do not reverse it; reversal requires a separate settlement policy.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Resolved: the guard now compares money via moneyToNumber(remaining) instead of a direct floating-point comparison.

packages/core/src/wallet/__tests__/bonus-rollover-concurrency.int.test.ts
marek-chmielowski-blurify: Resolved: bet and withdrawal paths acquire the wallet row lock first, and the test no longer uses a test-only deadlock retry.

packages/core/src/wallet/service/wallet.service.ts
marek-chmielowski-blurify: Resolved: updatedAt is explicitly set in onConflictDoUpdate, so the returned config contains the actual update timestamp.

packages/core/src/wallet/service/wallet.service.ts
marek-chmielowski-blurify: Resolved: the endpoint defaults to active and accepts an explicit status filter for completed history.

packages/core/src/wallet/contract/index.ts
marek-chmielowski-blurify: Resolved: schema validation now caps the multiplier at 100x, preventing accidentally permanent locks.

zaxovaiko (approved):

PR #104, opened 2026-08-23
feat(wallet): player withdrawal address book

zaxovaiko:
## Summary

Adds a player-facing withdrawal address book to the `wallet` module: `walletContract.withdrawalAddresses.{list,create,delete}`, backed by a new `wallet_withdrawal_address` table. Players save labelled payout destinations per currency and network and pick one at withdrawal time instead of retyping an address. Refs BF-408.

## Why

`walletContract` had no address-book surface, so a downstream operator building the "saved wallets" screen had to implement it as an overlay: its own contract package, table, migration set and service, duplicated by every operator who wants the same screen. Saving a payout destination is platform behaviour, not operator-specific, so it belongs in core.

## Alternatives considered

- **Leave it in the operator overlay.** Rejected: every operator shipping a crypto withdrawal flow needs the same thing, and each overlay copy owns its own table and migration journal.
- **Accept a saved-address id on `wallet.withdraw`** as an alternative to `destinationAddress`, so the server resolves the address itself and a tampered client cannot swap it after selection. Not done: the spec's withdrawal flow only pre-fills the form, and this keeps the `withdraw` contract unmoved. Worth revisiting as a hardening step on its own.
- **A confirmation or cooling-off window before a new address can receive a payout** (a common AML control). Not done: no such requirement in the spec for this feature. Calling it out as a deliberate gap rather than an oversight, since it is the kind of control a licensing regime can ask for later.
- **A separate `WithdrawalAddressService`.** Rejected: the three methods need only `drizzle` and `audit`, both already injected into `WalletService`, so a new service would have meant new DI wiring for no separation gained.

## Risks

- **New table, additive migration** (`0011`). Creates `wallet_withdrawal_address` and its indexes; touches no existing table, so rollout is forward-only with no backfill and no downtime.
- **`network` is required and normalised to upper case**, and `currency` to upper case, reusing the module's existing `WalletNetworkInputSchema` / `WalletCurrencyInputSchema`. A client that stores network keys in lower case must fold case on read.
- **Abuse limits are two layers**: a 50-row per-player cap, plus both writes consuming the existing per-user wallet mutation budget (30/min, shared with deposit and withdraw). The cap alone would not bound writes, since deleting frees a slot and lets a client churn create/delete forever. The cap itself is count-then-insert, so concurrent creates can land a few over it - marked in code with the upgrade path. It bounds storage abuse, it is not a security control.
- **Audit entries deliberately omit the address string** (they record label, currency and network only), because the audit log is admin-readable and a player's saved addresses are not. This means an admin-side investigation cannot recover which address was saved from the audit trail alone. Flagged as a deliberate trade-off; easy to reverse if compliance wants the opposite.
- **`delete` returns `false` rather than throwing** for both a foreign row and a nonexistent one, matching the sibling `autoWithdrawalRules.delete` and keeping the id space unprobeable.


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2299,4 +2346,199 @@ export class WalletService {
...
+        network: walletWithdrawalAddress.network,
+        providerName: walletWithdrawalAddress.providerName,
+        providerWalletId: walletWithdrawalAddress.providerWalletId,
+      })
+      .from(walletWithdrawalAddress)
+      .where(
+        and(
+          eq(walletWithdrawalAddress.userId, userId),
+          eq(walletWithdrawalAddress.currency, currency),
+          eq(walletWithdrawalAddress.address, address),
```
marek-chmielowski-blurify, line 2464: **Possible bug:** `whitelistedWalletId()` matches a saved withdrawal address by exact string equality, but the address is normalized differently on the create path vs. the withdraw path.

- `CreateWithdrawalAddressInputSchema.address` only trims the value.
- `WithdrawInputSchema.destinationAddress` is a bare optional string with no trim/case transform (unlike `currency`/`network`, which are uppercased on both paths).

If a player saves an EVM address in checksummed mixed case and later submits a withdrawal with the same address rendered lowercase (or with different incidental whitespace), `eq(walletWithdrawalAddress.address, address)` won't find the row, and `requireWhitelistedWalletId` will throw `DestinationAddressNotWhitelistedError` for an address the player legitimately saved.

Worth normalizing (trim + lowercase, or per-network canonical form) on both the create and withdraw paths before the equality check.

zaxovaiko: Good catch on the whitespace half - `destinationAddress` was a bare `z.string()` while the address book trims, so a pasted address with a trailing space would miss the row and 4xx as unwhitelisted. Both now share one `WalletAddressInputSchema` in `wallet/contract/index.ts`.

On lowercasing, I'd rather not - base58 rails are case-sensitive (BTC legacy, TRC20, Solana), so a blanket `.toLowerCase()` would corrupt those addresses, and it would change the dedupe semantics of the `(userId, currency, network, address)` unique index too. The mixed-case EVM path only breaks if the player re-types the address by hand instead of picking it from the book, and the proper fix there is a per-network canonical form. But that's for later.

Done in `bbded3fe` - trim on both paths, no case transform.

marek-chmielowski-blurify (approved):

PR #105, opened 2026-08-23
feat(identity): let players list and revoke their own sessions

zaxovaiko:
## Summary

- Adds `identity.sessions.listMine` (`GET /identity/sessions/me`) and `identity.sessions.revokeMine` (`POST /identity/sessions/me/revoke`), the self-service twins of the admin-only session routes, scoped to the caller's verified session instead of a `userId` input.
- `SessionItemSchema` gains `updatedAt` (last used) and `current`, so a device list can mark the session making the request as in-use rather than offering to revoke it.
- `SessionResolver.resolveSession` now returns the session id alongside the user id, `createApp` publishes it as `AuthContext.sessionId`, and `@openora/core/server` exports a `getSessionId(context)` helper to read it.
- Adds `useMySessions` and `useRevokeMySession` to `@openora/core/pam/react`.

## Why

BF-481. A consumer building the usual "Sessions" panel in account settings has no route to call: `identity.sessions.list` and `identity.sessions.revoke` both sit behind `AdminGuard`, so a player can neither see the devices their account is signed in on nor sign out the ones they do not recognise. The only workaround was a full password reset, which revokes everything.

## Alternatives considered

- Reusing the admin routes with a "caller may pass their own `userId`" carve-out inside `AdminGuard`. Rejected - it makes the guard's contract conditional on the input, and every future admin route inherits the ambiguity. Separate routes keep the authz rule one line long in each handler.
- Deriving the current session in the service by re-reading the cookie. Rejected - the runtime already resolves the session once per request, so this would be a second `getSession` round trip for a value we are throwing away.

## Risks

- `SessionItemSchema` gains two required output fields. Additive on the wire, but a consumer pinning the schema shape will need the version bump. The admin `sessions.list` reports `current` too, so an admin looking at their own row sees it correctly rather than always `false`.
- `AuthContext.sessionId` is optional, and `resolveSession` tolerates a `getSession` response without a `session` object, so a partially stubbed auth in a downstream test keeps working - the id is just absent and nothing is marked current.
- `revokeMine` filters on the caller's own `userId`, so a forged session id resolves to `NOT_FOUND`, never a cross-user revoke. Covered by an integration test.
- No migration, no schema change - `session.updatedAt` was already there.


packages/core/src/pam/identity/router/index.ts
```diff
@@ -186,6 +188,27 @@ export function createIdentityRouter(
...
+          sortOrder: input.sortOrder,
+        }),
+      ),
+
+      // Scoped to the caller's own userId, so a forged id cannot kill someone
+      // else's device - a miss is a 404, not a cross-user revoke.
+      revokeMine: os.sessions.revokeMine.handler(({ input, context }) => {
+        const userId = getUserId(context);
+        return mapErrors({ NOT_FOUND: SessionNotFoundError }, () =>
+          sessionSvc.revokeSession(userId, input.id, userId, context.clientMeta),
```
klaudia-blazyczek-blurify, line 209: Please do not pass the player's user ID as `actorId` here. The audit mapper treats any `actorId` as an admin action, so this self-service revoke is stored as an admin revoke with the wrong actor. Leave `actorId` unset for this route, or add an explicit actor type.

packages/core/src/pam/identity/contract/index.ts
```diff
@@ -37,8 +37,13 @@ export const SessionItemSchema = z.object({
   id: UuidSchema,
   expiresAt: TimestampSchema,
   createdAt: TimestampSchema,
+  // Last time better-auth refreshed the session - "last used" in a device list.
```
klaudia-blazyczek-blurify, line 40: `updatedAt` is presented as ‘last used’, but revoking a session updates this row and its `$onUpdateFn` changes the timestamp to the revocation time. The admin list will then show a wrong last-used value. Please preserve the original timestamp on revoke, or expose this as last-changed instead.

packages/core/src/pam/identity/react/account.ts
```diff
@@ -72,3 +72,17 @@ export function useSendEmailVerification() {
     ...utils.sendEmailVerification.mutationOptions(),
   });
 }
+
```
klaudia-blazyczek-blurify, line 86: This is a published React SDK hook, so its return type needs to be explicit. Please export a result type and annotate both new hooks; otherwise a query-options refactor can silently change the downstream API.

packages/core/src/pam/identity/router/index.ts
agniev-a-hub: Screenshot 2026-08-24 at 08 34 41

ActorId is correct. ActorId - ID of the user launched the scenario. 

Screenshot 2026-08-24 at 08 35 41

And audit topic handle is questionable, we always want to have either actor and target ids, regardless of admin/user action fired.

zaxovaiko: Kept `actorId` here - it is the user who fired the action, and we want an actor on every audit row. Fixed the mapper instead in `9f4cac6a`: `audit/plugin.ts` treated any `actorId` as an admin, now only a *different* one is. Self-revoke stays `actorType: player` with the resolved `playerId`.

zaxovaiko: Agreed, the topic handling was the actual bug. Done in `9f4cac6a` - `isForced` now compares `actorId` against `userId` instead of just checking presence, so both actor and target ids are always there.

packages/core/src/pam/identity/contract/index.ts
zaxovaiko: Good catch. Done in `9f4cac6a` - both revoke paths in `identity/session.service.ts` now write `updatedAt: session.updatedAt`, so `$onUpdateFn` cannot overwrite the last-used time. Covered by an integration test.

packages/core/src/pam/identity/react/account.ts
zaxovaiko: Done in `9f4cac6a` - exported `UseMySessionsResult` and `UseRevokeMySessionResult` and annotated both hooks. Needed `Paginated<T>` in `contracts/kit.ts` and `IdentitySuccess` in `schemas/identity.ts` to say it without hand-writing the shape.

PR #106, opened 2026-08-23
fix(wallet): close authz, deposit-gating and audit gaps found in the custody review

zaxovaiko:
## Summary

Closes every finding from the static review of BF-410, BF-455, BF-456 and BF-457: one privilege-escalation blocker, one unenforced money path, and the audit gaps that stopped the custody work from being defensible to a regulator.

## Why

Manual balance adjustment was not Super-Admin-only, which BF-410 requires. `player: read_write` expanded to every action in the `player` tuple, `adjust-balance` among them, so four seeded non-super roles - `admin`, `risk-fraud-manager`, `support-manager`, `vip-manager` - could credit any player's real balance. `AdminGuard` had no super-admin notion at all.

BF-457 made `depositEnabled` and `minDeposit` operator-editable, but neither was read on any deposit path: `getOrCreateDepositAddress`, `deposit()` and `creditDepositByAddress` never touched `wallet_asset`. A player could get a live deposit address for a pair the operator had disabled. The withdrawal side was already enforced correctly.

`deposit.idempotencyKey` was optional, and without it `insertIdempotentTransaction` skipped the unique-index guard entirely - a client retry after a timeout double-charged the PSP and double-credited the ledger. `withdraw` had the same hole on the payout side.

On auditability: sweeps were recorded once per cycle with no per-transfer entry, reconciliation findings were not audited at all, a reconciliation run with no config reported `completed` while doing nothing, and `audit_log` immutability was convention plus a hash chain with no write-path control behind it.

## What changed

**Authorization.** `adjust-balance` leaves the permission statement entirely, so no matrix cell can grant it. The route calls a new `AdminGuard.assertSuperAdmin`, backed by `AdminPermissionResolver.isSuperAdmin` - `null` means no DB assignment and falls back to the static `admin` role, matching the bootstrap clause the iam module already used for its own super-admin checks. Denials emit `identity.user.unauthorized_access` like every other denial.

**Deposit gating.** `assertDepositAllowed` and `assertAboveMinimumDeposit` mirror the withdrawal helpers and run in `deposit()` before the vendor call and in `getOrCreateDepositAddress` before the cache read. A currency with no catalog row passes through unchecked (a fiat PSP currency); an explicit network matching no row is rejected rather than silently accepted. Funds already sent to a previously issued address still credit - the money is real and refusing it strands it.

**Provider routing.** Withdrawal settlement and deposit-address issuance resolve the adapter from `wallet_asset.providerName` instead of always using the default binding. `DEFAULT_PAYMENT_PROVIDER` keeps using the bound `PAYMENT_ADAPTER`, so a single-vendor operator sees identical behaviour.

**Auditability.**
- `audit_log` is append-only at the database: a new `postSql` hook on `runMigrations` installs a `BEFORE UPDATE OR DELETE` trigger. `TRUNCATE` is deliberately still allowed so test teardown and retention work.
- One audit entry per swept balance plus the cycle entry, written inside `finishRun`'s transaction so the hash-chain advisory lock is taken once rather than per sweep.
- Every reconciliation finding is audited, with `address`/`tag`/`txHash` stripped - the finding row carries those for the operator, the hash-chained log does not need raw addresses.
- A run with no configuration records `skipped`, never `completed`.

**Also:** `wallet.treasuryRef` in config and on `sweepToPool`, with `poolRef` falling back to it; a `sweepDustThreshold` column so raising a deposit minimum no longer silently changes what gets swept; batch caps on `reconcileStuckWithdrawals`, `reconcileStuckSweeps` and `resolveInFlightSweeps`; asset deletion locks holding balances `FOR UPDATE` and is blocked while deposit addresses exist for the pair; a sign guard in `fromMinorUnits`; job workers throw instead of silently no-oping when a job arrives before the router factory has run; and the manual-adjustment reason is visible on the player's own transaction row.

## Alternatives considered

Keeping `adjust-balance` as a resource and giving it to super-admin only. Rejected: the cell stays grantable in the backoffice UI while being denied at runtime, which is worse than not offering it.

Defaulting `idempotencyKey` server-side rather than requiring it. Rejected: a server-generated key is unique per request, so it provides no retry protection at all.

Leaving the treasury vault to an overlay. Rejected after discussion: the destination of player funds and the source of payouts are exactly what a regulator asks about, and retrofitting it after go-live is expensive.

## Risks

Breaking for consumers: `idempotencyKey` is now required on `POST /wallet/deposit` and `POST /wallet/withdraw`. A client that omits it gets a 400.

`AdminPermissionResolver` gains a required `isSuperAdmin` method - any overlay binding its own resolver must implement it.

`sweepToPool` gains an optional `treasuryRef` in its options object; existing adapters keep compiling.

One migration adds `wallet_asset.sweep_dust_threshold` (nullable, falls back to `minDeposit`) and a `skipped` value on the job-run status enum.

The `audit_log` trigger blocks UPDATE and DELETE for every caller, including future retention or pruning jobs, which must disable it explicitly.

Any environment already running this code may have manual adjustments made by a non-super-admin - worth querying `audit_log` for `wallet.manual_adjustment.created` before deploying.

Merge order: this branch first, then rebase #104 and #105 - both touch `wallet.service.ts`.


klaudia-blazyczek-blurify (commented):
Comments below.

packages/core/src/wallet/router/index.ts
```diff
@@ -144,14 +146,19 @@ export function createWalletRouter({
...
-          userId: getUserId(context),
-          amount: input.amount,
-          currency: input.currency,
-          provider: input.provider,
-          idempotencyKey: input.idempotencyKey,
-        }),
+      mapErrors(
+        {
+          BAD_REQUEST: [UnsupportedNetworkError, BelowMinimumDepositError],
+          CONFLICT: [IdempotencyKeyReuseError, DepositDisabledError],
```
klaudia-blazyczek-blurify: `DepositDisabledError` is mapped to `CONFLICT` (409) here and also in the `getAddress` handler below (line ~422). A 409 CONFLICT normally signals a state conflict such as idempotency key reuse — clients that branch on status codes will misread a permanent policy rejection as a transient conflict and retry it. `BAD_REQUEST` (400) is the right code for both cases: the request itself is invalid because the operator has disabled this currency/network.

packages/core/src/wallet/service/custody-sweep.service.ts
```diff
@@ -476,16 +486,35 @@ export class CustodySweepService {
   ): Promise<void> {
     const error =
       err === undefined ? {} : { error: err instanceof Error ? err.message : String(err) };
+    const moved = await this.drizzle.db
+      .select()
+      .from(walletCustodySweep)
+      .where(and(eq(walletCustodySweep.runId, runId), eq(walletCustodySweep.status, 'processing')));
+
     await this.drizzle.db.transaction(async (txn) => {
       await txn
```
klaudia-blazyczek-blurify, line 506: The `SELECT` for `processing` sweeps runs outside the transaction that writes the audit entries. If a concurrent `resolveInFlightSweeps` path moves another sweep row to `processing` between this read and the `txn.update(walletJobRun, ...)` call, that row is silently absent from the per-sweep audit chain. Move the `SELECT` inside the transaction (before the `for` loop) so the read and the writes are atomic under the same snapshot.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -734,6 +829,7 @@ export class WalletService {
     idempotencyKey?: string;
   }): Promise<TransactionResult> {
     await this.rateLimit(userId);
+    await this.assertDepositable(currency, amount);
```
klaudia-blazyczek-blurify, line 867: `assertDepositable(currency, amount)` is called without a `network` argument. The `deposit()` method accepts a `provider` field but not an explicit `network`. If the caller can specify a network via `provider` routing, or if the PSP call itself resolves a specific network, the gate is bypassed for a disabled network because `assertDepositAllowed` checks `assets.some((asset) => asset.depositEnabled)` when `network` is undefined — any single enabled network clears the gate even if the target network is off. This is a narrower gap than the original unenforced path but still inconsistent with how `getOrCreateDepositAddress` enforces network-specific gating. Confirm this is intentional for the fiat PSP case or add a note in the doc comment.

packages/core/src/server/db/migrate.ts
```diff
@@ -148,6 +154,9 @@ export async function runMigrations(opts: RunMigrationsOptions) {
     } finally {
       lockClient.release();
     }
+    for (const sql of opts.postSql ?? []) {
```
klaudia-blazyczek-blurify: `postSql` runs after the advisory lock has been released. In a scenario where two migration processes race — one has just dropped the lock and the other acquires it before the first finishes `postSql` — the second process will run its own `postSql` concurrently with the first's. For the trigger this is harmless (`CREATE OR REPLACE` + `DROP IF EXISTS` are idempotent), but the design means `postSql` does not share the migration lock's mutual exclusion guarantee. If a future caller adds a non-idempotent `postSql` statement, this will silently double-execute it. Worth noting in a doc comment on `postSql` in `RunMigrationsOptions`.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -1980,11 +2080,24 @@ export class WalletService {
...
+        .select({ n: count() })
+        .from(walletDepositAddress)
+        .where(
+          and(
+            eq(walletDepositAddress.currency, currency),
+            eq(walletDepositAddress.network, network),
+          ),
+        );
+      if ((issued?.n ?? 0) > 0) {
         throw new WalletAssetInUseError();
```
klaudia-blazyczek-blurify, line 2158: The `walletBalance` check above uses `.for('update')` to take a row lock, but the `walletDepositAddress` count here does not. If `getOrCreateDepositAddress` runs concurrently inside its own transaction and inserts a new deposit address row between this count and the eventual DELETE of the asset, the delete goes through while a live deposit address still exists for that (currency, network) pair. Consider adding `.for('update')` on the deposit address query as well, or alternatively skipping the count in favor of a foreign-key constraint that would block the delete at the DB level.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -293,6 +307,57 @@ export function assertAboveMinimumWithdrawal(
...
+  minDeposit: string;
+  depositEnabled: boolean;
+};
+
+export function assertDepositAllowed(
+  assets: readonly DepositAsset[],
+  currency: string,
+  network?: string,
+): void {
+  if (assets.length === 0) {
```
klaudia-blazyczek-blurify, line 322: An empty catalog is accepted before checking `network`. If the asset is deleted while address issuance is in flight, the second gate passes and stores an address for the deleted pair. Allow an empty catalog only when `network` is undefined; otherwise throw `UnsupportedNetworkError`.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -649,6 +749,15 @@ export class WalletService {
     return row?.providerName ?? DEFAULT_PAYMENT_PROVIDER;
   }
 
+  private async paymentAdapterFor(currency: string, network: string | null) {
+    const providerName = await this.providerNameFor(currency, network);
+    const adapter =
+      providerName === DEFAULT_PAYMENT_PROVIDER
+        ? this.payment
+        : (this.paymentProviders.get(providerName)?.adapter ?? this.payment);
```
klaudia-blazyczek-blurify: A missing named provider falls back to the default adapter. After a provider is removed or misconfigured, an approved withdrawal can be sent through the wrong vendor while still recording the old provider name. Fail closed for a missing non-default provider.

packages/core/src/wallet/service/custody-sweep.service.ts
```diff
@@ -434,7 +440,11 @@ export class CustodySweepService {
     // (docs/standards/money.md: treat external payment calls as non-transactional).
     // The claim row above already holds the durable guard.
     try {
-      const result = await adapter.sweepToPool?.(balance, { idempotencyKey: claimed.id });
+      const treasuryRef = this.platformConfig?.wallet?.treasuryRef;
```
klaudia-blazyczek-blurify: `treasuryRef` has no provider scope but is sent to every provider. With two vendors, a treasury account ID for provider A can be sent to provider B and route custody funds to the wrong account. Store the reference per provider or let each adapter choose it.

packages/core/src/wallet/service/custody-sweep.service.ts
```diff
@@ -448,7 +458,7 @@ export class CustodySweepService {
         .set({
           status: 'processing',
           externalId: result.externalId,
-          poolRef: result.poolRef ?? null,
+          poolRef: result.poolRef ?? treasuryRef ?? null,
```
klaudia-blazyczek-blurify: `poolRef` is the recorded destination, but this stores the requested `treasuryRef` when the adapter does not return one. The adapter contract does not prove it used that reference, so the immutable audit record can be wrong. Keep this null or record the requested reference separately.

packages/core/src/wallet/service/custody-sweep.service.ts
```diff
@@ -267,7 +267,9 @@ export class CustodySweepService {
           inArray(walletCustodySweep.status, IN_FLIGHT_STATUSES),
           isNotNull(walletCustodySweep.externalId),
         ),
-      );
+      )
+      .orderBy(asc(walletCustodySweep.createdAt))
+      .limit(batchSize);
```
klaudia-blazyczek-blurify, line 274: Old `processing` sweeps that remain unresolved stay at the start of this fixed batch. They are retried every cycle while later in-flight sweeps never get a status check. Page or rotate the candidates.

packages/core/src/wallet/service/reconciliation.service.ts
```diff
@@ -541,7 +548,9 @@ export class ReconciliationService {
           eq(walletTransaction.type, 'withdrawal'),
           lt(walletTransaction.createdAt, cutoff),
         ),
-      );
+      )
+      .orderBy(asc(walletTransaction.createdAt))
+      .limit(batchSize);
```
klaudia-blazyczek-blurify, line 564: This fixed oldest-first batch can starve newer stuck withdrawals. Rows that remain `processing` are selected every run, and their deduplicated findings do not remove them, so later rows never produce a finding. Exclude already reported rows or page the backlog.

packages/core/src/wallet/service/reconciliation.service.ts
```diff
@@ -613,21 +631,27 @@ export class ReconciliationService {
           inArray(walletCustodySweep.status, ['pending', 'unknown']),
           lt(walletCustodySweep.createdAt, cutoff),
         ),
-      );
+      )
+      .orderBy(asc(walletCustodySweep.createdAt))
+      .limit(batchSize);
```
klaudia-blazyczek-blurify, line 655: The same oldest-first limit can starve later `pending` or `unknown` sweeps. The first unresolved rows keep occupying this batch after their findings are deduplicated, so operators never see later stuck sweeps. Exclude reported rows or page the backlog.

PR #107, opened 2026-08-24
fix(kernel): issue with stream cleanup

agniev-a-hub:
BF-431

  - Makes createEventStreamGenerator abort cleanup immediate and idempotent.
  - Prevents chat subscriptions from delivering messages after disconnect, including during async block-list loading.
  - Adds regression coverage for aborting a generator while it is waiting for the next event.


zaxovaiko:
Fix itself looks right, and nice that it lives in the helper so `streamSession` and the kyc stream get it for free.

Two things though.

Hm, the new test in `kernel/__tests__/helpers.test.ts` passes on the old code too - I ran it against `dev`. Aborting while the generator waits on the `await` always resumed the body and hit the `finally`, so it does not really lock anything in. The case that actually leaked is the generator parked on `yield`: `wake()` is a no-op there, the body never resumes, `finally` never runs. Lets cover that one instead:

```ts
const gen = createEventStreamGenerator<number>(
  (push) => {
    queueMicrotask(() => push(1));
    return unsubscribe;
  },
  { signal: controller.signal },
);
await gen.next();
controller.abort();
await new Promise((r) => setTimeout(r, 10));
expect(unsubscribe).toHaveBeenCalledOnce();
```

0 calls before the fix, 1 after. Would be good to cover the new `signal.aborted` branch too, since it is new code with nothing holding it.

And in `chat/chat.service.ts` - `publish` snapshots subscribers with `Array.from`, so we can still get a message after `unsubscribe()` ran, and the `pending.push` path is not guarded by `active`. Lets move the check into the transport callback instead of into `deliver`, then `pending.length = 0` in the teardown is not needed either.

agniev-a-hub:
@zaxovaiko Moved callback to the transport, test updated, please have a look.

zaxovaiko (approved):

PR #108, opened 2026-08-25
feat(wallet): destination tags, ledger direction and a realtime balance stream

zaxovaiko:
## What

Four commits on the wallet money path. The first two were already on the branch; the last two are new.

- `feat(wallet): carry a destination tag through the withdrawal path`
- `feat(wallet): show the destination tag on the withdrawal queue`
- `feat(wallet): record a direction on every ledger row and stream balance changes`
- `fix(server): keep sse responses untransformed end to end`

## Ledger direction

`wallet_transaction` gains a `direction` column. It used to be inferred from `type`, which is fine for eight of them and wrong for three: `gift`, `rain` and `tip` write the **same** type for the sender's debit leg and the recipient's credit leg, so a row of one of those types had no recoverable direction at all.

`writeLedgerRow` now takes `direction` as a required parameter with no default. Both social-transfer legs route through it, so the compiler enforces this rather than a convention. `insertIdempotentTransaction` is narrowed the same way.

Migration `0014` backfills only the eight unambiguous types. Historical `gift`/`rain`/`tip` rows are left NULL deliberately: guessing would fabricate money-direction data that was never recorded. Consumers render an unknown direction with no sign.

## Balance stream

`wallet.streamBalance` publishes `{ eventId, currency, reason }` over an event iterator, with **no amount on the wire**. Carrying the number would make the stream itself a money surface, where a dropped or reordered frame leaves a stale balance and nothing to correct it. A signal lets the client refetch instead, which is cheap and self-correcting.

The channel is keyed on the session's user id, never on an input parameter.

## SSE responses opt out of transformation

Found while measuring the stream above: an end-to-end balance push took **11.8 seconds**, arriving in batches. Against the API directly it was **87ms**.

The cause is compression, not the transport. An SSE body is a sequence of small frames that only mean anything on arrival, and Next's built-in gzip held them until its own flush threshold. nginx and a CDN buffer the same way by default.

`text/event-stream` responses now carry `Cache-Control: no-store, no-transform` (the standards-defined opt-out, which the `compression` middleware Next bundles honours) plus `X-Accel-Buffering: no` for nginx's separate proxy-buffering stage. Ordinary responses keep plain `no-store` and stay compressed.

Measured after the change: 119 / 77 / 55 ms.

## Verification beyond CI

The e2e test database on a machine that migrated before the `0012`/`0013` rewrite in 37bd2d01 carries journal rows with no matching file and fails to migrate. Recreate `oss_igaming_test` before running the integration suites locally.

### DIFF
--- a/packages/core/src/wallet/contract/index.ts
+++ b/packages/core/src/wallet/contract/index.ts
@@ -97,6 +97,12 @@ export const WithdrawInputSchema = z.object({
   provider: z.string().optional(),
   idempotencyKey: UuidSchema.optional(),
   destinationAddress: WalletAddressInputSchema.optional(),
+  // Tag/memo networks (XRP, XLM, EOS, and the exchange deposit addresses on them) carry the
+  // beneficiary in a separate field rather than in the address: one address serves every
+  // account behind it. A payout sent without the tag the player was given arrives
+  // unattributed and has to be recovered by hand, so it is carried end to end instead of
+  // being dropped between the request and the custodian.
+  destinationTag: z.string().trim().min(1).max(64).optional(),
 });
 
 export const MANUAL_ADJUSTMENT_DIRECTIONS = ['credit', 'debit'] as const;
--- a/packages/core/src/wallet/contract/index.ts
+++ b/packages/core/src/wallet/contract/index.ts
@@ -1,4 +1,4 @@
-import { oc } from '@orpc/contract';
+import { eventIterator, oc } from '@orpc/contract';
 import * as z from 'zod';
 import {
   KycStatusSchema,
@@ -60,9 +60,30 @@ export const WalletBalancesSchema = z.object({
   balances: z.array(WalletBalanceSchema),
 });
 
+// A change *signal*, never the balance itself: carrying `{ currency, balance }` on the
+// stream would make the stream a money surface - a dropped or reordered event leaves a
+// stale number on screen with no self-correction. This shape only tells the client WHAT
+// changed; the client re-fetches getBalance(s)/listTransactions, which is cheap and
+// always internally consistent.
+export const WalletBalanceChangeReasonSchema = z.enum(['deposit', 'withdrawal', 'adjustment']);
+export type WalletBalanceChangeReason = z.infer<typeof WalletBalanceChangeReasonSchema>;
+
+export const WalletBalanceUpdateSchema = z.object({
+  eventId: UuidSchema,
+  currency: WalletCurrencyCodeSchema,
+  reason: WalletBalanceChangeReasonSchema,
+});
+export type WalletBalanceUpdate = z.infer<typeof WalletBalanceUpdateSchema>;
+
 export const SetActiveCurrencyInputSchema = z.object({ currency: WalletCurrencyCodeSchema });
 
 export const ActiveCurrencySchema = z.object({ activeCurrency: WalletCurrencyCodeSchema });
+
+// Shared by a manual adjustment's input direction and a transaction row's recorded
+// direction - a wallet move only ever goes one of these two ways.
+export const MANUAL_ADJUSTMENT_DIRECTIONS = ['credit', 'debit'] as const;
+export const ManualAdjustmentDirectionSchema = z.enum(MANUAL_ADJUSTMENT_DIRECTIONS);
+export type ManualAdjustmentDirection = z.infer<typeof ManualAdjustmentDirectionSchema>;
 
 export const WalletTransactionSchema = z.object({
   id: UuidSchema,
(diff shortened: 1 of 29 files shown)


packages/core/src/wallet/plugin.ts
```diff
@@ -185,6 +274,7 @@ export default {
         reconciliation,
         jobQueue,
         reconciliationQueue: WALLET_RECONCILIATION_QUEUE,
+        realtime: realtimeTransport,
         limiter: c.get(RATE_LIMITER),
```
klaudia-blazyczek-blurify, line 278: **Type safety gap — `realtimeTransport` is `null` at the declaration site but passed as non-nullable.** `WalletRouterDeps.realtime` is typed `RealtimeTransport` (not `| null`), yet the closure variable is `RealtimeTransport | null`. TypeScript accepts this only because `c.get(REALTIME_TRANSPORT)` is assigned on the very first line of the same factory callback, so the value is guaranteed non-null by the time it arrives here. That ordering is correct today but is not enforced by the type system — reordering those two statements would compile and fail at runtime. A `realtimeTransport!` non-null assertion makes the precondition explicit and matches how `reconciliationRef` is handled (asserted via `?.` at every call-site rather than assumed non-null). Alternatively, assign to a `const` inside the factory and pass it directly.

packages/core/src/wallet/plugin.ts
```diff
@@ -60,6 +64,90 @@ export default {
...
+
+    ctx.events.on('wallet.deposit.completed', (payload, envelope) => {
+      const parsed = domainEventSchemas['wallet.deposit.completed'].safeParse(payload);
+      if (!parsed.success || !envelope) {
+        return;
+      }
+      publishBalanceChanged(parsed.data.userId, parsed.data.currency, 'deposit', envelope.eventId);
+    });
+    ctx.events.on('wallet.withdrawal.requested', (payload, envelope) => {
+      const parsed = domainEventSchemas['wallet.withdrawal.requested'].safeParse(payload);
```
klaudia-blazyczek-blurify, line 103: **Missing balance-change signal for gift / rain / tip transfers.** The plugin wires `wallet.deposit.completed`, `wallet.withdrawal.*`, and `wallet.manual_adjustment.created`, but there is no handler for any event emitted when a social transfer settles. Both the sender and the recipient experience a real balance change; neither session would receive a `streamBalance` push today. If a domain event for social transfers exists, wire it here. If not yet, add a `// TODO` so the gap is not missed when that event lands.

packages/core/src/wallet/contract/index.ts
```diff
@@ -97,12 +123,14 @@ export const WithdrawInputSchema = z.object({
   provider: z.string().optional(),
   idempotencyKey: UuidSchema.optional(),
   destinationAddress: WalletAddressInputSchema.optional(),
+  // Tag/memo networks (XRP, XLM, EOS, and the exchange deposit addresses on them) carry the
+  // beneficiary in a separate field rather than in the address: one address serves every
+  // account behind it. A payout sent without the tag the player was given arrives
+  // unattributed and has to be recovered by hand, so it is carried end to end instead of
+  // being dropped between the request and the custodian.
+  destinationTag: z.string().trim().min(1).max(64).optional(),
 });
```
klaudia-blazyczek-blurify, line 139: **`destinationTag` constraints do not reflect tag-chain wire rules.** XRP memo tags are unsigned 32-bit integers (0–4 294 967 295, max 10 digits); `max(64)` is too permissive for XRP, and there is no digit-only guard. If input validation is intentionally deferred to the adapter (which is the only side that knows the chain's tag format), document that explicitly with a comment here so a future contributor does not tighten this schema in isolation and accidentally break non-XRP chains.

packages/core/src/wallet/drizzle/migrations/0014_abandoned_madripoor.sql
```diff
@@ -0,0 +1,20 @@
...
+  WHEN 'win' THEN 'credit'
+  WHEN 'bonus' THEN 'credit'
+  WHEN 'manual_credit' THEN 'credit'
+  WHEN 'withdrawal' THEN 'debit'
+  WHEN 'bet' THEN 'debit'
+  WHEN 'loss' THEN 'debit'
+  WHEN 'manual_debit' THEN 'debit'
+  ELSE NULL
+END::"wallet_transaction_direction"
+WHERE "direction" IS NULL;
```
klaudia-blazyczek-blurify, line 20: **Full-table `UPDATE` will hold a lock for the entire migration on a large `wallet_transaction` table.** On a table with many rows the lock window could be long enough to affect live traffic during the cut-over. This is fine for a small deployment, but if `wallet_transaction` can grow to millions of rows in production, a batched backfill or an out-of-band script is safer. At minimum, a comment acknowledging the lock trade-off helps whoever coordinates the production migration.

jakubfilinger-b (dismissed):

packages/core/src/wallet/drizzle/migrations/0014_abandoned_madripoor.sql
zaxovaiko: We don't even have a production environment yet 😄. Skipping it for now. Later, when the envs are ready, we will focus on optimizations. Before pushing the v1 release, I think we should squash all migrations (or regenerate them from scratch).

marek-chmielowski-blurify:
NO-GO — `destinationTag` is not covered by the same whitelist as the address.

On networks such as XRP/XLM, many exchange accounts use one shared address, while the tag identifies the specific recipient. A user can have exchange address `A` whitelisted, but submit a tag belonging to another account; the code accepts the withdrawal because it checks only the address, then passes that other tag to the PSP, which is especially dangerous with auto-approval. The tag should be part of the saved/whitelisted destination, or withdrawals with a tag should always require manual verification; if our system supports tags only for deposits, the scope of this feature needs to be clarified first.

PR #109, opened 2026-08-25
Fix/chat commands removal

agniev-a-hub:


damianrzepka:
This PR removes every publisher of chat.gift.sent, chat.gift.claimed, chat.rain.distributed, and chat.donate.sent (confirmed via docs/catalog.json's generated events list, which no longer lists them), but the schema definitions themselves (events.ts L481, L490, L499, L507) and their priority weights (L581, L583, L586) are still there, now orphaned. Not part of this diff, so raise it as a standalone/general PR comment rather than an inline one.

packages/core/src/engagement/chat-commands/AGENTS.md
```diff
@@ -1,17 +1,16 @@
...
 
-Thin chat-command routing surface: the DB-backed command registry (`chat_command_config`),
-`@mention` autocomplete, and dedicated `postGift`/`claimGift`/`getGift`/`postRain` routes. All
-gift/rain/donate money movement, limit checks, and idempotency live in `social-transfers` behind
-the `GIFT_COMMANDS`/`RAIN_COMMANDS` command ports (ADR-0017) - this module never touches a wallet
-or opens a `db.transaction` for money. See `social-transfers/AGENTS.md` for the mechanics; this
-file only covers what actually lives here.
+Chat-command surface: the DB-backed command registry (`chat_command_config`), `@mention`
+autocomplete, and generic command routing. Money-moving implementations for donate, gift, and
+rain are supplied by downstream overlays through the `GIFT_COMMANDS` and `RAIN_COMMANDS` ports;
```
damianrzepka: No longer true, nothing in this module references those tokens anymore.

packages/core/src/engagement/chat-commands/AGENTS.md
```diff
@@ -36,26 +35,22 @@ gets `ChatRoomNotMemberError`, not the gift state. Like every other route on thi
...
 unauthenticated read path.
 
-`postRain` is the one route that does real work here beyond translation: it resolves the online
-recipient list via `CHAT_REALTIME_TRANSPORT.getOnlineUserIds(chatChannel(roomId))` (this module
-owns presence for the whole chat-command surface) BEFORE calling `RAIN_COMMANDS.sendRain` - see
-`social-transfers/AGENTS.md` > "This module never queries chat's presence tracking" for why that
-split exists.
+`postRain` resolves the online recipient list via
+`CHAT_REALTIME_TRANSPORT.getOnlineUserIds(chatChannel(roomId))` (this module owns presence for
+the whole chat-command surface) before calling `RAIN_COMMANDS.sendRain`.
```
damianrzepka: Entire "## Transfer routes" section describing postGift/claimGift/getGift/postRain as live routes; none of them exist in contract/index.ts or router/index.ts anymore.

packages/core/src/engagement/chat-commands/AGENTS.md
```diff
@@ -36,26 +35,22 @@ gets `ChatRoomNotMemberError`, not the gift state. Like every other route on thi
...
 
 ## Ports consumed
 
 - `ADMIN_USER_DIRECTORY` - `findPlayerIds`/`lookupPlayers` for `@mention` autocomplete.
 - `CHAT_BLOCK_WRITER` - `getExcludedUserIds(viewerId)` to filter blocked/ignored players out of
   mention search.
-- `GIFT_COMMANDS` / `RAIN_COMMANDS` - bound by `social-transfers`; this module never implements
-  gift/rain mechanics itself.
+- `GIFT_COMMANDS` / `RAIN_COMMANDS` - required from the downstream transfer overlay.
```
damianrzepka: Same issue, stale.

packages/core/src/contracts/adapters/gift-commands.ts
damianrzepka: Header comment still says "chat-commands posts/claims gifts into the chat stream... behind this port" and points to chat-commands AGENTS.md for the translation logic — that logic was deleted from chat-commands in this PR.

packages/core/src/contracts/adapters/rain-commands.ts
```diff
@@ -1,11 +1,11 @@
...
-// CHAT_REALTIME_TRANSPORT and hands `social-transfers` the plain id list - social-
-// transfers never queries chat's presence tracking directly.
+// CHAT_REALTIME_TRANSPORT and hands the transfer service the plain id list -
+// the transfer service never queries chat's presence tracking directly.
 //
 // Same discriminated-result shape as GIFT_COMMANDS and the same reason it
 // exists: chat-commands' router can't `instanceof`-match an error class
-// defined in social-transfers without a forbidden cross-module internals
+// defined in the transfer service without a forbidden cross-module internals
 // import. ADR-0017.
```
damianrzepka: Same issue: "chat-commands' router can't instanceof-match an error class defined in the transfer service" — chat-commands' router no longer calls this port at all.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
```diff
@@ -1,423 +0,0 @@
-import { describe, it, expect, vi, beforeAll, afterAll } from 'vitest';
```
damianrzepka, line 1: Deleting the whole file also removes describe blocks at old L230 (authz negatives on /backoffice/wallet/bonus-rollover-config — not gift-specific) and old L287 (multi-credit waterfall test — uses gift only as a funding mechanism) plus probes at old L320 and L360. Worth flagging that only the gift-dependent setup needed to change, not the whole file.

agniev-a-hub:
@damianrzepka cleanup performed

packages/core/src/engagement/chat-commands/drizzle/migrations/meta/_journal.json
```diff
@@ -5,15 +5,8 @@
...
-      "tag": "0000_daffy_lady_bullseye",
-      "breakpoints": true
-    },
-    {
-      "idx": 1,
-      "version": "7",
-      "when": 1787009522084,
-      "tag": "0001_sturdy_barracuda",
+      "when": 1787729046139,
+      "tag": "0000_striped_trauma",
```
jakubfilinger-b: This rewrites applied migration history and will hard-fail every existing deployment.

`runMigrations` (`packages/core/src/server/db/migrate.ts`) picks pending work by **content hash**, not by index or timestamp:

```ts
const appliedHashes = new Set(applied.rows.map(({ hash }) => hash));
const pending = readMigrationFiles(migrationConfig).filter((m) => !appliedHashes.has(m.hash));
```

A DB that already ran this module has the hashes of `0000_daffy_lady_bullseye` + `0001_sturdy_barracuda` recorded. `0000_striped_trauma.sql` is new content (it creates only `chat_command_config`, the old 0000 also created `chat_gift`), so its hash is absent → it is treated as pending → `CREATE TABLE "chat_command_config"` runs against a database where that table already exists → `relation already exists`, the tx rolls back, `runMigrations` throws, boot fails. Fresh DBs are fine, which is why CI is green.

The end state you want is already reached on existing DBs: old `0001` dropped `chat_gift`. So nothing new is needed here — please restore `0000_daffy_lady_bullseye.sql`, `0001_sturdy_barracuda.sql` and the original `_journal.json`/snapshots, and drop `0000_striped_trauma`. If a squash is genuinely wanted later it needs a separate baseline strategy, not an in-place rewrite.

packages/core/src/engagement/chat-commands/plugin.ts
```diff
@@ -4,24 +4,20 @@ import {
...
-  RAIN_COMMANDS,
   AUDIT_WRITER,
 } from '@openora/core/contracts';
 import { ChatCommandsService } from './service/chat-commands.service.js';
 import { createChatCommandsRouter } from './router/index.js';
 
 export default {
   id: 'chat-commands',
-  dependsOn: ['chat', 'social-transfers', 'audit'],
+  dependsOn: ['chat', 'iam', 'audit'],
```
jakubfilinger-b: `iam` looks like the wrong module here. The four ports this plugin resolves are `ADMIN_USER_DIRECTORY`, `CHAT_BLOCK_WRITER`, `CHAT_REALTIME_TRANSPORT`, `AUDIT_WRITER`. `ADMIN_USER_DIRECTORY` is provided by **identity**, not iam — `packages/core/src/pam/identity/plugin.ts:44`:

```ts
ctx.provide(
  ADMIN_USER_DIRECTORY,
  (c) => new DrizzleAdminUserDirectory(c.get(DRIZZLE), c.get(EVENT_BUS)),
);
```

`grep ADMIN_USER_DIRECTORY packages/core/src/iam/` returns nothing. Did you mean `['chat', 'identity', 'audit']`? It doesn't blow up today because the router factory resolves lazily, but the declared dependency points at a module that provides nothing this one consumes.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
```diff
@@ -296,18 +266,12 @@ describe('bonus-rollover authz negatives (probe #2)', () => {
 
 describe('bonus-rollover waterfall across multiple simultaneously-active credits (probe #3)', () => {
   it('a single bet drains the oldest active credit first and cascades the leftover into the next, through the real HTTP surface', async () => {
```
jakubfilinger-b: The `it` title still says "through the real HTTP surface", but after this change the credits under test are no longer created through any HTTP surface — `creditBonus` reaches into `appMain.container.get(WALLET_COMMANDS)` and calls `credit()` in-process. Same for probes #1, #2, #4 and #5.

That's a real drop in coverage, not just a wording problem: the previous version proved the bonus-rollover lifecycle was reachable by a player over HTTP. Now nothing does. Either retitle these to say they drive the wallet port directly, or point them at whatever route is supposed to fund a bonus credit after this PR — see my general comment about there no longer being one.

jakubfilinger-b (commented):
Two things that aren't anchored to a line in this diff, plus inline notes below.

**1. Bonus-rollover is left without a producer.**

This is the one I'd want answered before merge.

`wallet-commands.service.ts:161` gates bonus-credit creation on `type === 'gift' || type === 'rain'`, and `wallet/contract/index.ts:266` declares `BONUS_CREDIT_SOURCE_TYPES = ['gift', 'rain'] as const`. This PR deletes every producer of those two transaction types — social-transfers is gone, and chat-commands' `postGift`/`claimGift`/`postRain` are gone.

Grepping the head of this branch for `type: 'gift'` / `type: 'rain'` outside enum declarations returns **only test files**. So after this merges, no route in the platform can create a `wallet_bonus_credit` row, and the whole bonus-rollover feature shipped in #103 becomes unreachable code — including `/wallet/bonus-rollover/status`, the super-admin multiplier config, the auto-release job and the withdrawal block.

That also explains why the e2e test had to be rewritten to bypass HTTP: there is no longer an HTTP path that reaches it.

Is decommissioning bonus-rollover part of the intent here? If yes it should be its own change with its own migration/deprecation story. If not, this PR needs a replacement funding path before the gift/rain producers disappear.

**2. chat-commands still advertises gift/rain/donate.**

Same class of leftover as the event schemas, but on the module's own live surface — three places still advertise commands that no longer have an implementation:

- `engagement/chat-commands/contract/index.ts:11-21` — `CHAT_COMMAND_TYPES` still lists `'gift'`, `'rain'`, `'donate'`, and `CommandConfigSchema` still carries `maxAmount`/`minAmount`/`maxRecipients`, which only ever meant anything for those three.
- `engagement/chat-commands/seed/index.ts` — still inserts all three with `enabled: true` and money config (`minAmount: '1.00000000'`, `maxRecipients: 50`).
- `contracts/schemas/chat-command-metadata.ts:13,29` — `z.literal('gift')` / `z.literal('rain')` message variants.

Net effect: `GET /chat-command/commands` keeps returning gift/rain/donate as enabled to every client, and an admin can still toggle and configure them through `adminUpdateCommand`, with no route behind any of it. None of those files are in this diff, hence a general comment.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
```diff
@@ -55,35 +53,20 @@ async function deposit(client: TestClient, amount: string, currency = 'USD') {
...
-    idempotencyKey: randomUUID(),
+async function creditBonus(client: TestClient, userId: string, amount: string) {
+  await deposit(client, '1');
+  const walletCommands = appMain.container.get(WALLET_COMMANDS);
+  await appMain.container.get(DRIZZLE).db.transaction(async (tx) => {
+    const result = await walletCommands.credit(tx, {
+      userId,
+      amount,
+      currency: 'USD',
+      type: 'gift',
```
jakubfilinger-b, line 71: `type: 'gift'` is load-bearing and non-obvious, and this PR removes the mechanic it is named after. In `wallet-commands.service.ts:161`:

```ts
if (type === 'gift' || type === 'rain') {
  await this.createBonusCredit(txn, { ..., sourceType: type });
}
```

`'gift'` and `'rain'` are the *only* two transaction types that produce a `wallet_bonus_credit` row. Switching this to `'bonus'` or `'manual_credit'` would silently create no credit and every assertion below would fail on an empty `status.credits`.

Please add a comment saying exactly that, otherwise the next cleanup pass sees a `'gift'` in a PR that deleted gifting and "tidies" it.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
```diff
@@ -55,35 +53,20 @@ async function deposit(client: TestClient, amount: string, currency = 'USD') {
...
-  return giftId;
-}
-
-async function claimGift(claimer: TestClient, giftId: string) {
-  await claimer.post('/wallet/deposit', {
-    amount: '1',
-    currency: 'USD',
-    idempotencyKey: randomUUID(),
+async function creditBonus(client: TestClient, userId: string, amount: string) {
+  await deposit(client, '1');
```
jakubfilinger-b, line 60: Worth a one-liner: this deposit is not a leftover from `claimGift`, it's what materialises the wallet row. `credit()` fails closed — `wallet-commands.service.ts:141`:

```ts
// Fail closed: a credit never creates a wallet - a missing one is a caller bug.
if (!row) return { ok: false, reason: 'wallet not found' };
```

Without a note it reads as copy-paste residue. Also flagging that it adds 1 USD of real (non-bonus) balance per call — twice in the waterfall probe — which is fine for the current assertions but will bite anyone who later asserts on total balance.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
```diff
@@ -164,24 +144,18 @@ afterAll(async () => {
...
-    const giftId = await sendGift(sender, '40');
-    const claimResult = await claimGift(recipient, giftId);
-    expect(claimResult.claimedBy).toBe(recipientId);
+    await creditBonus(recipient, recipientId, '40');
 
     let status = await rolloverStatus(recipient);
     expect(status.credits).toHaveLength(1);
     const credit = status.credits[0]!;
-    expect(credit).toMatchObject({ sourceType: 'gift', status: 'active' });
+    expect(credit.status).toBe('active');
```
jakubfilinger-b, line 165: `expect(credit).toMatchObject({ sourceType: 'gift', status: 'active' })` became `expect(credit.status).toBe('active')` — the `sourceType` assertion was dropped rather than updated. The credit still *has* a `sourceType` (`'gift'`, from the `type` passed at line 64), so this can assert the real value instead of dropping coverage. Same at line 349, where `sourceType: 'gift'` was removed from the audit-payload `toMatchObject`.

packages/core/src/engagement/chat-commands/drizzle/migrations/0000_daffy_lady_bullseye.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE "chat_command_config" (
```
damianrzepka, line 1: This restoration still breaks upgrades from a database that ran the immediately preceding PR head with 0000_striped_trauma. runMigrations compares only SQL hashes, so it treats 0000_daffy…  as new and tries to create chat_command_config again, failing boot with relation already exists. Please add a compatible re-baseline/hash-alias strategy for both histories and cover this upgrade path with a test.

jakubfilinger-b: Fixed in 5292055e.

Added `migrationHashAliases` to `runMigrations`: when a legacy journal hash is recorded, its mapped current hashes are treated as already applied. `chat-commands/migrate.ts` maps the `0000_striped_trauma` hash → `0000_daffy_lady_bullseye` + `0001_sturdy_barracuda` (same final schema: striped created only `chat_command_config`; daffy created it + `chat_gift`, sturdy dropped `chat_gift`).

Coverage:
- `runMigrations` unit test: current hashes aliased from an applied legacy hash are skipped.
- `migration-upgrade.int.test.ts`: seeds a Postgres DB in the striped-trauma post-state, runs `migrate()`, asserts no re-run and `chat_command_config` present / `chat_gift` absent. Fails without the alias (`relation already exists`).

Fresh-DB path is unchanged (no legacy hash → normal apply).

packages/core/src/engagement/chat-commands/AGENTS.md
jakubfilinger-b: Resolved: \`packages/core/src/engagement/chat-commands/AGENTS.md\` was deleted in this PR (the module no longer has gift/rain translation logic to document).

packages/core/src/engagement/chat-commands/AGENTS.md
jakubfilinger-b: Resolved: the whole AGENTS.md file (including the "## Transfer routes" section) was deleted in this PR.

packages/core/src/engagement/chat-commands/AGENTS.md
jakubfilinger-b: Resolved: AGENTS.md was deleted in this PR.

packages/core/src/contracts/adapters/gift-commands.ts
jakubfilinger-b: Resolved: \`packages/core/src/contracts/adapters/gift-commands.ts\` was deleted in this PR along with the port itself.

packages/core/src/contracts/adapters/rain-commands.ts
jakubfilinger-b: Resolved: \`packages/core/src/contracts/adapters/rain-commands.ts\` was deleted in this PR along with the port itself.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
jakubfilinger-b: Resolved: the file was restored rather than deleted. Only the gift-specific funding step changed — the \`creditBonus\` helper now drives \`WALLET_COMMANDS.credit\` directly instead of going through \`claimGift\`. The authz-negative and waterfall describe blocks you flagged are all still there (probes #2 and #3).

packages/core/src/engagement/chat-commands/drizzle/migrations/meta/_journal.json
jakubfilinger-b: Resolved: \`0000_daffy_lady_bullseye\` / \`0001_sturdy_barracuda\` / \`_journal.json\` / snapshots were restored (de6e717f) and \`0000_striped_trauma\` dropped. The upgrade path from a DB that ran the short-lived \`striped_trauma\` head is now handled by a hash-alias strategy in 5292055e (\`migrationHashAliases\` in \`runMigrations\`), covered by a unit test and a Postgres upgrade integration test.

packages/core/src/engagement/chat-commands/plugin.ts
jakubfilinger-b: Resolved: \`dependsOn\` is now \`['chat', 'identity', 'audit']\` (de6e717f).

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
jakubfilinger-b: Resolved: the "real HTTP surface" wording was dropped from the affected \`it\` titles (now "against the real app" / "driving WALLET_COMMANDS directly to fund the credits"), and \`creditBonus\` carries a comment saying it stands in for the downstream HTTP funding path that now lives in the social-transfers extension. Probe #2 (authz negatives) still exercises the real backoffice routes over HTTP.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
jakubfilinger-b: Resolved: added a comment at the \`type: 'gift'\` call (qa-bonus-rollover.e2e.test.ts:67-71) spelling out that \`wallet-commands.service.ts\` only writes a \`wallet_bonus_credit\` row for sourceType \`gift\`/\`rain\`.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
jakubfilinger-b: Resolved: added a comment at the deposit call (qa-bonus-rollover.e2e.test.ts:57-59) — it materialises the wallet row because \`credit()\` fails closed on a missing wallet, and it notes the 1 USD real-balance side effect.

packages/testing/src/__tests__/qa-bonus-rollover.e2e.test.ts
jakubfilinger-b: Resolved: \`expect(credit.sourceType).toBe('gift')\` restored at line 166, and \`sourceType: 'gift'\` restored in the audit-payload \`toMatchObject\` at line 360.

damianrzepka (approved):

PR #110, opened 2026-08-26
feat(notifications): in-app notifications backend

marek-chmielowski-blurify:
## Summary

Adds the backend for in-app notifications: triggers from wallet, chat, and social events, correctly formatted money amounts in notification bodies, and a 30-day retention purge job.

## Why

Consumers need a headless notifications surface that reacts to platform events (wallet, chat, social) and exposes them through the notifications contract/router, without unbounded row growth over time.

## Alternatives considered

None.

## Risks

None. `pnpm verify` passes except `test:integration`, which fails locally due to a pre-existing stale Postgres state (`type "wallet_custody_sweep_status" already exists`) unrelated to this change — this branch does not touch wallet schema or migrations.

### DIFF
--- a/.changeset/notifications-list-pagination.md
+++ b/.changeset/notifications-list-pagination.md
@@ -1,7 +1,0 @@
----
-'@openora/core': minor
----
-
-`GET /notifications` now returns the shared offset-pagination envelope (`{ items, total, page, limit }`, via `paginated(NotificationSchema)`) instead of a bare `Notification[]`, and accepts `page`/`limit`/`sortBy`/`sortOrder` query params.
-
-BREAKING CHANGE: a consumer reading the old bare-array response (e.g. calling `.map()` directly on the payload) must switch to reading `.items` off the new envelope. The `useNotifications` React hook already returns the paginated shape.
--- a/.changeset/notifications-list-pagination.md
+++ b/.changeset/notifications-list-pagination.md
@@ -1,0 +1,7 @@
+---
+'@openora/core': minor
+---
+
+`GET /notifications` now returns the shared offset-pagination envelope (`{ items, total, page, limit }`, via `paginated(NotificationSchema)`) instead of a bare `Notification[]`, and accepts `page`/`limit`/`sortBy`/`sortOrder` query params.
+
+BREAKING CHANGE: a consumer reading the old bare-array response (e.g. calling `.map()` directly on the payload) must switch to reading `.items` off the new envelope. The `useNotifications` React hook already returns the paginated shape.
--- a/packages/core/src/contracts/schemas/events.ts
+++ b/packages/core/src/contracts/schemas/events.ts
@@ -217,15 +217,6 @@ export const domainEventSchemas = {
     currency: CurrencyCodeSchema,
     creditedAmount: MoneyAmountSchema,
   }),
-  // Reserved: no emitter exists yet. A future settlement engine publishes this.
-  'gaming.bet.settled': z.object({
-    roundId: UuidSchema,
-    userId: UuidSchema,
-    playerId: UuidSchema.nullable(),
-    outcome: z.enum(['win', 'loss']),
-    amount: MoneyAmountSchema,
-    currency: CurrencyCodeSchema,
-  }),
 
   'chat.message.sent': z.object({
     messageId: UuidSchema,
--- a/packages/core/src/contracts/schemas/events.ts
+++ b/packages/core/src/contracts/schemas/events.ts
@@ -535,14 +535,6 @@ export const domainEventSchemas = {
     requesterId: UuidSchema,
     addresseeId: UuidSchema,
   }),
-
-  // Reserved: no bonus module exists yet and nothing emits this.
-  'bonus.granted': z.object({
-    bonusId: UuidSchema,
-    userId: UuidSchema,
-    amount: MoneyAmountSchema,
-    currency: CurrencyCodeSchema,
-  }),
 } as const;
 
 export type DomainEventName = keyof typeof domainEventSchemas;
--- a/packages/core/src/contracts/schemas/events.ts
+++ b/packages/core/src/contracts/schemas/events.ts
@@ -217,6 +217,15 @@ export const domainEventSchemas = {
     currency: CurrencyCodeSchema,
     creditedAmount: MoneyAmountSchema,
   }),
+  // Reserved: no emitter exists yet. A future settlement engine publishes this.
+  'gaming.bet.settled': z.object({
+    roundId: UuidSchema,
+    userId: UuidSchema,
+    playerId: UuidSchema.nullable(),
+    outcome: z.enum(['win', 'loss']),
+    amount: MoneyAmountSchema,
+    currency: CurrencyCodeSchema,
+  }),
 
   'chat.message.sent': z.object({
     messageId: UuidSchema,
@@ -525,6 +534,14 @@ export const domainEventSchemas = {
     friendshipId: UuidSchema,
     requesterId: UuidSchema,
     addresseeId: UuidSchema,
+  }),
+
+  // Reserved: no bonus module exists yet and nothing emits this.
+  'bonus.granted': z.object({
+    bonusId: UuidSchema,
+    userId: UuidSchema,
+    amount: MoneyAmountSchema,
+    currency: CurrencyCodeSchema,
   }),
 } as const;
 
--- a/packages/core/src/engagement/chat/service/chat.service.ts
+++ b/packages/core/src/engagement/chat/service/chat.service.ts
@@ -95,7 +95,7 @@ import {
 import { moderateContent, validateAttachment } from '../moderation/index.js';
 const logger = createLogger('chat');
 
-const MENTION_USERNAME_PATTERN = /@([a-zA-Z0-9_]{2,32})/g;
+const MENTION_USERNAME_PATTERN = /(?<![\w.])@([a-zA-Z0-9_]{2,32})/g;
 const MAX_MENTIONS_PER_MESSAGE = 20;
 const MENTION_RESOLVE_CONCURRENCY = 5;
 
--- a/packages/core/src/engagement/chat/service/chat.service.ts
+++ b/packages/core/src/engagement/chat/service/chat.service.ts
@@ -10,6 +10,7 @@ import {
   findOneOrThrow,
   serializeRow,
   pageToOffset,
+  mapConcurrent,
   withAdvisoryXactLock,
   type DrizzleDb,
   type DrizzleTx,
@@ -93,6 +94,57 @@ import {
 } from '../contract/constants.js';
 import { moderateContent, validateAttachment } from '../moderation/index.js';
 const logger = createLogger('chat');
+
+const MENTION_USERNAME_PATTERN = /@([a-zA-Z0-9_]{2,32})/g;
+const MAX_MENTIONS_PER_MESSAGE = 20;
+const MENTION_RESOLVE_CONCURRENCY = 5;
+
+function parseMentionedUsernames(content: string): string[] {
+  const usernames = new Set<string>();
+  for (const match of content.matchAll(MENTION_USERNAME_PATTERN)) {
+    const username = match[1];
+    if (username) {
+      usernames.add(username.toLowerCase());
+    }
+    if (usernames.size >= MAX_MENTIONS_PER_MESSAGE) {
+      break;
+    }
+  }
+  return [...usernames];
+}
+
+function emitMentions(
+  events: EventBus,
+  directory: AdminUserDirectory,
+  content: string,
+  byUserId: User['id'],
+  roomId: ChatRoom['id'] | null,
+  messageId: ChatMessage['id'],
+): void {
+  const usernames = parseMentionedUsernames(content);
+  if (usernames.length === 0) {
+    return;
+  }
+  void mapConcurrent(usernames, MENTION_RESOLVE_CONCURRENCY, (username) =>
+    directory.getPlayerByUsername(username).catch((err: unknown) => {
+      logger.error({ err, username, messageId }, 'chat mention lookup failed');
+      return null;
+    }),
+  )
+    .then((resolved) => {
+      const mentionedIds = new Set(
+        resolved.flatMap((summary) =>
+          summary && summary.userId !== byUserId ? [summary.userId] : [],
+        ),
+      );
+      for (const mentionedUserId of mentionedIds) {
+        events.emit('chat.user.mentioned', { mentionedUserId, byUserId, roomId, messageId });
+      }
+    })
+    .catch((err: unknown) => {
+      logger.error({ err, messageId }, 'chat mention resolution failed');
+    });
+}
 
 function publishChatEvent<T>(transport: RealtimeTransport, roomId: string | null, event: T): void {
   void Promise.resolve()
@@ -1140,6 +1192,7 @@ export class ChatService {
 
     const message = toPublicMessage(record);
     publishChatEvent(this.transport, roomId, message);
+    emitMentions(this.events, this.directory, safeContent, userId, roomId, record.id);
     return message;
   }
 
@@ -1216,6 +1269,7 @@ export class ChatService {
 
     const message = toPublicMessage(record);
     publishChatEvent(this.transport, null, message);
+    emitMentions(this.events, this.directory, safeContent, userId, null, record.id);
     return message;
   }
 
--- a/packages/core/src/engagement/chat/service/chat.service.ts
+++ b/packages/core/src/engagement/chat/service/chat.service.ts
@@ -113,37 +113,47 @@ function parseMentionedUsernames(content: string): string[] {
   return [...usernames];
 }
 
-function emitMentions(
-  events: EventBus,
-  directory: AdminUserDirectory,
-  content: string,
-  byUserId: User['id'],
-  roomId: ChatRoom['id'] | null,
-  messageId: ChatMessage['id'],
-): void {
+function emitMentions({
+  events,
+  directory,
+  isBlockedByMentioned,
+  content,
+  byUserId,
+  roomId,
+  messageId,
+}: {
+  events: EventBus;
+  directory: AdminUserDirectory;
+  isBlockedByMentioned: (mentionedUserId: User['id']) => Promise<boolean>;
+  content: string;
+  byUserId: User['id'];
+  roomId: ChatRoom['id'] | null;
+  messageId: ChatMessage['id'];
+}): void {
   const usernames = parseMentionedUsernames(content);
   if (usernames.length === 0) {
     return;
   }
-  void mapConcurrent(usernames, MENTION_RESOLVE_CONCURRENCY, (username) =>
-    directory.getPlayerByUsername(username).catch((err: unknown) => {
+  void mapConcurrent(usernames, MENTION_RESOLVE_CONCURRENCY, async (username) => {
+    const summary = await directory.getPlayerByUsername(username).catch((err: unknown) => {
       logger.error({ err, username, messageId }, 'chat mention lookup failed');
       return null;
-    }),
-  )
-    .then((resolved) => {
-      const mentionedIds = new Set(
-        resolved.flatMap((summary) =>
-          summary && summary.userId !== byUserId ? [summary.userId] : [],
-        ),
-      );
-      for (const mentionedUserId of mentionedIds) {
-        events.emit('chat.user.mentioned', { mentionedUserId, byUserId, roomId, messageId });
-      }
-    })
-    .catch((err: unknown) => {
-      logger.error({ err, messageId }, 'chat mention resolution failed');
     });
+    if (!summary || summary.userId === byUserId) {
+      return;
+    }
+    const mentionedUserId = summary.userId;
+    const blocked = await isBlockedByMentioned(mentionedUserId).catch((err: unknown) => {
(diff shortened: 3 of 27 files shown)


.changeset/notifications-list-pagination.md
zaxovaiko, line 1: We can remove it.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -93,6 +94,67 @@ import {
 import { moderateContent } from '../moderation/index.js';
 const logger = createLogger('chat');
 
+const MENTION_USERNAME_PATTERN = /@([a-zA-Z0-9_]{2,32})/g;
```
damianrzepka: This regex has no boundary before the `@`, so it matches inside arbitrary text, not just standalone mentions. `hello me at info@acme.com` captures `@acme` and looks it up as a username — if a real player is named `acme`, they get a `chat.mention` notification *and* an email (this map entry has `sendEmail: true`) for text that was never meant to mention them. Worth anchoring the match so it only fires after whitespace/start-of-string, e.g. `(?<![\w.])@([a-zA-Z0-9_]{2,32})`, otherwise this is a real false-positive/spam vector for arbitrary chat content (emails, code snippets, social handles).

packages/core/src/engagement/notifications/service/notifications.service.ts
```diff
@@ -42,13 +43,33 @@ export class NotificationsService {
...
 
-  async listForUser(userId: User['id']) {
-    return this.drizzle.db
-      .select()
+  async listForUser({
+    userId,
+    page,
+    limit,
+    sortOrder,
+  }: PaginationOptions<{ userId: User['id'] }, NotificationSortBy>) {
```
damianrzepka, line 56: `sortBy` isn't destructured here, so it's silently ignored — `listForUser` always orders by `createdAt` no matter what the caller passes for `sortBy` on the `list` route. Not a functional bug today, since `NOTIFICATION_SORT_BY_VALUES` only has one value, but then the query param doesn't do anything either. Either wire it through (`orderBy(dir(notification[sortBy]))`) or drop `sortBy` from the input schema until there's a second sort key — as-is it's dead API surface.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -210,6 +210,15 @@ export const domainEventSchemas = {
     userId: UuidSchema,
     playerId: UuidSchema.nullable(),
   }),
+  // Reserved: no emitter exists yet. A future settlement engine publishes this.
```
damianrzepka: Question on scope: `gaming.bet.settled` and `bonus.granted` have no emitter anywhere in the codebase — no gaming or bonus module exists yet — but this PR adds their schemas here, plus a full `notificationEventMap` entry each, plus dedicated `it.each` coverage in `notification-event-map.test.ts`. That's a fair amount of surface (schema + mapping + tests) for events nothing can fire today. Is there a concrete near-term module landing that needs this, or could these two entries wait until the emitting module actually exists? Building ahead of the emitter is the kind of speculative-future-need the repo's own conventions call out to avoid.

damianrzepka (commented):
Solid feature overall — pagination envelope + breaking-change changeset, realtime stream, retention purge, and the mention detection all look well-tested. Left three comments: one real correctness/product concern (the mention regex catching arbitrary @-text and emailing whoever it happens to match), and two smaller nits (an unused `sortBy` param, and a scope question about the two dormant event mappings with no emitter yet). Nothing here should block on its own, but please take a look at the mention-regex one before merge.

packages/core/src/engagement/notifications/schema/index.ts
```diff
@@ -8,6 +8,7 @@ export const notification = pgTable(
     type: text().notNull(),
     title: text().notNull(),
     body: text().notNull(),
```
zaxovaiko, line 10: Title and body should be language-agnostic. Let's remove them from here and just keep the `type` as enum value, so that later on frontend we could use this enum for mapping type to exact title/description

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -3,30 +3,283 @@ import {
...
+  mapEvent(
+    'chat.user.mentioned',
+    (p) => ({
+      userId: p.mentionedUserId,
+      type: 'chat.mention',
+      title: 'You were mentioned',
+      body: 'You were mentioned in a chat message.',
+      data: compactData({ roomId: p.roomId, messageId: p.messageId }),
+    }),
+    { sendEmail: true },
```
zaxovaiko: Hm, `sendEmail: true` on `chat.mention` worries me - any player can drop 20 `@name` tokens into one global chat message and that is 20 emails to recipients they picked. The chat send limiter is `10 / 10_000ms` with `onUnavailable: 'allow'`, so it fails open on top of that. Let's keep mentions in-app only, the realtime path already covers it, and revisit email once we have a notification preference model.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -3,30 +3,283 @@ import {
...
   userId: UuidSchema,
   reason: z.string().nullable(),
 });
 
+// p.amount on every wallet/chat/gaming event payload is a MoneyAmount decimal string
+// scaled to MONEY_SCALE = 18 (crypto-first precision), eg "100.000000000000000000" -
+// display-only trimming for notification text; Number() can lose precision below its
+// ~15-17 significant digits for an extreme 18-decimal value, an accepted tradeoff.
+function formatMoneyAmount(amount: string): string {
+  const value = Number(amount);
```
zaxovaiko: `Number(amount)` on an 18-scale decimal bites us here - `0.000000001000000000` formats to `0`, so a small withdrawal emails "Your withdrawal of 0 ETH has been approved". Let's format the decimal string directly (trim zeros, group the integer part) instead of going through a float.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -3,30 +3,283 @@ import {
...
+    { sendEmail: true },
+  ),
+
+  mapEvent(
+    'wallet.bonus_rollover.completed',
+    (p) => ({
+      userId: p.userId,
+      type: 'wallet.bonus_rollover.completed',
+      title: 'Bonus unlocked',
+      body: `Your ${p.creditedAmount} ${p.currency} bonus credit has cleared its rollover requirement and is now fully withdrawable.`,
```
zaxovaiko: This one skips `formatMoneyAmount` - player sees `100.000000000000000000 USD`. Let's route it through the same formatter as the rest.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -3,30 +3,283 @@ import {
...
+  createNotificationsRouter,
+  notificationsChannel,
+  toNotificationDto,
+} from './router/index.js';
 import { NotificationsService } from './service/notifications.service.js';
+import type { CreateNotificationInput } from './contract/index.js';
 
 const KYC_RESUBMISSION_NOTIFY_QUEUE = queue('kyc-resubmission-notify');
+const NOTIFICATIONS_RETENTION_PURGE_QUEUE = queue('notifications-retention-purge');
+const NOTIFICATION_RETENTION_DAYS = 30;
```
zaxovaiko: Shouldn't retention be operator config rather than a module constant? We hard-delete withdrawal, deposit, balance-adjustment and KYC notifications at 30 days, and that obligation is usually measured in years for a licensed operator. `wallet/plugin.ts` already reads its cron from `platformConfig` - let's do the same here with these as defaults. Same applies to the cron below.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -59,77 +313,38 @@ export default {
...
-      const p = parsed.data;
-      const title = 'Friend request accepted';
-      const body = `${p.accepterUsername} accepted your friend request.`;
-      svcRef
-        .create({ userId: p.requesterId, type: 'social.friend_request.accepted', title, body })
-        .catch((err) =>
-          logger.error({ err }, 'social.friend_request.accepted notification failed'),
-        );
-    });
+    for (const entry of notificationEventMap) {
```
zaxovaiko, line 302: Handlers are at-least-once - `redis-streams-broker.ts` reclaims pending entries and says core has no dedup layer - so a reclaim sends a second "Withdrawal approved" email. Can we carry the envelope `eventId` into `create()`, add a unique column on it and `onConflictDoNothing().returning()`, then publish and email only when a row actually comes back?

packages/core/src/engagement/notifications/contract/index.ts
```diff
@@ -1,46 +1,91 @@
...
+export const MarkNotificationReadOutputSchema = z.object({ success: z.literal(true) });
+export type MarkNotificationReadOutput = z.infer<typeof MarkNotificationReadOutputSchema>;
+
 export const notificationsContract = {
-  list: oc.route({ method: 'GET', path: '/notifications' }).output(z.array(NotificationSchema)),
+  list: oc
+    .route({ method: 'GET', path: '/notifications' })
+    .input(
+      PageQuerySchema.extend({
+        sortBy: NotificationSortBySchema.optional(),
```
zaxovaiko, line 67: `sortBy` never reaches `listForUser` - it destructures `userId, page, limit, sortOrder` only. Harmless while the enum has one value, but the moment we add a second sort key it will keep validating and keep being dropped. Let's remove it until we actually need it.

packages/core/src/engagement/notifications/router/index.ts
```diff
@@ -1,33 +1,81 @@
...
-          return [
-            {
-              ...n,
-              type: type.data,
-              readAt: n.readAt ? n.readAt.toISOString() : null,
-              createdAt: n.createdAt.toISOString(),
-            },
-          ];
+    list: os.list.handler(({ input, context }) =>
+      notifications.listForUser({ userId: getUserId(context), ...input }).then((page) => ({
```
zaxovaiko: Let's flip this to `{ ...input, userId: getUserId(context) }` - zod strips an injected `userId` today, but the ownership boundary should not depend on the schema staying strict.

packages/core/src/engagement/notifications/router/index.ts
```diff
@@ -1,33 +1,81 @@
...
+  if (!type.success) {
+    return null;
+  }
+  return {
+    id: row.id,
+    userId: row.userId,
+    type: type.data,
+    title: row.title,
+    body: row.body,
+    data: row.data ?? null,
```
zaxovaiko: `type` right above is parsed defensively but `data` goes out raw, and the column has no check constraint - so anything an overlay writes rides straight through the realtime publish too. Can we `safeParse` it the same way?

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -59,77 +313,38 @@ export default {
...
+        if (!input || !svcRef) {
+          return;
+        }
+        const svc = svcRef;
+        svc
+          .create(input)
+          .then((record) => {
+            publishNotification(record);
+            if (entry.sendEmail) {
+              sendEmail(input.userId, input.title, input.body);
```
zaxovaiko: This is where I think we are not production-ready yet. `sendEmail` is fire-and-forget with a `.catch(log)`, so a transient SMTP or provider blip just drops the email and nobody ever finds out - and `create` above it is the same, an insert failure is logged and gone. Everything else durable in the repo goes through `JOB_QUEUE`, which already has `attempts`, `backoff` and a dead-letter hook. Can we push create-plus-notify onto a queue with a few attempts and exponential backoff, and wire `onDeadLetter` so an exhausted withdrawal email surfaces somewhere?

packages/core/src/engagement/notifications/service/notifications.service.ts
```diff
@@ -76,4 +97,15 @@ export class NotificationsService {
...
   }
+
+  // Unconditional on age (read or unread), not gated on readAt - deletes anything past
+  // the retention window regardless of read state.
+  async purgeExpired(retentionDays: number) {
+    const cutoff = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
+    const rows = await this.drizzle.db
+      .delete(notification)
+      .where(lt(notification.createdAt, cutoff))
+      .returning({ id: notification.id });
```
zaxovaiko: `.returning({ id })` just to count - the first run against a backlog pulls every deleted id into memory. I know `markAllRead` does the same above, but there it is bounded per user and here it is not. Maybe worth a raw delete and reading `rowCount`?

zaxovaiko (commented):
Nice refactor, the declarative map reads much better than five hand-rolled `events.on` blocks, and the SSE and channel wiring matches what `wallet` and `compliance` already do.

The main thing before this is production-ready is that there is no retry anywhere on the notify path. The 14 mapped events dispatch inline in the event handler, so a failed insert or a failed email is logged and gone. The KYC path does go through a job, but `enqueue` passes no `attempts` or `backoff` and BullMQ defaults to one attempt, so that one does not retry either - and nothing wires `onDeadLetter`, which the port already gives us. Commented inline on where I would put the queue. Worth pairing with the idempotency point, since a retry without a dedup key just sends the email twice.

The one I would like fixed before merge is `sendEmail: true` on `chat.mention`, also commented inline. The rest are smaller.

Two more I could not anchor to the diff. `kyc-resubmission-notify:${envelope?.eventId ?? p.userId}` - on the `?? p.userId` branch the key is stable per player and BullMQ keeps the `jobId` for 24h, so a genuine second resubmission request inside that window is dropped silently and the player never hears about it. Let's key on `eventId` only and log-and-skip when it is missing. And `mock/mock-notification-adapter.ts` writes the player's real email to stdout as the default bind - pre-existing, but mentions and tips push a lot more volume through it now, so maybe worth redacting `to`.

One on the client side too: `useEventStream` reconnects with backoff, which is good, but there is no backfill on reconnect and nothing invalidates `unreadCount`, so the badge can drift after a dropped connection. The list refetch covers the list itself.

Also the mention tests do not cover the blocked-recipient branch or the fail-closed rejection - those are the two that matter most here.

zaxovaiko:
One more, and this one actually changes the shape of the PR.

We have an agreed direction for mail where the platform never composes the body - it passes a template key plus data, and a mail module resolves the recipient's language and renders. `sendEmail: true` is a boolean, and a flag cannot carry the key or the data, so every entry here hand-composes an English string instead. That is the exact thing the direction is meant to remove, and this PR takes that path from 2 emails to 10.

Could we land this in-app only for now, and add the mail channel as `{ templateKey, data }` once the mail module is in? That drops the `chat.mention` email concern above for free. Worth flagging too that the emails actually agreed are a smaller set - withdrawal approved and rejected, KYC resubmission, admin invite. Deposit, tip, mention, bet, bonus and the extra withdrawal states are not settled yet, and "does a player get several emails for one withdrawal" is a product question we have not answered.

On the plus side, the KYC path in this same file - event to job, idempotency key from the event id - is the pattern we want everywhere. Let's extend that one rather than grow a second inline path next to it.

.changeset/notifications-list-pagination.md
marek-chmielowski-blurify: Removed the changeset in d0e130b7.

packages/core/src/engagement/chat/service/chat.service.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — anchored the pattern with a negative lookbehind so it only matches after whitespace/start-of-string/a non-word-non-dot boundary: `/(?<![\w.])@([a-zA-Z0-9_]{2,32})/g`. Also added regression coverage for the `info@acme.com` case, plus the blocked-recipient and fail-closed-on-error branches that were missing.

packages/core/src/engagement/notifications/service/notifications.service.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — `listForUser` now destructures `sortBy` and orders by `notification[sortBy]`.

packages/core/src/contracts/schemas/events.ts
marek-chmielowski-blurify: Agreed, removed both in d0e130b7 — pulled `gaming.bet.settled` and `bonus.granted` out of the schema, the notification map, and the type enum. They can come back once a module actually emits them.

packages/core/src/engagement/notifications/schema/index.ts
marek-chmielowski-blurify: Not doing this in this PR — it ties into the templated-mail/i18n direction from your top-level comment, which we're deferring to a separate task. `title`/`body` stay as-is for now; will revisit alongside that work.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — flipped `sendEmail` to `false` for `chat.mention` and every other newly-added trigger type. Only kept email on `withdrawal.approved`/`rejected` and KYC resubmission, matching what's already on `dev`. Full scope decision is in the top-level comment thread.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — rewrote `formatMoneyAmount` to operate on the decimal string directly (trim trailing zeros, group the integer part) instead of round-tripping through `Number()`. Added a regression test for the tiny-18-decimal case across every mapped event.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — routed `creditedAmount` through `formatMoneyAmount` too.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — retention days and the purge cron now come from `platformConfig.notifications.retention`, same pattern as `wallet.sweep`, falling back to the old defaults (30 days) when unset.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — added a nullable `eventId` column with a unique index on `notification`, and `create()` now does `onConflictDoNothing({ target: notification.eventId })`. A reclaimed/retried delivery is a DB-level no-op instead of a duplicate row/email.

packages/core/src/engagement/notifications/contract/index.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — same fix as the `listForUser` thread, it's wired through now.

packages/core/src/engagement/notifications/router/index.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — flipped to `{ ...input, userId: getUserId(context) }`.

packages/core/src/engagement/notifications/router/index.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — `toNotificationDto` now `safeParse`s `data` through `NotificationDataSchema`, same failure mode as `type`.

packages/core/src/engagement/notifications/plugin.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — event handlers now enqueue onto a new `notifications-dispatch` queue (JOB_QUEUE, 5 attempts, exponential backoff, `onDeadLetter` logging) instead of dispatching inline. Paired with the eventId dedup from the other thread so a retry doesn't double-create or double-email.

packages/core/src/engagement/notifications/service/notifications.service.ts
marek-chmielowski-blurify: Fixed in d0e130b7 — switched to a raw delete and read `rowCount` instead of `.returning()`.

marek-chmielowski-blurify:
Pushed d0e130b7 addressing review feedback (62fc09a6..d0e130b7).

On the mail/i18n direction from the top-level comment: going with your suggestion — landing in-app-only for now. Email stays exactly where it already was on `dev` (withdrawal approved/rejected, KYC resubmission); everything else this PR added (deposit, tip, mention, the extra withdrawal states, the bet/bonus placeholders) is in-app only. The `{templateKey, data}` mail model is a separate task, as you suggested — that also drops the `chat.mention` email concern for free.

A few other things from the review that weren't anchored to a line, also in d0e130b7:
- KYC resubmission idempotency key: dropped the `?? p.userId` fallback, keys purely on `envelope.eventId` now (logs + skips enqueue if it's somehow missing).
- Added the missing mention test coverage: blocked-recipient and fail-closed-on-error branches.

Not addressed yet, flagging explicitly rather than going quiet on them:
- `mock-notification-adapter` still logs the raw email to stdout unredacted — still true, though lower volume now that most of this PR's new types don't email at all.
- `useEventStream` reconnect backfill / `unreadCount` invalidation on reconnect — that's the React SDK, leaving it out of this backend PR.

Also dropped the two dormant event mappings (`gaming.bet.settled`, `bonus.granted`) per the scope question on that thread — no emitter exists yet, they can come back with whichever module actually publishes them.

packages/core/src/engagement/notifications/schema/index.ts
zaxovaiko: Just don't forget about it 👌

PR #111, opened 2026-08-26
fix(identity): forward better-auth error codes to the client

jakubfilinger-b:
better-auth builds every error body as { message, code } (APIError.from), but ensureOk read only the message and dropped the code. A login rejected for an unverified email was therefore indistinguishable from a wrong password: both arrived as a bare FORBIDDEN/UNAUTHORIZED whose only difference was display text, which is free to change between releases. A consumer was left choosing between one catch-all message and matching on English strings.

The code now travels as data.code, so every better-auth-backed identity route - login, register, password, 2FA - hands the client something typed to branch on.

The anti-enumeration path (genericMessage) still sends no code, and an error body without one produces no data at all, so no error that clients read today changes shape.

### DIFF
--- a/packages/core/src/pam/identity/service/identity.service.ts
+++ b/packages/core/src/pam/identity/service/identity.service.ts
@@ -14,6 +14,7 @@ import {
 } from '@openora/core/server';
 import { parseCookies } from 'better-auth/cookies';
 import { and, eq, isNull, sql } from 'drizzle-orm';
+import * as z from 'zod';
 import { user, session, account, verification, twoFactor } from '../schema/index.js';
 import { player } from '@openora/core/pam/schema/profile';
 import type {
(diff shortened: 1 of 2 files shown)


.changeset/forward-better-auth-error-codes.md
zaxovaiko, line 1: We don't need changesets for now

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -170,15 +169,22 @@ async function ensureOk(res: globalThis.Response, opts?: { genericMessage?: stri
     });
   }
   let message = `Request failed (${res.status})`;
+  let betterAuthCode: string | undefined;
   try {
-    const parsed = JSON.parse(await res.text()) as { message?: string };
+    const parsed = JSON.parse(await res.text()) as { message?: string; code?: string };
```
zaxovaiko: We can even use a Zod schema for double verification. I don't trust `as { ... }`

zaxovaiko (dismissed):

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: good point, zod is added.

zaxovaiko (approved):

PR #115, opened 2026-08-27
feat(compliance): add tiered kyc verification

damianrzepka:
## Summary

Adds tiered KYC verification (BF-504): `basic` and `advanced` are tracked as independent `kyc_verification` histories, each with its own current status. `player.kycStatus` (and therefore the withdrawal gate) still tracks basic tier only; advanced tier is a separate signal for operators to build on.

## Why

Operators need a second, higher verification level (advanced) that doesn't gate withdrawals but tracks its own lifecycle independently of basic KYC, without forcing every consumer through a single flat verification record.

## Alternatives considered

`tier` was also added to `requestKycResubmission`/`overrideKycStatus`/`bulkApproveKyc` for schema consistency with the tiered model, even though backoffice KYC review was scoped out of the ticket - flagging that decision here since it wasn't recorded on the ticket. The ticket's open question ("does the vendor expose separate Basic/Advanced workflows?") is answered in code: `submit()` takes a `tier`, and a single vendor workflow can serve both tiers (see `docs/adapters/kyc.md`) - an adapter that can attribute a decision to one tier sets the new optional `KycResult.tier`, and `reconcile()` then scopes to that tier; otherwise it fans the decision out across the shared reference. Recording that decision on the ticket itself is left to the ticket owner.

## Risks

**Migration/rolling-deploy.** Migration 0006 drops the old `reference_id`-only unique index in the same migration that adds the new composite one. A pod still running the pre-tier build's `onConflictDoUpdate({ target: referenceId })` will fail its submits once 0006 has run and before that pod is replaced. Mitigation: sequence the deploy so this migration runs only after every pod is on the new build, never mid-rollout. (A "keep both indexes for one release" split was tried and reverted - it's incompatible with this PR's shared-vendor-workflow feature, since two tiers can legitimately share one `referenceId`.)

**Breaking contract change.** `PlayerKycView` reshaped, `submitKyc.tier` now required, `KycAdapter.submit`/`getStatus` gained a required `tier` param, `KycStatusWriter.setStatus`'s return type changed. Covered by the attached changeset (major).

See inline review responses below for the rest.


packages/core/src/compliance/service/kyc.service.ts
```diff
@@ -155,14 +189,20 @@ export class KycVerificationService {
...
+        .where(eq(kycVerification.referenceId, result.referenceId))
+        .for('update');
       const inserted = findOneOrThrow(
         await trx
           .insert(kycVerification)
           .values({
             userId,
             provider: this.provider,
             referenceId: result.referenceId,
+            tier: input.tier,
```
dkrzysztof: Shouldn't we make the unique/conflict key tier-aware and scoped to the user? With `referenceId` globally unique, the same vendor workflow cannot be stored for both Basic and Advanced: the second submission updates the existing row, keeps its original tier, and can overwrite Basic with an Advanced decision.

damianrzepka: Good catch - this was valid. I changed the unique/conflict key to (userId, referenceId, tier) and reconciliation now applies a shared vendor workflow decision to every matching tier, so Basic and Advanced remain independent while one provider session can complete both. Added regression coverage. Fixed in b8afe07c.

dkrzysztof (dismissed):
lgtm

dkrzysztof (dismissed):

packages/core/src/compliance/drizzle/migrations/0006_military_sally_floyd.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX "kyc_verification_reference_id_key";--> statement-breakpoint
```
jakubfilinger-b: this drops the unique index that the currently-deployed `submit()` still relies on for its `onConflictDoUpdate({ target: referenceId })`. during a rolling deploy the old pods will throw `no unique or exclusion constraint matching the ON CONFLICT specification` on every kyc submit until they get replaced. can we add the new `(user_id, reference_id, tier)` unique index now and drop the old one in a follow-up release instead?

packages/core/src/compliance/router/index.ts
```diff
@@ -113,6 +113,8 @@ export function createComplianceRouter({
       return kyc.getForPlayer(input.userId);
     }),
 
+    getMyKyc: os.getMyKyc.handler(({ context }) => kyc.getForPlayer(getUserId(context))),
```
jakubfilinger-b: `getMyKyc` returns the full `PlayerKycViewSchema`, so the player gets `riskSignals` (vpn/tor, datacenter ip, duplicate device, high-risk country, device fingerprints), raw `checks`, `decisionReason`, `provider` and `referenceId`. once a webhook fills those in, a player can poll `/compliance/kyc/me` and read our fraud-detection internals. can we give the player-facing route a trimmed schema (status, tier, documentTypes, timestamps) and keep the rest on the admin route?

packages/core/src/compliance/service/kyc.service.ts
```diff
@@ -143,10 +153,42 @@ export class KycVerificationService {
...
+    reason: string | null;
+    source: 'vendor' | 'manual' | 'webhook' | 'reverify';
+  }) {
+    if (params.tier === 'basic' && !params.playerTransition) {
+      return;
+    }
+    if (params.tier === 'advanced' && params.previousStatus === params.status) {
+      return;
+    }
+    this.events.emit('compliance.kyc.updated', {
```
jakubfilinger-b, line 219: this emit now runs after the transaction commits, best-effort. `docs/standards/compliance.md` says the status change, history row and audit event should be in one transaction, and previously the emit happened inside the writer's tx. a crash between commit and this line silently loses the `compliance.kyc.updated` audit event (5AMLD trail). either emit through the outbox with `emitInTransaction`, or update the standard + add an ADR for the change - the contract doc in kyc.ts was already reworded without one.

packages/core/src/compliance/plugin.ts
```diff
@@ -95,6 +95,7 @@ export default {
       return realtimeTransport.publish(kycStatusChannel(parsed.data.userId), {
         eventId: envelope.eventId,
         status: parsed.data.status,
+        tier: parsed.data.tier,
```
jakubfilinger-b, line 104: advanced-tier updates now go out on `kycStatusChannel(userId)` too, which is the stream the withdrawal-status UI listens on. same in notifications/plugin.ts - an advanced resubmission request sends the player "resubmit your verification documents" with no tier context. tag-evaluation filters on `tier === "basic"`, these two don't. intended?

jakubfilinger-b (changes requested):
had a look. main things before merge:

- migration 0006 drops a unique index the previous release still needs (rolling-deploy breakage) - inline
- new `/compliance/kyc/me` exposes risk signals / checks / decisionReason to the player - inline
- `compliance.kyc.updated` emit moved outside the tx, best-effort - conflicts with compliance.md, inline
- advanced-tier events leak onto the basic realtime channel + resubmission notifications - inline
- no changeset for what is a breaking contract change (`PlayerKycView` shape, new required `tier` inputs), and the PR description is still the empty template

nothing structurally wrong with the tiering approach itself. holding off until the migration ordering and the risk-signal exposure are sorted.

marek-chmielowski-blurify (dismissed):

packages/core/src/compliance/drizzle/migrations/0006_military_sally_floyd.sql
damianrzepka: Tried exactly that split - add the new composite index now, drop the old one later - and reverted it: a plain `UNIQUE(reference_id)` can't coexist with this PR's own shared-vendor-workflow feature (one `referenceId` legitimately spanning both a Basic and an Advanced row, see the `keeps a shared vendor workflow separate for Basic and Advanced` test). Keeping it isn't a one-release stopgap, it permanently breaks that feature. The actual mitigation is deploy sequencing: this migration needs to run only after every pod is on the new build, not interleaved with a rolling restart. Documented in the PR description and the changeset.

packages/core/src/compliance/router/index.ts
damianrzepka: Fixed - added a trimmed `PlayerKycSummaryViewSchema` (`tier`, `status`, `documentTypes`, `submittedAt`, `decidedAt`, `createdAt`, `updatedAt` only) for `getMyKyc`; `riskSignals`/`checks`/`decisionReason`/`provider`/`referenceId` stay admin-only on `getPlayerKyc`. Added e2e coverage asserting both the trimmed shape (with a webhook actually populating those fields first, to prove they're stripped, not just absent) and that it scopes to the session's own user.

packages/core/src/compliance/service/kyc.service.ts
damianrzepka: Fixed - moved the `compliance.kyc.updated` emit back inside `db.transaction(...)` in all 5 call sites (`submit`, `reconcile`, `handleDeposit`, `requestResubmission`, `overrideStatus`), restoring the pre-PR guarantee. Went with this over `emitInTransaction` because no module in the codebase currently uses the outbox and requiring it here would break every non-outbox deployment for a core compliance flow; restoring the in-tx placement undoes only this PR's regression without a new infra dependency, so no ADR needed either.

packages/core/src/compliance/plugin.ts
damianrzepka: Not intended - added the same `tier !== 'basic'` guard `tag-evaluation.service.ts` already used, to both the realtime publish (`compliance/plugin.ts`) and the resubmission notification (`notifications/plugin.ts`), so both now match tag-evaluation's precedent.

damianrzepka:
Pushed a follow-up commit (0673d90) addressing all four inline threads above (see individual replies), plus the changeset and PR description jakubfilinger-b flagged in the review summary.

A colleague ran a second, independent review pass and surfaced additional issues. Verified each against the code before touching anything; confirmed and fixed in this same commit:

**`reconcile()` fan-out.** It applied one vendor decision to every `kyc_verification` row sharing a `referenceId`, with no `tier`/`userId` scoping - could write one tier's `checks`/`documentTypes` onto a row that never ran them, and (combined with dropping the old unique index) had no DB-level guarantee a `referenceId` belongs to one player. Added an optional `KycResult.tier` (threaded through `parseWebhook` -> the job payload -> `syncDecision` -> `reconcile`) so a decision scopes to one tier when the adapter can attribute it, keeping today's shared-workflow fan-out when it can't. `reconcile` now fails closed - throws, updates nothing - on a cross-player reference collision or a webhook/`resolveDecision` tier disagreement.

**Sticky `advanced_kyc_needed` tag.** `tag-evaluation.service.ts` bailed on `tier !== 'basic'` before ever reaching the tag-clearing branch, so an Advanced submission or approval could never clear its own needed-tag, while an unrelated Basic approval cleared it as a side effect. Each tier now clears only its own tag.

**Event forward-compat.** `tier` was required on 4 event schemas; since every deployment now binds `RedisStreamsBroker` (a durable backlog that survives a restart), an old-shaped event still queued at rollout would be silently dropped by `safeParse`. Defaulted to `'basic'`, matching `audit/plugin.ts`'s already-forward-compatible pattern.

**`previousStatus` audit-log bug.** A fresh Advanced resubmission (new `referenceId`) after a rejection defaulted `previousStatus` to `'not_started'` instead of the real prior status, corrupting the audit trail. `submit()` now reads it from `latestVerification(userId, tier)`.

**Duplicate `high_risk_signal_detected`.** Fired once per changed row, so a shared-reference webhook touching both tiers emitted the same real signal twice. Now emitted once per `reconcile()` call.

Also dropped a redundant pre-transaction select in `reconcile()`, fixed it returning the newest row instead of the one that actually changed, added `getMyKyc` test coverage (trimmed shape + user scoping), extended the changeset to cover the `KycAdapter`/`KycStatusWriter` port signature changes, and corrected a doc/contract wording contradiction in `docs/adapters/kyc.md` (`tier` is caller-chosen, not platform-owned).

Not adopted: a deprecation shim for the `{current,history}` -> `{basic,advanced}` reshape - this repo's convention is a changeset + major bump for a breaking change, which is what's attached, not a compat shim. Recording the ticket's "does the vendor expose separate workflows" answer on the ticket itself is left to me/the ticket owner rather than done here.

.changeset/kyc-tiered-verification.md
zaxovaiko, line 1: Let's remove that for now

docs/adapters/kyc.md
zaxovaiko, line 1: Let's remove the code from `kyc.md`. IMO, docs should only describe the business logic. They shouldn't duplicate the implementation (we'd have to update it in two places). So, please remove the code here

packages/core/src/compliance/drizzle/migrations/0005_marvelous_kulan_gath.sql
zaxovaiko, line 1: Let's squash migrations (combine two into a single one)

jakubfilinger-b (dismissed):

.changeset/kyc-tiered-verification.md
damianrzepka: Done - dropped the changeset in 0ef0f990, same as the identity/wallet precedent of removing a changeset on review request.

docs/adapters/kyc.md
damianrzepka: Done in 0ef0f990 - reworded the paragraph to describe the tier behavior without repeating the submit()/getStatus() signatures. Left the recipe class bodies (Sumsub/hosted adapter examples) as-is since those are the actual "how to implement" code the doc exists to show, not a duplicate of the contract.

packages/core/src/compliance/drizzle/migrations/0005_marvelous_kulan_gath.sql
damianrzepka: Done in 0ef0f990 - deleted 0005/0006 plus their meta snapshots, dropped the 0006 journal entry, and reran `drizzle-kit generate` off the current schema. That produced one migration (0005_marvelous_kulan_gath) with the same statements, just combined.

jakubfilinger-b (approved):

PR #116, opened 2026-08-28
feat(chat): assign and revoke the moderator role in a private room

mp-blurify:
## Summary

Let a private-room owner grant and revoke the chat `moderator` role (BF-493), and close the
private-room lifecycle and realtime gaps that route exposed: a named signal lane the default
transport actually carries, user-keyed channel revocation, and a serialised private-room delete.

## Why

`chat_room_member.role` and the whole moderation authorization matrix already existed, but
nothing could write the column: no route, no service method, and no DI seam an overlay could
decorate. The moderator tier was unreachable, and the promote/demote UI had nothing to call.

Making it reachable surfaced the rest. A promoted member kept rendering as a plain one until
their roster cache expired, because the `EventBus` is server-side only. `deletePrivateRoom`
validated outside its advisory lock and fanned out to the realtime transport before emitting
its audit event, so two concurrent deletes both "succeeded" and one unreachable client could
fail a deletion that had already committed. And channel revocation was keyed on the
per-connection `clientId` a caller picks itself, while every caller passed a user id.

## Alternatives considered

An overlay writing `chatRoomMember.role` through the exported `engagement/schema/chat` subpath.
Rejected: it puts a core-owned table under two writers and forces a second overlay table for
the assignment timestamp ownership transfer needs.

Reusing the existing `assertModerator` guard for the new route. Rejected: it admits moderators,
and a moderator creating peers is the one thing this capability must not allow. Granting is
owner-only.

Carrying signals on the channel's existing payload lane. Rejected: that lane is typed
`ChatMessage` end to end, so a control signal there corrupts the message stream. Signals get
their own lane and their own SSE route instead.

## Risks

`RealtimeTransport.revokeClientFromChannel(clientId, channel)` is now
`revokeUserFromChannel(userId, channel)`. Nothing in core is affected, but a third-party
adapter implementing the port must rename the method and revoke every connection issued under
that user. There is no changeset carrying this note - removed on review request.

Migration `0010` backfills `role_assigned_at` from `joined_at` for rows that already hold a
role above `member`. It is a one-shot data write over `chat_room_member`, idempotent under
its `IS NULL` guard.

A private-room owner can no longer leave their own room: they must transfer ownership or
delete it. Promoting a moderator used to raise the owner+moderator count past 1 and let the
owner out, stranding a room whose `creatorId` still pointed at them with nobody able to manage
it. Ownership transfer does not exist yet, so an owner's only exit today is deleting the room.


packages/core/src/audit/plugin.ts
```diff
@@ -359,6 +359,19 @@ export async function mapEventToRecord(
     };
   }
 
+  // actorId = the granting/revoking owner's resolved playerId; resource = the member whose
+  // role moved. after carries the new role, so the trail reads as a role history per member.
+  if (topic === 'chat.room.member.role-changed') {
+    return {
+      ...base,
+      actorType: 'player',
```
jakubfilinger-b: Could we preserve both the real actor and the previous role here? For an admin-owned room, `getPlayerIdByUserIdSafe()` returns `null`, while `changedBy` still contains the acting user ID. This mapper currently records `actorType: 'player'`, `actorId: null`, and discards `changedBy`. It also records only the new role, although permission changes require meaningful before/after state. Please include `previousRole` in the event and fall back to `changedBy` with the appropriate actor type when `playerId` is unavailable.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -1633,7 +1666,23 @@ export class ChatService {
...
+          .select({ userId: chatRoomMember.userId })
+          .from(chatRoomMember)
+          .where(eq(chatRoomMember.roomId, roomId));
+        await t.update(chatRoom).set({ deletedAt }).where(eq(chatRoom.id, roomId));
+        return rows;
+      }),
+    );
+    // Only once the delete is committed: the room is gone, so cut every member off the
+    // room channel rather than leaving them subscribed to a room that 404s on every call.
+    await mapConcurrent(members, ROOM_REVOKE_CONCURRENCY, async ({ userId: memberId }) => {
```
jakubfilinger-b: The room has already been soft-deleted when this realtime fan-out starts. If any `revokeClientFromChannel` call rejects, `mapConcurrent` throws before `chat.private_room.deleted` is emitted, so the endpoint reports failure for a committed deletion and no audit event is produced. A retry then returns `NOT_FOUND` and cannot repair the missing cleanup or audit. Please emit the domain event immediately after commit and isolate/log individual realtime failures, or enqueue an idempotent cleanup job. A rejecting-transport regression test would cover this path.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -239,4 +241,93 @@ export class ChatRoomMembershipService {
...
+        // Actor's resolved playerId, matching the kicked/banned events the audit trail
+        // already attributes this way.
+        playerId: await this.identityReader.getPlayerIdByUserIdSafe(actorId),
+        ip: ip ?? null,
+        userAgent: userAgent ?? null,
+      });
+      // The EventBus is server-side only, so without this the promoted member keeps
+      // rendering as a plain one until their roster cache happens to expire.
+      const signalPayload: ChatMemberRoleChangedSignal = { roomId, userId, role };
+      await this.transport?.signal?.(
```
jakubfilinger-b: A rejected `signal()` happens after the role update has committed, but it still makes the route return an error. Retrying does not repair the notification because the repeated role assignment becomes a no-op at lines 300–301, leaving connected clients stale. Please treat realtime signaling as best-effort with contextual logging, or persist an idempotent retry separately from the mutation, and add a test with a rejecting signal transport.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -1633,7 +1666,23 @@ export class ChatService {
...
+      await this.transport.revokeClientFromChannel?.(memberId, chatChannel(roomId));
```
jakubfilinger-b: This passes a user ID to a port that revokes one `clientId`. `getConnection` allows clients to request a custom per-connection ID, so connections such as a second tab remain authorized after the room is deleted. The current test hides this because it subscribes with `clientId === userId`. Could the port revoke by authenticated user and channel, or could all issued client IDs be enumerated and revoked? Please cover a custom `clientId` in the test.

packages/core/src/contracts/adapters/realtime.ts
```diff
@@ -54,6 +54,20 @@ export type RealtimeTransport = {
    * invoke on teardown (eg an SSE handler on request abort).
    */
   subscribe<T>(channel: string, handler: (event: T) => void, clientId?: string): () => void;
+  /**
+   * Push a NAMED control signal to a channel - "something about this room changed, refetch"
+   * - as opposed to `publish`/`remove`, which carry the channel's payload stream itself (for
+   * chat, `ChatMessage`). A managed transport delivers it as its own named event, the way
+   * `revokeClientFromChannel` already surfaces `chat:access-revoked`, so a signal never
+   * reaches a payload subscriber and cannot corrupt that stream. Optional capability per
```
jakubfilinger-b: The stale-roster fix is not available through OpenORA’s shipped default realtime path: `signal` is optional, `InProcessRealtimeTransport` does not implement it, and the SSE stream carries only chat messages. As a result, default deployments still receive no role-change notification. Could we carry a typed message/signal envelope through the in-process transport and SSE contract, or add a separate typed SSE signal stream with an end-to-end default-transport test?

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -239,4 +241,93 @@ export class ChatRoomMembershipService {
...
+  }: {
+    actorId: Uuid;
+    roomId: Uuid;
+    userId: Uuid;
+    role: ChatRoomAssignableRole;
+  } & ClientMeta) {
+    // Same lock leaveRoom takes, so a role write serializes against the last-moderator check.
+    const changed = await this.drizzle.db.transaction((t) =>
+      withAdvisoryXactLock(t, `chat-room:${roomId}`, async () => {
+        const [room] = await t
```
jakubfilinger-b, line 284: The changeset describes this as a private-room operation, but this lookup does not check `isPublic`. An admin-created public room already has an owner membership, so that owner can promote members through this route. If the intended scope is private rooms, please add `isPublic = false` to the locked lookup and cover the public-room negative case; otherwise the public contract and changeset should explicitly describe the broader scope.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -239,4 +241,93 @@ export class ChatRoomMembershipService {
...
+        if (actorId === userId) {
+          throw new ChatRoomSelfModerationError();
+        }
+        if (target.role === 'owner') {
+          throw new ChatRoomNotModeratorError(roomId);
+        }
+        if (target.role === role) {
+          return false;
+        }
+        await t
```
jakubfilinger-b: `removeMember` does not share this advisory lock, so it can delete the target after the lookup at line 286 but before this update. Because the update result is not checked, the method can still emit a role-change event and signal for someone who is no longer a member. Please share the lock with removal or use `.returning()` and emit only when exactly one row was updated.

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -1633,7 +1666,23 @@ export class ChatService {
...
-    await this.drizzle.db.update(chatRoom).set({ deletedAt }).where(eq(chatRoom.id, roomId));
+    // Same lock the join/leave paths take, so a member joining concurrently either lands
+    // before the roster snapshot (and gets revoked) or after the soft-delete (and is rejected).
+    const members = await this.drizzle.db.transaction((t) =>
+      withAdvisoryXactLock(t, `chat-room:${roomId}`, async () => {
+        const rows = await t
+          .select({ userId: chatRoomMember.userId })
+          .from(chatRoomMember)
+          .where(eq(chatRoomMember.roomId, roomId));
+        await t.update(chatRoom).set({ deletedAt }).where(eq(chatRoom.id, roomId));
```
jakubfilinger-b: The active-room and ownership checks happen before the advisory lock, while this update does not re-check `deletedAt`. Two concurrent delete requests can therefore both pass validation, update the room, and emit duplicate deletion/audit events. Please move validation into the locked transaction and make the update conditional on `deletedAt IS NULL`, using the returned row to decide whether an event should be emitted.

packages/core/src/engagement/chat/drizzle/migrations/0008_majestic_blindfold.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "chat_room_member" ADD COLUMN "role_assigned_at" timestamp with time zone;
```
jakubfilinger-b: Existing moderator and owner rows remain with `role_assigned_at = NULL`, although the schema defines null as a plain member and the planned ownership-transfer logic depends on the earliest assignment. This gives identical elevated roles different semantics depending on when the database was upgraded. Please backfill existing elevated rows, for example from `joined_at`, or explicitly model and handle the legacy-null state.

packages/core/src/engagement/chat/__tests__/chat.router.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  ChatRoomLastModeratorError,
+  ChatRoomNotFoundError,
+  ChatRoomNotMemberError,
+} from '../service/errors/chat-moderation.errors.js';
+
+const ROOM_ID = '9a2f7c11-0000-4000-8000-000000000001';
+const USER_ID = '9a2f7c11-0000-4000-8000-000000000002';
+const CTX = testContext({ auth: { userId: USER_ID } });
+
+// leaveRoom neither reads nor writes a row in the router - it authenticates, then delegates -
```
klaudia-blazyczek-blurify: Please remove these explanatory test comments. The helper name and test cases already show the behavior, and OSS tests must not contain narration.

packages/core/src/react/context/__tests__/realtime-client.test.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Stands in for a vendor adapter: it records what each channel was subscribed with, and
+ * `deliver`/`emit` play the two lanes back the way Ably's named events would.
+ */
+function fakeAdapter() {
+  const channels = new Map<string, RealtimeSubscribeHandlers<Message>>();
+  const closed = vi.fn();
+  const adapter: RealtimeClientAdapter = {
+    subscribe<T>(channel: string, handlers: RealtimeSubscribeHandlers<T>) {
+      channels.set(channel, handlers as RealtimeSubscribeHandlers<Message>);
```
klaudia-blazyczek-blurify: Please avoid this cast. `fakeAdapter` stores a `Message` handler but implements the generic adapter contract, so this can hide a type mismatch. Model the stored handler safely instead.

klaudia-blazyczek-blurify:
Please add PR description, that matches our description pattern

klaudia-blazyczek-blurify (changes requested):

.changeset/chat-room-lifecycle-realtime.md
zaxovaiko, line 1: Remove it, we don't need changeset for patch versions

.changeset/chat-room-member-role.md
zaxovaiko, line 1: Remove all changesets

packages/core/src/audit/plugin.ts
mp-blurify: Done. The event now carries `previousRole`, and the mapper records before/after instead of only the new role. When `playerId` is null it falls back to `changedBy` with `actorType: 'admin'`, so the actor is never dropped.

packages/core/src/engagement/chat/service/chat.service.ts
mp-blurify: Done. The deletion event is emitted right after commit, before the fan-out, and each revoke is isolated and logged. Covered by a rejecting-transport regression test.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
mp-blurify: Done. Signalling is best-effort now: a rejection is logged with context, not returned to the caller. Covered by a test with a rejecting signal transport.

packages/core/src/engagement/chat/service/chat.service.ts
mp-blurify: Good catch, and it went further than the call site: all ten callers in core already passed a user id, so the port was simply misnamed. `revokeClientFromChannel(clientId, channel)` is now `revokeUserFromChannel(userId, channel)`, `subscribe`'s third argument is `userId` to match, and the in-process transport matches on the user, so every connection that user holds is cut. Test covers a second connection of the same user. Breaking for third-party adapters.

packages/core/src/contracts/adapters/realtime.ts
mp-blurify: Done, via the separate stream option. `InProcessRealtimeTransport` now implements `signal` plus a new `subscribeSignal` over a second lane per channel, and `chat.streamSignals` (`GET /chat/signals`) serves that lane under the same access rules as `/chat/stream`. The two lanes cannot cross. End-to-end test on the default transport asserts the signal arrives on the signal lane while the message lane stays empty.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
mp-blurify: Private rooms was the intended scope, so the locked lookup now filters `isPublic = false`, with a public-room negative test.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
mp-blurify: Done. The update uses `.returning()` and the emit is driven by its result, so a member removed concurrently cannot produce a role-change event or signal.

packages/core/src/engagement/chat/service/chat.service.ts
mp-blurify: Done. Validation and `assertOwnership` moved inside the locked transaction, the soft-delete is conditional on `deletedAt IS NULL`, and the returned row decides whether the event is emitted. Test: two concurrent deletes give one success, one `NOT_FOUND`, and one deletion event.

packages/core/src/engagement/chat/drizzle/migrations/0008_majestic_blindfold.sql
mp-blurify: Agreed, that was the wrong call on my side. Migration `0009_backfill_role_assigned_at.sql` sets `role_assigned_at = joined_at` for rows that already held a role above `member`, so an upgraded database orders its moderators the way a fresh one does.

.changeset/chat-room-lifecycle-realtime.md
mp-blurify: Removed. Worth flagging what goes with it: this branch renames `RealtimeTransport.revokeClientFromChannel` to `revokeUserFromChannel`, which breaks third-party adapters, and with no changeset that will not reach any CHANGELOG. Happy to put the note wherever you prefer it to live.

.changeset/chat-room-member-role.md
mp-blurify: Removed both. (For the record they were `minor`, not patch - the branch adds a route, an event field and a transport capability.)

mp-blurify:
Description added under the template, and the PR title fixed: `feat(chat) member role` is missing the colon, so it fails commitlint - and since squash-merge derives the `dev` commit message from the title, merging as-is would have landed a failing commit on `dev`. It is now `feat(chat): assign and revoke the moderator role in a private room`.

All review threads are addressed and answered individually. Two things worth a second look rather than a silent resolve: the port rename `revokeClientFromChannel` -> `revokeUserFromChannel` is breaking for third-party realtime adapters and, with the changesets removed, has no CHANGELOG home; and `leaveRoom` lets an owner leave once they have promoted someone, which this branch is what makes reachable - flagged under Risks.

packages/core/src/server/kernel/realtime-transport.ts
```diff
@@ -85,19 +89,64 @@ export class InProcessRealtimeTransport implements RealtimeTransport {
...
     for (const subscription of Array.from(subscribers)) {
-      if (subscription.clientId === clientId) {
+      if (subscription.userId === userId) {
         subscribers.delete(subscription);
       }
     }
     if (subscribers.size === 0) {
       this.channels.delete(channel);
     }
+    const signalSubscribers = this.signalChannels.get(channel);
```
klaudia-blazyczek-blurify: [BLOCK] If a user has only `streamSignals` open, `channels.get(channel)` is empty and this return skips `signalChannels`. A removed or banned member keeps receiving room signals. Revoke both lanes independently and add a signals-only regression test.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
```diff
@@ -235,7 +244,108 @@ export class ChatRoomMembershipService {
...
+          throw new ChatRoomNotModeratorError(roomId);
+        }
+        if (target.role === role) {
+          return null;
+        }
+        // `removeMember` does not take this lock, so the row read above can be gone by now.
+        // The update decides whether anything actually changed, not the read.
+        const updated = await t
+          .update(chatRoomMember)
+          .set({ role, roleAssignedAt: role === 'member' ? null : new Date() })
```
klaudia-blazyczek-blurify, line 326: [BLOCK] Promoting a member allows the owner to leave: `leaveRoom` counts the owner and moderator together, then deletes the owner membership. `creatorId` still points to the departed user, so no remaining member can manage or delete the private room. Block owner leave until ownership is transferred, and add a regression test.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/server/kernel/realtime-transport.ts
mp-blurify: Confirmed and fixed. The early return when the channel had no payload subscribers skipped the signal lane entirely, so a signals-only connection survived a removal or ban. Both lanes are now pruned independently through a shared helper. Regression test subscribes to signals only, removes the member, then signals the channel and asserts nothing arrives - it fails against the old code.

packages/core/src/engagement/chat/service/chat-room-membership.service.ts
mp-blurify: Done. An owner is now refused outright on `leaveRoom` (`ChatRoomOwnerCannotLeaveError`, mapped to `BAD_REQUEST`) and has to transfer ownership or delete the room. The count check that remains covers the moderator case, so a room that has moderators but no owner still cannot lose its last one. Three tests: owner alone, owner after promoting a moderator - the path that made this reachable - and the sole moderator of an ownerless room. Worth noting ownership transfer does not exist yet, so deleting the room is an owner's only exit today.

jakubfilinger-b (approved):

klaudia-blazyczek-blurify (approved):

PR #117, opened 2026-08-28
feat(identity): mandatory backoffice 2fa and device-bound admin sessions

jakubfilinger-b:
## What

Mandatory Backoffice two-factor authentication and device-bound admin sessions, with the recovery paths that make the policy survivable in production.

Everything is off by default and switched on per operator through `PlatformConfig.adminSecurity`, so an existing deployment upgrades with no behaviour change. Whether Backoffice 2FA is mandatory is a licensing decision the operator owns, not one this package makes for them.

The three commits split as: the policy itself, the corrections better-auth's actual behaviour forced, and the recovery and self-service paths.

### Policy

- `ADMIN_SECURITY_POLICY`, resolved optionally by `AdminGuard`, so enrolment and session-fingerprint checks run on every admin route from one place. Leaving the identity module out keeps the guard's previous behaviour.
- A trusted browser skips the second factor only while **both** better-auth's cookie and an unrevoked row in `admin_trusted_device` exist, which is what makes the window listable, auditable and revocable.
- A mid-session User-Agent change (or an IP change, per `ipChangePolicy`) revokes the session **and** the trust that device held. `ipChangePolicy: 'country'` resolves through `GEO_IP_ADAPTER` and degrades to `off` when nothing is bound - never to `any`, which would sign admins out on every NAT change.
- A 2FA lockout is counted in `user` columns of its own, so it never consumes the password-login budget, and follows the account rather than the browser.

### Recovery

Backup codes were being issued at enrolment and were then unusable - `verify2fa` only ever called `verifyTOTP`, and the contract capped a code at ten characters while better-auth issues eleven. Verification now takes a `method`, spends a backup code, and reports how many remain; a password-guarded route reissues a set. A Super Admin can reset an account to the unenrolled state, which is the only way back once an authenticator is lost along with its codes.

### Trusting a browser without signing out

better-auth mints its trust cookie only on the sign-in leg of a challenge - a call carrying a live session takes the settings branch and gets no cookie. `adminSecurity.trustCurrentDevice` therefore replays the sign-in leg behind the account password and a fresh factor, so the cookie the browser keeps is better-auth's own rather than one this module forged from its internals.

## Compatibility

`UserSchema` gains a required `twoFactorEnabled`; `SessionItem` gains `lastSeenAt`, `deviceLabel`, `browser`, `os`; `AdminSecurityStatus` gains `backupCodesRemaining` and `trustedDeviceDays`. `identity.2fa.enabled` / `.disabled` gain a required `method` at payload version 2. Two identity migrations, both additive.

`identity.2fa.enabled` previously fired on every successful sign-in; it now marks the enrolment challenge only. A consumer counting that event as "sign-ins" will see the number drop to roughly one per account.

## Not in scope

SMS and WebAuthn as second factors: `twoFactor()` is configured with an `otpOptions.sendOTP` seam the identity module binds to a transport-agnostic hook, but no route or flow drives it yet, and there is no passkey plugin. IP-based session location is likewise unbuilt - it needs a `GEO_IP_ADAPTER` binding that ships with no default.


.changeset/backoffice-mandatory-two-factor.md
zaxovaiko, line 1: Remove it

docs/adr/0036-admin-session-security-policy.md
zaxovaiko, line 1: This also

packages/core/src/audit/plugin.ts
```diff
@@ -631,6 +631,15 @@ export async function mapEventToRecord(
     topic === 'identity.user.logout' ||
     topic === 'identity.2fa.enabled' ||
     topic === 'identity.2fa.disabled' ||
+    topic === 'identity.2fa.verified' ||
+    topic === 'identity.2fa.failed' ||
+    topic === 'identity.2fa.backup_codes_regenerated' ||
+    topic === 'identity.2fa.reset' ||
```
damianrzepka: identity.2fa.reset and identity.trusted_device.revoked are in this shared branch, so when playerId is null they get actorId = p['userId'] — but userId here is the target of the action, not the admin who performed the reset/revoke. Both events carry their own actorId field precisely because they're admin-on-behalf actions. Should these two topics get a dedicated branch reading p['actorId'] instead, the way compliance.kyc.updated is handled?

packages/core/src/pam/identity/router/index.ts
```diff
@@ -205,10 +220,85 @@ export function createIdentityRouter(
...
+            userId,
+            context.clientMeta,
+            getSessionId(context),
+          ),
         );
       }),
+
+      resetUserTwoFactor: os.adminSecurity.resetUserTwoFactor.handler(
+        async ({ input, context }) => {
+          const caller = await adminGuard.assert(context, 'admin', 'update');
```
damianrzepka: This clears 2FA on an arbitrary target account, but only checks admin:update, not assertSuperAdmin. The contract comment above calls this a Super Admin operation — should this call adminGuard.assertSuperAdmin(context) instead, plus reject input.userId === caller.userId and require a reason?

packages/core/src/pam/identity/router/index.ts
```diff
@@ -205,10 +220,85 @@ export function createIdentityRouter(
...
+              ip: caller.ip,
+              userAgent: caller.userAgent,
+            }),
+          );
+        },
+      ),
+
+      revokeUserTrustedDevice: os.adminSecurity.revokeUserTrustedDevice.handler(
+        async ({ input, context }) => {
+          const caller = await adminGuard.assert(context, 'sessions', 'revoke');
```
damianrzepka: Same gap as resetUserTwoFactor above — cross-user device revoke only checks sessions:revoke, not super-admin. Worth gating the same way.

packages/core/src/pam/identity/router/index.ts
```diff
@@ -205,10 +220,85 @@ export function createIdentityRouter(
...
+              userId,
+              input.id,
+              getSessionId(context),
+              context.clientMeta,
+            ),
+        );
+      }),
+
+      listAll: os.sessions.listAll.handler(async ({ input, context }) => {
+        await adminGuard.assert(context, 'sessions', 'view');
```
damianrzepka, line 239: This is gated on sessions:view, but listAllActiveSessions returns email/role/ipAddress for every active session platform-wide (players included), with free-text email search. That turns a session-hygiene permission into player PII lookup. Should this also require player:view, or drop email from the response?

packages/core/src/pam/identity/service/admin-security.service.ts
```diff
@@ -0,0 +1,349 @@
...
+  /**
+   * Ends a session whose device or network no longer matches, and drops the trust that
+   * device held - otherwise the next login would silently skip the second factor on the
+   * very device that just failed the check.
+   */
+  private async endCompromisedSession(
+    { userId, sessionId, ip, userAgent }: AdminSessionContext & { sessionId: Session['id'] },
+    mismatch: 'user_agent' | 'ip',
+  ): Promise<never> {
+    await this.sessions.revokeSession(userId, sessionId, userId, { ip, userAgent });
```
damianrzepka: revokeSession(userId, sessionId, userId, ...) and revokeForDevice(userId, userAgent, userId) below pass the affected user's own id as the actor for a guard-forced revoke (fingerprint mismatch). An audit reviewer would read this as a self-revoke rather than a system-triggered kill. Should the actor here be something like 'system'/null instead of userId?

packages/core/src/pam/identity/service/admin-security.service.ts
```diff
@@ -0,0 +1,349 @@
...
+    const [account] = await this.drizzle.db
+      .select({ id: user.id })
+      .from(user)
+      .where(eq(user.id, userId))
+      .limit(1);
+    if (!account) {
+      throw new UserNotFoundError(userId);
+    }
+
+    await this.drizzle.db.delete(twoFactor).where(eq(twoFactor.userId, userId));
```
damianrzepka: The delete + update + revoke-devices + revoke-sessions here (through line 334) aren't wrapped in a transaction. If the process dies between the delete(twoFactor) and the update(user) steps, the account ends up with twoFactorEnabled: true and no twoFactor row — locked out, and the only repair route (resetUserTwoFactor) now requires enrolled 2FA itself to reach it. Worth a db.transaction(...) around this?

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1120,26 +1194,159 @@ export class IdentityService {
...
+        message: 'Two-factor authentication is not enabled for this account',
+      });
+    }
+
+    // Only the cookies the challenge just issued: leaving the live session cookie in
+    // place makes better-auth treat the call as a settings change and skip the trust.
+    const challengeHeaders = new Headers(headers);
+    challengeHeaders.set('cookie', requestCookieHeader(signIn));
+
+    const body = { code: input.code, trustDevice: true };
```
damianrzepka: This verifies a TOTP/backup code below but never touches TwoFactorLockoutService (assertNotLocked/recordFailure), unlike verifyTwoFactor above. The only guard is TWO_FACTOR_PASSWORD_RATE_LIMIT (line 291), which doesn't set onUnavailable: 'deny' and is a no-op with no rate limiter bound. With a hijacked session + known password, this route lets someone grind the second factor outside the lockout path. Also: this sets trustDevice: true for backup_code the same as totp — should a spent recovery code really buy a 30-day trust window?

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1120,26 +1194,159 @@ export class IdentityService {
...
+      playerId: await this.identityReader.getPlayerIdByUserIdSafe(userId),
+      method: input.method,
+      trustedDevice: true,
+      ip,
+      userAgent,
+    });
+    return SUCCESS;
+  }
+
+  async regenerateBackupCodes(
```
damianrzepka, line 1363: Same shape as disableTwoFactor below — password-only, no current-TOTP check, no revocation of existing trust/sessions, and it hands back a fresh set of standing bypass credentials. Worth the same tightening (require a fresh 2FA check, revoke trust/sessions like resetTwoFactor does)?

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1120,26 +1194,159 @@ export class IdentityService {
...
         playerId: await this.identityReader.getPlayerIdByUserIdSafe(userId),
         ip,
         userAgent,
       });
     }
-    return SUCCESS;
+    return { backupCodes: body.backupCodes };
   }
 
   async disableTwoFactor(input: Disable2faInput, reqHeaders: NodeHeaders, resHeaders: Headers) {
```
damianrzepka, line 1403: This only requires input.password, and doesn't revoke trusted devices or other sessions the way resetTwoFactor (admin-security.service.ts) does. With a stolen session + reused password, someone could disable 2FA and re-enrol their own authenticator without ever holding the original TOTP secret. Should this require a fresh 2FA check and the same trust/session teardown as resetTwoFactor?

packages/core/src/pam/identity/plugin.ts
```diff
@@ -108,6 +161,7 @@ export default {
         }),
         c.get(ADMIN_GUARD),
         c.get(EVENT_BUS),
+        makeAdminSecurity(c),
```
damianrzepka: makeAdminSecurity(c) gets called here directly, and separately at line 119 to provide ADMIN_SECURITY_POLICY — two independent instances instead of one resolved through the container. Harmless today since both wrap the same singletons, but it means an overlay that rebinds ADMIN_SECURITY_POLICY wouldn't actually affect these routes. Should this resolve c.get(ADMIN_SECURITY_POLICY) instead of calling the factory again?

damianrzepka (changes requested):
Left comments on what I think are the two things to fix before this ships: resetUserTwoFactor/revokeUserTrustedDevice aren't gated to super-admin even though the contract calls them out as Super Admin operations, and the audit trail for admin-triggered 2FA reset / device revoke / forced session kill records the affected user as the actor instead of the admin who acted — so a takeover step both becomes reachable by a non-super-admin and unreconstructable afterwards. Also flagged trustCurrentDevice skipping the 2FA lockout path, disableTwoFactor/regenerateBackupCodes not requiring a fresh 2FA check or revoking trust, a PII exposure in sessions.listAll, a non-transactional resetTwoFactor, and a small DI wiring issue in plugin.ts.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: On the `trustDevice` part: the reason backup codes accept `trustDevice` on the same footing as TOTP is that they are meant to be a real account-recovery path, not just an enrolment artefact. Before this PR they were issued at enrolment and then unusable (`verify2fa` only ever called `verifyTOTP`, and the contract capped the code at 10 chars while better-auth issues 11), so an admin who lost their authenticator had no way back in. A user who lost their TOTP device and signs in with a recovery code needs a usable session on the other side, the same way GitLab and other platforms let a recovery code carry you through a full login.

That said, the point about grinding the second factor outside the lockout path (no `assertNotLocked` / `recordFailure` here, and `TWO_FACTOR_PASSWORD_RATE_LIMIT` is a no-op with nothing bound) is fair and I'll wire this route through `TwoFactorLockoutService` like `verifyTwoFactor`. If you feel strongly that a spent backup code should only buy a session and not the 30-day trust window, I can gate `trustDevice` to `method === 'totp'` — happy to go that way.

.changeset/backoffice-mandatory-two-factor.md
jakubfilinger-b: Done in f1c1ae64 - removed .changeset/backoffice-mandatory-two-factor.md.

docs/adr/0036-admin-session-security-policy.md
jakubfilinger-b: Done in f1c1ae64 - removed docs/adr/0036-admin-session-security-policy.md as well.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Done in 73291781. verify2fa now forces trustDevice off unless method === 'totp', and trustCurrentDevice only runs verifyTOTP - a backup code clears the challenge and lands a session, but never the 30-day window. The grinding path is closed too: both this route and the new step-up checks go through TwoFactorLockoutService (assertNotLocked/recordFailure/reset).

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Done in 73291781. regenerateBackupCodes now takes a live TOTP on top of the password (checked through TwoFactorLockoutService) and revokes the trusted devices whose trust rode on the old recovery set.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Done in 73291781. disableTwoFactor now requires a fresh TOTP on top of the password and performs the same teardown as the Super Admin reset - revokes the account's trusted devices and live sessions. IdentityService picked up a SessionService dependency for it.

packages/core/src/audit/plugin.ts
```diff
@@ -622,6 +626,20 @@ export async function mapEventToRecord(
...
   }
 
+  // Admin-on-behalf topics: `userId` names the affected account, not the actor - the
+  // acting admin travels in its own `actorId` field, same as compliance.kyc.updated.
+  // `identity.trusted_device.revoked` can also fire with no actorId at all, when
+  // AdminGuard revokes the trust itself on a fingerprint mismatch.
+  if (topic === 'identity.2fa.reset' || topic === 'identity.trusted_device.revoked') {
+    return {
+      ...base,
+      actorType: p['actorId'] ? 'admin' : 'system',
```
damianrzepka: This mapper labels any present `actorId` as `'admin'`, but self-service actions also pass `actorId === userId` (e.g. `adminSecurity.revokeTrustedDevice(userId, id, userId, ...)`, or `disableTwoFactor`/`regenerateBackupCodes` calling `trustedDevices.revokeAllForUser(userId, userId)`). Since `actorId` is truthy in that case, this branch mislabels self-service 2FA/device actions as admin-performed in the audit log.

The sibling `identity.session.revoked`/`revoked_all` mapper right above (line 267) correctly distinguishes self vs. forced via `isForced = !isSystem && rawActorId !== p['userId']`. Should this branch apply the same self-vs-forced check instead of `p['actorId'] ? 'admin' : 'system'`? Also no test in `audit/__tests__/map-event.test.ts` covers this topic.

packages/core/src/audit/plugin.ts
jakubfilinger-b: Went with a dedicated branch for both topics in ba447f58 and tightened it further in 41d8910a after your follow-up: identity.2fa.reset / identity.trusted_device.revoked now resolve actorType as system (no actorId) / admin (actorId !== userId) / player (self-service teardown, actorId === userId), the same split the identity.session.revoked mapper uses. Covered in audit/__tests__/map-event.test.ts.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed in 9197b89e: resetUserTwoFactor now calls adminGuard.assertSuperAdmin(context), rejects input.userId === caller.userId, and requires a reason (recorded on identity.2fa.reset).

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed in 9197b89e: the cross-user revokeTrustedDevice path is gated on assertSuperAdmin the same way.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed in 9197b89e: listAll now asserts player:view on top of sessions:view, since the cross-user list exposes email/role/IP with free-text email search.

packages/core/src/pam/identity/service/admin-security.service.ts
jakubfilinger-b: Fixed in ba447f58: the guard-forced revoke path (fingerprint mismatch) now passes no actorId, and the mappers attribute a missing actorId to the system rather than reading the subject's own userId.

packages/core/src/pam/identity/service/admin-security.service.ts
jakubfilinger-b: Fixed in ba447f58: resetTwoFactor now wraps the delete + update + device/session revocation in db.transaction(...).

packages/core/src/audit/plugin.ts
jakubfilinger-b: Good catch - fixed in 41d8910a. The identity.2fa.reset / identity.trusted_device.revoked branch now applies the same isSystem / isForced (rawActorId !== userId) / self check as the identity.session.revoked mapper above, so disableTwoFactor / regenerateBackupCodes tearing down their own trust (actorId === userId) log as a player self-action, not admin. Added the topic to audit/__tests__/map-event.test.ts (self / cross-user / system-forced / super-admin reset).

packages/core/src/pam/identity/plugin.ts
jakubfilinger-b: Addressed in ba447f58 by building the AdminSecurityService once (let adminSecurity = ... ??= makeAdminSecurity(c)) and sharing that instance between the ADMIN_SECURITY_POLICY provider and the router, so the two-independent-instances problem is gone. I kept it as a shared memo rather than c.get(ADMIN_SECURITY_POLICY) in the router because the token is typed to the narrow AdminSecurityPolicy port (assertEnrolled / assertSessionIntact only) and the router needs the concrete service (resetUserTwoFactor, revokeTrustedDevice, status, ...). Happy to widen the token or split a second one if you'd rather the router resolve through the container - let me know.

damianrzepka (dismissed):

damianrzepka (approved):

zaxovaiko (approved):

PR #118, opened 2026-08-28
chore(agents): trace requests end to end and gate tests by tier

zaxovaiko:
## Summary

- `/review` and `/oss-review` are one skill now, with `--ci` (no questions, exit code from the verdict) and a parseable `FINDING:` / `TRACE:` / `CRITERION:` / `VERDICT:` block.
- The review walks every changed entry point through seven hops (contract, handler, service, query, table, side effects, response) and checks the blast radius: `git grep -w` every changed export, table symbol and SQL table name, open every caller, and read every changed migration for a `DROP`, `RENAME`, type change or `NOT NULL` without default.
- The reviewer agents (`contract-`, `security-`, `quality-reviewer`) are scoped to those callers, not only to "changed files plus immediate callees".
- `docs/standards/testing.md` gets a tier rule: one E2E per new or changed route through `bootTestApp`, unit tests for pure logic, never an in-process mock of the database or a sibling service. `dev`, `module-author` and `qa` point at it.
- `qa` drives a consumer UI with the Playwright CLI first and reports evidence as file paths; the browser MCP is the fallback.

## Why

A review that reads the diff hunk can approve a change that breaks a caller it does not name, or a migration that fails on existing rows. The reviewer agents were scoped to "changed files plus immediate callees", which stops the trace before the database. And the implementation agents kept adding in-process integration tests with mocked query chains, which prove a call order and nothing else.

## Alternatives considered

- A `review:blast-radius` script that printed the caller list and flagged risky SQL. Measured on the same task with `claude -p`: 3 turns vs 5, about 3 percent cheaper, same findings. Not worth a script; the rule tells the agent what to grep instead.
- Bruno collections for the endpoint tier - the Playwright request specs and `bootTestApp` E2Es already give the same black-box test, typed and in CI, so a second runner adds maintenance without coverage.

## Risks

None. Docs and agent prompts only; nothing enforces the new tier yet beyond review.


tools/review/blast-radius.mjs
```diff
@@ -0,0 +1,121 @@
...
+const git = (...cmd) => {
+  try {
+    return execFileSync('git', cmd, { encoding: 'utf8' }).trim();
+  } catch {
+    return '';
+  }
+};
+const lines = (s) => (s ? s.split('\n') : []);
+
+const branchDiff = lines(git('diff', `${base}...HEAD`, '--name-only'));
```
marek-chmielowski-blurify: This helper derives its input from the local checkout via `git diff dev...HEAD`, but the review workflow obtains the PR diff separately with `gh pr diff`. If `/review 118` is run while another branch (or `dev`) is checked out, this command analyzes that branch—or an empty diff—instead of PR #118, so the required blast-radius trace can be missing or unrelated. Please pass the PR head/diff into the helper or run it against an isolated checkout of the PR revision.

tools/review/blast-radius.mjs
```diff
@@ -0,0 +1,121 @@
...
+ */
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const args = process.argv.slice(2);
+const baseIndex = args.indexOf('--base');
+const base = baseIndex >= 0 ? args[baseIndex + 1] : 'dev';
+const asJson = args.includes('--json');
+
+const git = (...cmd) => {
```
marek-chmielowski-blurify: All Git failures are swallowed here, and line 29 then falls back to the working-tree status. With an invalid base ref (for example, a fresh checkout without a local `dev` ref), the helper can silently analyze unrelated files—or no files—and produce a false zero-finding result. Please distinguish an empty diff from a Git error and fail closed when the requested ref or command is invalid.

tools/review/blast-radius.mjs
zaxovaiko: Outdated.

tools/review/blast-radius.mjs
zaxovaiko: Outdated.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
-# Review workflows
+# Review workflow
 
-## OSS review
+One skill, `/review`, for a context-aware, convention-grounded review of the current branch or a given PR.
```
jakubfilinger-b, line 3: After updating an existing checkout, `/oss-review` is still present. `pnpm gen:agents` runs with `delete: false`, so it does not remove the stale mirrors under `.agents`, `.claude`, or `.github`; I reproduced this locally. Please add a safe cleanup for these specific generated directories or keep an explicit migration alias. As written, the “one skill” claim is only true for a fresh checkout.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
...
-- `--agents N` selects one to six reviewers; default one reviewer per applicable dimension.
 - `--base <ref>` changes the base; default `dev`.
 - A PR number reviews that PR with `gh pr diff` and `gh pr view`.
 - Paths limit the review scope.
-- `--fix` permits the orchestrator to apply BLOCK and WARN fixes after reporting.
+- `--agents N` selects one to four reviewers; default one reviewer per applicable dimension.
+- `--fix` applies BLOCK and WARN fixes in the working tree after the report.
+- `--post` publishes the findings to the PR; it needs a PR number.
+- `--yes` skips the confirmation before posting.
+- `--ci` runs without questions: never ask, never post, never fix; print the report and exit `1` on NO-GO, `0` on GO.
```
jakubfilinger-b: `--ci` is only an instruction to the model, so it cannot set the exit status of the CLI process by itself. There is no wrapper or CI job in this change that maps `VERDICT: NO-GO` to exit code 1. Please add an executable wrapper that parses the verdict, or remove the exit-code guarantee.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
...
 
-1. Scope the diff from the supplied PR or `<base>...HEAD`, defaulting to `dev`.
-2. Gather ticket acceptance criteria and unresolved PR discussion once, then distill them to a small context block.
-3. Do not expose raw ticket text, internal URLs, attachments, or people's names in review output.
-4. Assume each changed behaviour is broken until a concrete happy path and hostile path prove otherwise.
-5. Trace empty, falsy, error, unauthorized, concurrent, and repeated inputs where applicable.
-6. For a small diff, review inline from the applicable specialised-agent checklists.
-7. For a larger diff, fan out only matching specialised reviewers in one parallel batch, passing the scoped files and context.
-8. Deduplicate findings and discard any without a concrete location and trigger.
+- For each caller found in the blast radius, run its existing test: `pnpm -F @openora/core vitest related <path>`.
```
jakubfilinger-b: This command currently fails with `ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT` because `@openora/core` has no `vitest` script. Even with `exec`, the default config excludes `.int.test.ts` files, and filtering to core cannot find E2E tests owned by `@openora/testing`. Please select the owning workspace and the correct test script/config for the test tier.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
...
+- A changed event or job payload: every handler, and that it accepts both the old and the new shape during rollout.
+
+Use `docs/catalog.json` or the `oss-dev` MCP tools to list the users of a table, route, or event before grepping.
+
+### Check the migration
+
+For each changed migration, run `pnpm check:drift` and read the generated SQL.
 
-## Pull request review
+- A `DROP`, a `RENAME`, or a column type change breaks the code still deployed; it is a BLOCK unless the code that reads the old shape is already gone.
```
jakubfilinger-b: The exception for code that already stopped reading the old shape allows a column to be removed together with its readers in the same PR, while older instances in a rolling deployment may still require it. This also contradicts the expand/contract rule below. Please require destructive schema changes to happen in a separate later release after the old readers have been fully rolled out.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
...
+## Arguments
 
-### Arguments
-
-- `--agents N` selects one to six reviewers; default one reviewer per applicable dimension.
 - `--base <ref>` changes the base; default `dev`.
 - A PR number reviews that PR with `gh pr diff` and `gh pr view`.
 - Paths limit the review scope.
-- `--fix` permits the orchestrator to apply BLOCK and WARN fixes after reporting.
+- `--agents N` selects one to four reviewers; default one reviewer per applicable dimension.
```
jakubfilinger-b: `--agents N` does not define how reviewers are allocated. For example, `/review --agents 1` on a wallet contract change requires quality, contract, and security coverage, but the workflow does not say which dimensions to merge or omit. Please define priorities for `N < dimensions` and file-splitting rules for `N > dimensions`.

docs/standards/skills/review.md
```diff
@@ -1,78 +1,107 @@
...
-3. Exactly one GO or NO-GO sentence.
+- Every BLOCK or WARN cites a concrete `file:line`, trigger path, and rule or ADR.
+- Drop uncertain, theoretical, duplicate, or tooling-only findings.
+- Do not report style nits already enforced by `pnpm verify` or `pnpm check:boundaries`.
+- Trace called functions whenever a finding depends on their behaviour.
 
-### Posting to the PR
+## Output
+
+Print exactly this block; a CI job parses it line by line.
```
jakubfilinger-b: “Print exactly this block” conflicts with `--post`, which must first show the exact conversational comment bodies without severity markers. Please make the machine-readable block specific to `--ci` and keep human-readable drafts for normal review and posting.

.rulesync/subagents/qa.md
```diff
@@ -14,10 +14,10 @@ You are a QA engineer for the OSS igaming platform. The platform is headless - t
...
 ## A QA pass is TWO deliverables, not one
 
-1. **Automated tests** - co-located Vitest unit/integration tests (via `bootTestApp`), Playwright `request`-context specs for black-box API flows (no browser), and/or a Bruno collection (`.bru`, runnable with `bru run`). Always leave at least one runnable regression artifact.
-2. **A hands-on walkthrough** - actually exercise the feature on the running surface, reading responses/logs live. MANDATORY on every pass, not failure-only: automated assertions miss what a human driving the flow catches (an extra call, a stale value, a 500 in a log). In a browser-facing consumer repo use the chrome-devtools MCP (console + network + DOM); against a bare API drive the flow with real requests and capture evidence (request/response traces).
+1. **Automated tests** - one E2E per touched route in `packages/testing/src/__tests__/<flow>.e2e.test.ts` via `bootTestApp` (happy + one hostile path), or a Playwright `request`-context spec against a running consumer API. Unit tests only for pure logic. Never an in-process test that mocks the database or a service - delete one if you find it in the diff. Always leave at least one runnable regression artifact.
```
jakubfilinger-b, line 17: This removes Bruno from the allowed QA artifacts, but the agent description still advertises Bruno. Please update the frontmatter so agent routing matches the workflow it will actually follow.

jakubfilinger-b (changes requested):
Changes requested: the CI exit-code contract, test-selection command, stale `/oss-review` mirrors, and rolling-deployment migration rule need to be corrected before this is ready to merge.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - `gen:agents` now prunes a mirror whose source left `.rulesync`, so `/oss-review` disappears on the next regen in an existing checkout.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - dropped the exit-code claim; the doc now shows the `grep -q '^VERDICT: GO'` wrapper the CI job runs.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - `exec vitest related` per owning workspace and config. Fixed the same line in `testing.md` too.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - removed the exception; destructive changes ship in a later release.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - added the drop order for N below and the file-group split for N above.

docs/standards/skills/review.md
zaxovaiko: Done in `b2e953ca` - block is `--ci` only; default and `--post` keep the conversational drafts.

.rulesync/subagents/qa.md
zaxovaiko: Done in `b2e953ca`.

docs/standards/skills/review.md
zaxovaiko: Simpler in `63dfb3ba` - dropped the prune script, `delete: true` in `rulesync.jsonc` does it. Checked with `--dry-run` and a real run: only the orphan mirror goes, `.github/workflows` and `.claude/settings.json` stay.

PR #123, opened 2026-08-29
fix(wallet): validate a payout address per network and close the address-cap race

zaxovaiko:
## Summary

- A saved payout destination is now format-checked against its network before it is whitelisted with the custody provider, in both `wallet.withdrawalAddresses.create` and `wallet.withdraw`.
- The per-chain predicate `isWalletAddressValidForNetwork` is exported from `@openora/core/wallet/contract`, so a consumer's frontend validates the form with the same rule the server enforces.
- The 50-address-per-player cap is now race-safe: the count and the insert share one transaction, serialized per player by a transaction-scoped advisory lock.

## Why

The address contract only bounded the string at 8-128 characters, so `abcdefgh` could be saved as a payout destination and submitted as a withdrawal's `destinationAddress`. A payout is irreversible, so the malformed paste comes back as a failed transfer with the funds already debited, and the operator recovers it by hand. Consumers were compensating with their own private regex table in the frontend, which the server never saw.

The cap was a `select count()` followed by an insert, outside any transaction, with a comment admitting the race. It is real: a test firing 60 concurrent saves lands 53 rows before this change and exactly 50 after.

## Alternatives considered

A single `insert ... select ... where (select count(*)) < limit` statement avoids the lock, but under `READ COMMITTED` the count subquery takes no lock, so it does not actually close the race - and a raw-SQL select source is inserted positionally against the table's columns, so the next migration that adds a column silently corrupts the write. The advisory lock makes the plain Drizzle `.values()` insert correct instead.

Tightening the address bound alone (say, per-currency) was not enough: a currency does not identify a chain, and USDT on ERC20, TRC20 and BEP20 has three different address shapes.

## Risks

A network with no entry in the pattern table is accepted and left to the provider, so an operator adding a chain in the Backoffice is never blocked by a table shipped inside core. The patterns accept each chain's testnet form alongside mainnet, because a sandbox asset table issues testnet addresses.

The format check is a first pass, not proof of a live address - the payment provider still rejects a well-formed dead one at payout time.

Existing rows are untouched: this validates writes only, so an address saved before this change is still usable, and the withdrawal path re-checks it only when the request carries a `network`.


zaxovaiko:
Superseded by #125, which carries this branch together with the other wallet, money and player-settings work so the four get reviewed and merged as one. Closing here; the branch itself is untouched.

PR #124, opened 2026-08-29
feat(fx): add the exchange-rate seam (ports, DB-backed reader, refresh job)

zaxovaiko:
## Summary

Adds a new `fx/exchange-rate` module: two provider ports (crypto and fiat, because the two rails will come from different vendors), a read port `EXCHANGE_RATE_READER`, a `exchange_rate_quote` table holding the last known good quote per pair, and a scheduled refresh job that is the only thing which ever calls a provider.

**No provider is bound.** No crypto or fiat rate vendor has been picked yet, so core ships the seam empty: `EXCHANGE_RATE_READER` resolves `null` for every pair until an operator binds a provider through an overlay. Consumers get a working, honestly empty platform rather than a fabricated number. A mocked stub for local dev and CI lives downstream in the consumer repo, never here, and refuses to load in production.

## Why

Display currency, largest-balance defaults, and any cross-currency view all need a rate, and today nothing in the platform can produce one - the only rate table in existence is a hardcoded array in a consumer's swap UI.

Two decisions are baked into the shape:

- **The read path never calls a provider.** `getRate` goes CACHE -> table -> `null`, one hour TTL. A rate here is for display, so a slow or down vendor must never block a page or a balance; only the job talks outward.
- **Crypto and fiat are separate tokens.** They will be separate vendors with separate credentials and separate failure modes, and one rail degrading must not take the other with it. The refresh job records that per rail (`skippedNoProvider`, `skippedNoQuote`, `skippedError`) and never clobbers a stored row on a failed fetch.

Cross pairs are derived through a USD pivot. The pivot is a comparison unit only - it is not a system base currency, no balance is stored in it, and nothing converts through it.

## Alternatives considered

**Fetch on read, cache the result.** Simpler, and it is what a rate library gives you out of the box. Rejected because it puts a third-party HTTP call on a page render: one slow vendor and every balance view stalls. Moving the fetch to a schedule makes the worst case a stale number instead of a hung request.

**One provider port for both rails.** Fewer moving parts, but it forces one adapter to know both a crypto market-data API and a fiat reference API, and a failure on either side becomes indistinguishable.

**Persist the rate next to each transaction.** Deliberately not done here. Under the operator's licence a ledger has to be reproducible, so a rate is only worth storing beside money once it comes from a real, auditable source. Writing a placeholder into an audit column is worse than leaving it empty, and this branch stores no rate on any transaction.

## Risks

- `railFor` and `DEFAULT_CRYPTO_CURRENCIES` moved from the wallet service into shared contracts so the fx module can classify a currency without importing wallet. The wallet service re-exports both, so existing call sites are unchanged.
- New table and migration (`exchange_rate_quote`), additive only.
- The refresh cron defaults to 6 hours. A pair is therefore up to 6 hours stale by design; what a consumer shows for a stale or missing rate is a product decision and is not decided here.

## Also in this branch

`docs(agents)`: the platform dev agent and the consumer builder agent both assumed the published package was the only way to consume core, so a core-side fix and a consumer-side fix cost a publish per round trip. They now point at a locally linked checkout instead, with the two things that actually bite while linked - the lockfile is hidden, so dependencies must not move, and the consumer imports `dist`, so a core edit is invisible until it is rebuilt.

It rides along here because this branch is what proved the need: the stub overlay downstream was written and verified against this branch through exactly that link, before anything was published.


zaxovaiko:
Superseded by #125, which carries this branch together with the other wallet, money and player-settings work so the four get reviewed and merged as one. Closing here; the branch itself is untouched.

PR #125, opened 2026-08-29
feat: exchange-rate seam, display currency and per-currency RG limits; move gift/rain/donate out of core

zaxovaiko:
## Summary

The currency work, as one branch, because every piece lands in the same seam and splitting it would mean rebasing the same files four times. 45 commits, 172 files.

**This PR and betfeel MR !303 must land together.** This branch carries four breaking changes the operator side depends on, and it also *removes* the gift/rain/donate surface that betfeel now owns.

Four things, in dependency order:

1. **An exchange-rate seam** (`fx/exchange-rate`) - ports, a DB-backed read-through reader, and quote storage. Core binds no vendor; an overlay supplies the provider.
2. **Display currency** - a player picks a currency and every amount renders in it. Presentation only: the ledger is never rewritten.
3. **RG limits in the limit's own currency**, with a reduce-only admin override and player self-service.
4. **A Redis-backed realtime transport**, replacing the in-process one.

## Read this first: gift, rain and donate leave core

PR #109 removed the player-to-player transfer surface from `dev` on purpose - it is operator-specific, not platform behaviour. When `dev` was merged into this branch, every one of those deletions was resolved in favour of the branch, which quietly put the whole surface back. Most of them did not even conflict, because this branch had never touched those lines; the generated catalog is what caught it, dropping from 20 modules to 19.

That is now undone. `social-transfers`, both command ports, the four `chat.gift|rain|donate` events, the delegation routes and the plugin registration are gone again, and the catalog is back to **19 modules / 89 events** - exactly what `dev` had.

Core keeps what is genuinely platform-level: `gift`, `rain` and `donate` stay in `CHAT_COMMAND_TYPES` as registry keys, and `chat_command_config` still carries their per-currency `minAmount`/`maxAmount`. An operator configures the command here and implements the mechanics in its own overlay, reading that row through the sanctioned read-only `/schema` subpath. betfeel does exactly that in MR !303.

## Breaking changes

- **`REALTIME_TRANSPORT` now requires Redis.** `InProcessRealtimeTransport` is deleted and the token joins `REQUIRED_DURABLE_SEAMS`. It was the one seam that silently degraded in a multi-replica deployment, and the one the durability guard skipped. Presence is shared through a per-channel sorted set scored by heartbeat, so a crashed replica's members expire instead of lingering.
- **`RG_LIMITS.checkWager`/`checkDeposit` take a leading `tx`.** See below - this one is a defect fix, not a refactor.
- **RG limits carry a currency.** A limit is enforced in the currency it was set in. Pre-existing rows are deliberately left `NULL` and resolved once, on first touch, rather than backfilled: a 100,000 JPY deposit limit rewritten to 100,000 USD is not the same protection.
- **`CommandConfigSchema.minAmount`/`maxAmount` are per-currency maps.** A currency with no entry has no limit enforced - an operator opts a currency *into* a limit rather than getting one by accident. A single flat minimum cannot be correct across USD and BTC at once.
- **The fx refresh cron is gone**, replaced by a read-through cache. There is no longer a refresh cadence to configure.

## Two defects found reviewing this branch against its own standards

**The RG gate could exhaust the connection pool.** `checkWager` took no `tx` and read on its own pooled connection while `WalletCommandsService.debit` held the wallet row `FOR UPDATE` - and the gate's own `resolveLimitCurrency` opened a *third*, a fresh transaction plus an advisory lock on the same slot the caller's transaction was already inside. The `pg` pool defaults to 10 with no connect timeout, so ten concurrent bets took every connection and each then waited for an eleventh that only another of them could release. The check also could not see the caller's uncommitted state, so it was never really "inside the transaction" that `compliance.md` requires. Both methods now take the caller's `tx`, and the gate holds no `DrizzleService` at all, so it cannot regress. Callers with no transaction of their own - the PSP deposit path, the advisory pre-check in `startRound`, the flag sweep, the player's own limits view - pass the pool explicitly.

**The exchange-rate routes were unauthenticated and could reach a vendor.** Neither handler resolved a caller, so both were anonymous, while a missing or hard-stale pair fetches synchronously from the provider. `getRates` takes 50 source currencies per call, and an unresolvable pair is never persisted, so every repeat re-issued the request - attacker-controlled, unbounded vendor spend. Both routes now resolve a caller, a failed synchronous fetch is not retried for the provider timeout window, and the contract comment no longer claims these routes never call a vendor.

## Fail-closed, deliberately

Money paths refuse rather than guess:

- The rate reader returns `null` for both "unavailable" and "too stale", and every caller treats that as a refusal.
- RG's `spendFor` throws `RgRateUnavailableError` rather than counting an unconvertible prior spend as zero. Its three callers each handle it in the way that suits them: the gate refuses the move, the flag sweep skips one limit and continues, and the player's limits view degrades to `null` usage instead of a wrong number or a 500.
- A reduce-only override that only loses a race is still refused inside the transaction, not just in the UI.

## Also in here

Deduplication found in review: rain's per-recipient amount now travels on `chat.rain.distributed` instead of being re-derived by BigInt string slicing in the notification handler (the two agreed only while `totalAmount == perRecipient * recipients.length`, and nothing failed if they drifted); the RG limit insert-or-update is one helper instead of two byte-identical copies; `limitChangeBase` replaces four repeated event schemas.

All 24 changesets are removed at the author's request.

## Verification

`pnpm verify` green: **15/15 tasks**, 66 unit + 77 integration + 16 e2e files, `check:drift` clean.

Two known issues, neither introduced here:

- `rg-monitoring.service.int.test.ts` has two load-sensitive tests. They pass in isolation and under a quiet gate, and failed under heavy parallel load. The fail-closed behaviour they assert is genuinely implemented; the tests are what is fragile.
- Deployed `chat_command_config` rows still hold the old flat-string shape. Core fails safe, so limits will silently not apply until those rows are migrated.

## Tickets

BF-386, BF-387, BF-407, BF-408, BF-457, and the rain fixes from BF-324.


packages/core/src/compliance/drizzle/migrations/0005_reflective_magma.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TYPE "public"."limit_change_kind" AS ENUM('increase', 'removal');--> statement-breakpoint
+DROP INDEX "user_limit_user_id_type_period_key";--> statement-breakpoint
+ALTER TABLE "user_limit" ALTER COLUMN "amount" SET DATA TYPE numeric(38, 18);--> statement-breakpoint
+ALTER TABLE "user_limit" ADD COLUMN "currency" text;--> statement-breakpoint
+-- The session-type limit is measured in minutes, not money, so it always carries the
+-- SESSION_LIMIT_CURRENCY sentinel (see schema/index.ts) - there is no player-currency
+-- question for it, and a sentinel rather than NULL keeps it inside the uniqueness check
+-- below.
+UPDATE "user_limit"
```
klaudia-blazyczek-blurify: [BLOCK] Generated migrations must not contain hand-written data changes. Move this backfill to an approved data-migration mechanism and regenerate the schema migration.

packages/core/src/compliance/adapters/rg-limit-gate.ts
```diff
@@ -0,0 +1,138 @@
...
+  private async check(
+    tx: unknown,
+    userId: string,
+    types: readonly LimitType[],
+    amount: string,
+    amountCurrency: string,
+  ): Promise<RgLimitDecision> {
+    const txn = tx as DrizzleTx;
+
+    const rows = await txn
```
klaudia-blazyczek-blurify, line 64: [BLOCK] Please read the limit after taking this slot lock. A concurrent lower limit can commit before the lock, and this check then uses the stale higher amount to allow a bet that the new limit must reject.

packages/core/src/fx/exchange-rate/adapters/exchange-rate-reader.service.ts
```diff
@@ -0,0 +1,228 @@
...
+  private readonly drizzle: DrizzleService;
+  private readonly pivot: string;
+  private readonly cryptoProvider?: ExchangeRateProvider;
+  private readonly fiatProvider?: ExchangeRateProvider;
+  private readonly cryptoCurrencies?: readonly string[];
+  private readonly freshTtlMs: number;
+  private readonly hardMaxAgeMs: number;
+  private readonly providerTimeoutMs: number;
+  private readonly inFlight = new Map<string, Promise<ExchangeRateQuote>>();
+  private readonly failedUntil = new Map<string, number>();
```
klaudia-blazyczek-blurify, line 85: `updatedAt` is reset on every persist, so an old provider `asOf` is treated as fresh and can drive RG conversion for the full hard-age window. Reject stale quotes or calculate freshness from `providerAsOf`.

packages/core/src/fx/exchange-rate/adapters/exchange-rate-reader.service.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    const quote = await withTimeout(
+      provider.getRate(currency, this.pivot),
+      this.providerTimeoutMs,
+      `${currency}/${this.pivot}`,
+    );
+    if (!quote) {
+      throw new Error(`exchange rate provider returned no quote for ${currency}/${this.pivot}`);
+    }
+
```
klaudia-blazyczek-blurify, line 240: This only rejects `null`. A zero or negative provider rate is cached and can make a converted wager count as zero, bypassing an RG limit. Validate a positive bounded rate and timestamp before persisting.

packages/core/src/fx/exchange-rate/contract/index.ts
```diff
@@ -0,0 +1,44 @@
...
+export const ExchangeRateQuoteSchema = z.object({
+  rate: MoneyAmountSchema,
+  asOf: TimestampSchema,
+});
+export type ExchangeRateQuoteDto = z.infer<typeof ExchangeRateQuoteSchema>;
+
+export const GetExchangeRateInputSchema = z.object({
+  from: CurrencyCodeSchema,
+  to: CurrencyCodeSchema,
+});
```
klaudia-blazyczek-blurify, line 19: These inputs accept arbitrary currency codes. Unique unsupported codes still call the provider and create unbounded cooldown entries, so a signed-in user can cause vendor-cost and memory pressure. Restrict requests to configured currencies and bound the failure cache.

packages/core/src/pam/profile/service/profile.service.ts
```diff
@@ -60,4 +95,75 @@ export class ProfileService implements PlayerProvisioning {
...
+    input: SetDisplayCurrencyInput,
+  ): Promise<DisplayCurrencyInfo> {
+    if (!this.supportedDisplayCurrencies.includes(input.currency)) {
+      throw new UnsupportedDisplayCurrencyError(input.currency);
+    }
+
+    const { row } = await this.ensureProfileRow(userId);
+    const before = row.displayCurrency;
+
+    await this.drizzle.db
```
klaudia-blazyczek-blurify: Please update the player row and call `audit.recordInTransaction` in one transaction. If audit insertion fails after this update, the display-currency change remains without its required audit record.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -78,6 +81,10 @@ export class GamingService {
     if (await this.playEligibility.isRestricted(userId)) {
       throw new RgRestrictedError();
     }
+    const decision = await this.rgLimits?.checkWager(this.drizzle.db, userId, betAmount, currency);
```
klaudia-blazyczek-blurify, line 84: This pre-check uses the requested game currency, but the debit below falls back to the wallet currency. The two checks can calculate different moves; pass the same `currency` into `walletCommands.debit`.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2729,31 +2754,34 @@ export class WalletService {
...
-    // 500 - a double-submitted form is a 409, not an incident.
-    const [row] = await this.drizzle.db
-      .insert(walletWithdrawalAddress)
-      .values({ userId, ...input, ...whitelisted })
-      .onConflictDoNothing()
-      .returning();
-    if (!row) {
-      throw new WithdrawalAddressAlreadyExistsError();
-    }
+    const row = await this.drizzle.db.transaction((txn) =>
```
klaudia-blazyczek-blurify: The cap is checked after the custody vendor has already whitelisted the address. A player at the cap can create orphaned vendor destinations while every local insert is rejected. Reserve a slot durably before the vendor call and release it idempotently on failure.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/compliance/drizzle/migrations/0005_reflective_magma.sql
zaxovaiko: Done in `fe26011a` - `0005` is pure generated DDL again, and the backfill moved to its own `0006_backfill_session_limit_currency.sql` with a journal entry, same as the `0010_backfill_role_assigned_at` precedent on `dev`. Heads-up: `0005` changed content, so a local test DB needs `pnpm db:setup:test:fresh` (it is unreleased, nothing deployed is affected).

packages/core/src/compliance/adapters/rg-limit-gate.ts
zaxovaiko: Done in `fe26011a` - the row is re-read inside `withAdvisoryXactLock` on the slot key now, and an unresolvable currency on the fresh row still refuses.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -884,6 +907,8 @@ export class WalletService {
         .where(eq(wallet.userId, userId));
     }
 
+    await this.assertWithinRgDepositLimit(userId, amount, currency);
```
zaxovaiko: Done in `fe26011a` - the deposit is now two-phase: under the per-user slot lock we check the idempotency row, run the RG check and insert a `pending` row, then call the PSP, then settle with a `status = pending` guard. A PSP throw releases the reservation, and `rg-monitoring` counts `pending` alongside `completed` so a reservation actually holds the limit.

packages/core/src/fx/exchange-rate/adapters/exchange-rate-reader.service.ts
zaxovaiko: Done in `fe26011a` - freshness is calculated from `providerAsOf` now, `updatedAt` is not even selected any more.

packages/core/src/fx/exchange-rate/adapters/exchange-rate-reader.service.ts
zaxovaiko: Done in `fe26011a` - `assertUsableQuote` runs before `persist` and refuses a non-finite, zero, negative or absurd rate, plus an unparseable, future (beyond a 60s skew) or already hard-stale `asOf`.

packages/core/src/fx/exchange-rate/contract/index.ts
zaxovaiko: Done in `fe26011a`, but in `exchange-rate.service.ts` rather than the contract - the supported set is per-operator (`displayCurrencies` plus the pivot), so it cannot live in a static schema. An unsupported code now throws `UnsupportedExchangeCurrencyError` mapped to `BAD_REQUEST` before the reader is touched, so no provider call and no cooldown entry. The cooldown map is also pruned and capped at 512 entries as a second line of defence.

packages/core/src/pam/profile/service/profile.service.ts
zaxovaiko: Done in `fe26011a` - the player update and `audit.recordInTransaction` share one transaction now.

packages/core/src/casino/gaming/service/gaming.service.ts
zaxovaiko: Done in `fe26011a` - the same `currency` goes into `walletCommands.debit`.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `fe26011a` - the slot is reserved first: the capped count plus the row insert happen in one transaction under a per-user advisory lock, then the vendor whitelists, and a throw deletes the reserved row.

klaudia-blazyczek-blurify (dismissed):

PR #130, opened 2026-09-01
feat(mail): consolidate outbound email onto one mail module and send path

klaudia-blazyczek-blurify:
## Summary

Consolidates all outbound email onto one `mail` module and one send path: callers hand `MAIL_DISPATCH` a `{ key, data }` template, it enqueues onto `mail-send`, and the worker resolves the recipient, renders once, and sends off the request path. Breaking: `SEND_EMAIL` and `NOTIFICATION_DELIVERY_ADAPTER` are removed in favour of `EMAIL_SENDER`, and the renderer contract changes shape.

## Why

Email left the platform three different ways - the identity OTP hook rendered and sent inside the request, `compliance/rg.service` rendered and swallowed the error in a warn log, and the notifications module carried a `sendEmail: boolean` flag and mailed the in-app notification's English body. None of the three had a delivery guarantee; several template keys never saw the recipient's language and formatted dates as `en-GB`; rendering ran on the request path. The two ports it went through (`SEND_EMAIL`, `NOTIFICATION_DELIVERY_ADAPTER`) both existed only to move an email, with different argument shapes, and an operator had to bind both.

## Alternatives considered

- **Reuse `NOTIFICATION_DELIVERY_ADAPTER` under a new argument shape** - rejected: a silent change of contract under a familiar name; a compile error on a renamed port is the safer break.
- **Bind the send path in `identity` or `notifications`** - rejected: `notifications` depends on `identity` and `identity` needs to send, so a mail binding in either closes a load-order cycle. The `mail` module declares no `dependsOn` and consumers resolve `MAIL_DISPATCH` lazily.
- **Keep `sendEmail: boolean` on the notification input** - rejected: a flag can't carry the template key or its data, which is why four mails used to go out as the notification's plain English body.
- **Change the broker to ack only after a successful enqueue** - rejected here: it touches every subscriber on the platform. Interim guard is a short enqueue-retry; the durable fix (an outbox row in the caller's transaction) is deferred.

## Risks

- **Breaking:** `SEND_EMAIL` / `NOTIFICATION_DELIVERY_ADAPTER` removed (bind `EMAIL_SENDER`); the notification-delivery port loses its email method; `EMAIL_TEMPLATE_RENDERER.render` now takes the `{ key, data }` template whole and returns `{ subject, html, text }`; `createAuth` swaps `sendEmail` / `templateRenderer` / `getUserLanguage` for one `dispatchOtpMail` hook. ADR-0038 and `docs/adapters/mail.md` cover the move.
- **Rollout:** the `notifications-dispatch` and `kyc-resubmission-notify` job payload shapes changed (mail template instead of a bool; `eventId` now carried). The new schemas accept the previous release's shape on a rolling redeploy - the in-app notification still lands and the mail is skipped rather than dead-lettering.
- **New boot dependency:** the `mail-send` payload is authenticated-encrypted with `AUTH_SECRET`, so the mail plugin now asserts `AUTH_SECRET` is at least 32 characters at register time.
- **Infra:** the Redis service in `docker-compose.yml` and the consumer template gets an fsync'd append-only journal and a volume so queued work survives a restart. A dedicated queue instance is deferred.
- **Deferred:** the ack gap between an event handler and its enqueue (interim retry only), and a dedicated Redis instance for the queue - both tracked for a follow-up.


packages/core/src/mail/plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+      schema: MailSendJobSchema,
+      // Mail providers rate-limit; keep the fan-out modest.
+      options: { concurrency: 5 },
+      handler: async ({ payload }) => {
+        if (!svcRef) {
+          throw new Error('mail: service not constructed yet');
+        }
+        await svcRef.deliver(payload);
+      },
+      onDeadLetter: (jobCtx, error) => {
```
marek-chmielowski-blurify, line 75: The `mail-send` BullMQ worker is registered here and starts consuming immediately (BullMQ Workers autorun by default), but `svcRef` is only set the first time some other module's router factory resolves `MAIL_DISPATCH` — which happens later, after all plugins register (per the module doc comment referencing ADR-0036). This PR also adds Redis AOF persistence + a data volume specifically so queued jobs survive a restart (see the docker-compose.yml diff). On a restart with a leftover `mail-send` job still queued, this worker can start polling and pick it up before the router-factory loop runs, hitting the `throw new Error('mail: service not constructed yet')` above. If retries exhaust while `svcRef` is still null, `onDeadLetter` just logs and returns — skipping the audit-writer entry ADR-0036 expects for RG/KYC mail failures.

Worth gating worker registration on `svcRef` being set, or holding/retrying the first jobs until it is.

packages/core/src/pam/identity/plugin.ts
```diff
@@ -83,8 +74,9 @@ export default {
           drizzle: c.get(DRIZZLE),
           events: c.get(EVENT_BUS),
           identityReader: c.get(IDENTITY_READER),
-          email: c.get(SEND_EMAIL),
-          templateRenderer: c.get(EMAIL_TEMPLATE_RENDERER),
+          // Resolved lazily (router factory runs after every plugin registered) so identity
+          // does not depend on the mail plugin's load order. See ADR-0036.
+          mailDispatch: c.get(MAIL_DISPATCH),
```
marek-chmielowski-blurify, line 143: This resolves `MAIL_DISPATCH` unconditionally via `c.get(...)`, and this plugin declares no `requiresPorts: [MAIL_DISPATCH]`. `iam`'s router factory does the same at `packages/core/src/iam/plugin.ts:76`.

`mail` is registered as a plain module in `extensions.config.ts` (not `kind: 'infra'`), so `applyServiceManifest` drops it from a split deployment unless it's explicitly listed. Running `SERVICE_MANIFEST=identity` (or `iam`) without also including `mail` will hit this `c.get(MAIL_DISPATCH)` with no provider bound. Since neither plugin declares `requiresPorts`, that's a generic "No provider for token MAIL_DISPATCH" crash instead of the descriptive `assertRequiredPorts` boot-time error this codebase built for exactly this scenario (ADR-0024).

Suggest adding `requiresPorts: [MAIL_DISPATCH]` to both `identity` and `iam` plugin definitions.

packages/core/src/contracts/adapters/email-template.ts
```diff
@@ -1,97 +1,153 @@
...
+      }).format(new Date(iso))} UTC`;
 
-export type EmailTemplateRenderer = {
-  render<K extends EmailTemplateKey>(
-    key: K,
-    data: EmailTemplateData[K],
-    locale: string,
-  ): Promise<{ subject: string; body: string }> | { subject: string; body: string };
-};
+const formatMoney = (amount: string, currency: string): string => `${amount} ${currency}`;
```
marek-chmielowski-blurify: `formatMoney` here does plain string concatenation with no thousands-separator grouping, which diverges from `formatMoneyAmount` in `engagement/notifications/plugin.ts` (used for the in-app notification of the same withdrawal events), which does group digits.

Concretely, a $10,000.00 withdrawal renders as `10,000.00 USD` in the in-app notification but `10000.00 USD` in this email — a player-visible inconsistency between channels for the same event. Consider reusing/sharing the existing formatter instead of duplicating the logic.

packages/core/src/contracts/schemas/mail.ts
```diff
@@ -0,0 +1,105 @@
...
+    description: z.string(),
+  }),
+  rgCoolingOffActivated: z.object({ expiresAt: TimestampSchema }),
+  rgCoolingOffLifted: z.object({}),
+  rgSelfExclusionActivated: z.object({
+    expiresAt: TimestampSchema.nullable(),
+    isPermanent: z.boolean(),
+  }),
+  rgSelfExclusionLifted: z.object({}),
+  withdrawalApproved: z.object({ ...WithdrawalDetailsShape, status: z.literal('approved') }),
```
marek-chmielowski-blurify: `withdrawalApproved`/`withdrawalRejected` both carry a `status` literal (`'approved'`/`'rejected'`) that duplicates the discriminated union's own `key`, and it doesn't look like `PLAIN_EMAIL_TEMPLATES.withdrawalApproved`/`withdrawalRejected` in `email-template.ts` ever reads it.

Every caller (`notifications/plugin.ts`) has to hand-write a `status` value that matches the variant it's constructing, with nothing deriving it from `key` — e.g. a caller could pass `status: 'approved'` inside a `withdrawalRejected` payload and nothing would catch the mismatch. Worth dropping the field, or deriving it from `key` instead of duplicating it.

marek-chmielowski-blurify (commented):
Automated review findings (see inline comments for the four tied to this diff). One additional finding has no line in this diff to anchor to:

**`packages/core/generators/src/config.ts:373`** — the adapter scaffold generator's token-choice list still offers the deleted `SEND_EMAIL` and `NOTIFICATION_DELIVERY_ADAPTER` tokens and wasn't updated to offer the new `EMAIL_SENDER` token. A consumer who runs the generator and picks either old token gets a generated overlay plugin importing a token/type this PR deletes (`packages/core/src/contracts/adapters/email.ts` and `notification.ts` are both removed) — guaranteed compile failure with no working replacement offered.

packages/core/src/mail/plugin.ts
klaudia-blazyczek-blurify: Fixed in 7ca0444e. The mail plugin now registers an empty router whose factory calls mailService(c), so the MailService is built during the boot-time router loop - before the mail-send worker picks up any job, and in a mail-only split deployment where nothing else resolves MAIL_DISPATCH. The null-guard in the handler stays as a belt for the sub-millisecond window before that loop runs (it throws, BullMQ retries, the backoff outlasts the window).

packages/core/src/pam/identity/plugin.ts
klaudia-blazyczek-blurify: Fixed in 7ca0444e - requiresPorts: [MAIL_DISPATCH] added to both identity and iam, so a split deployment that omits mail fails at boot with assertRequiredPorts' descriptive error instead of a bare no-provider crash. compliance already resolves it with c.has(...) ? ... : null, so it degrades gracefully and doesn't need the declaration.

packages/core/src/contracts/adapters/email-template.ts
klaudia-blazyczek-blurify: Fixed in 7ca0444e. formatMoneyAmount moved to contracts/schemas/common.ts and the email renderer's formatMoney now calls it, so the email and the in-app notification group thousands the same way. Added a $10,000 assertion to default-email-template-renderer.test.ts.

packages/core/src/contracts/schemas/mail.ts
klaudia-blazyczek-blurify: Fixed in 7ca0444e - status dropped from withdrawalApproved / withdrawalRejected data in the schema and in the notifications builders. The key is the discriminant; nothing read status.

.changeset/mail-module-single-send-path.md
zaxovaiko, line 1: Let's remove it

docker-compose.yml
```diff
@@ -39,9 +39,20 @@ services:
   # See ADR-0028.
   redis:
     image: redis:7-alpine
-    command: ['redis-server', '--maxmemory-policy', 'noeviction']
+    command:
+      [
+        'redis-server',
+        '--maxmemory-policy',
+        'noeviction',
+        '--appendonly',
```
zaxovaiko, line 47: The doc put queue durability in scope specifically because of the consumer template - `consumer/docker-compose.yml` is what an operator copies to production, and it still has neither `appendonly` nor a volume. Let's add the same journal + volume there too.

packages/core/src/mail/service/mail.service.ts
```diff
@@ -0,0 +1,168 @@
...
+  async deliverEncrypted(job: EncryptedMailSendJob): Promise<void> {
+    await this.deliver(this.payloadCipher.decrypt(job));
+  }
+
+  async onDeliveryExhausted(job: MailSendJob, error: Error): Promise<void> {
+    logger.error(
+      { err: error, key: job.template.key, recipient: job.recipient.kind },
+      'mail delivery exhausted retries',
+    );
+    if (!this.audit || !REGULATORY_KEYS.has(job.template.key)) {
```
zaxovaiko, line 176: Hm, the doc wants the audit entry to record both the fact and the result of a send - template key, language, attempt, success or failure - so a delivered regulatory mail should leave an entry too, that's the "player was notified" proof the regulator asks about. Here we only audit the exhausted case, and without language or attempt. The ADR argues volume, but that re-decides what the doc settled - can we sync on this before merge?

packages/core/src/compliance/service/rg.service.ts
```diff
@@ -368,22 +360,23 @@ export class RgService {
...
-    });
+  async notifyLimitUpdated(userId: User['id'], row: LimitRow): Promise<void> {
+    await this.notify(
+      userId,
+      {
+        key: 'rgLimitUpdated',
+        data: {
+          period: row.period,
+          type: row.type,
+          description: row.amount ?? `${row.minutes} minutes`,
```
zaxovaiko: This bakes English ("minutes") into the template data, so a consumer renderer can't translate it - and `amount` comes without a currency, so it can't be money-formatted either. Let's pass `amount`, `currency` and `minutes` as raw fields and let the renderer compose the sentence.

packages/core/src/iam/service/iam.service.ts
```diff
@@ -857,10 +857,14 @@ export class IamService {
       new InvitationNotFoundError(input.email),
     );
 
-    await this.email.send({
-      to: input.email,
-      subject: 'You have been invited as an administrator',
-      body: `Your admin invitation token: ${token}. It expires at ${expiresAt.toISOString()}.`,
+    await this.mailDispatch.toAddress({
+      email: input.email,
+      locale: 'en',
```
zaxovaiko, line 862: Do we want `en` hardcoded here? I know the back-office ships English only for now, but once it localizes the invite mail is stuck like this - maybe take the inviter's language or the platform default. But that's for later, fine with a follow-up.

zaxovaiko (changes requested):

.changeset/mail-module-single-send-path.md
klaudia-blazyczek-blurify: Done in 2a2dca9b - dropped the changeset (matching 3a9d419b's precedent). The breaking-change details live in ADR-0038 and docs/adapters/mail.md.

docker-compose.yml
klaudia-blazyczek-blurify: Added in 2a2dca9b - tools/templates/consumer/docker-compose.yml now has the same appendonly journal + redis_data volume. (packages/create-openora/template/docker-compose.yml is git-ignored and copied from that source at build time.)

packages/core/src/mail/service/mail.service.ts
klaudia-blazyczek-blurify: Fixed in 2a2dca9b. The worker now audits both outcomes: mail.regulatory_delivery.sent on a successful send and .failed on exhaustion, each carrying { templateKey, locale, attempt }. attempt is threaded from the BullMQ job context. ADR-0038's audit section rewritten to match.

packages/core/src/compliance/service/rg.service.ts
klaudia-blazyczek-blurify: Fixed in 2a2dca9b. rgLimitUpdated template data is now { period, type, amount, currency, minutes } - raw fields, no pre-composed English. rg.service passes row.amount / toWireCurrency(row.type, row.currency) / row.minutes; the default renderer composes the sentence (and money-formats the amount). period/type stay as enum keys for the renderer to label.

packages/core/src/iam/service/iam.service.ts
klaudia-blazyczek-blurify: Deliberate for now - the invitee has no account yet, so there's no locale to read, and the back-office ships en only (this is a locked decision in the design brief). Agreed it should take the inviter's language or the platform default once the back-office localizes; leaving as a follow-up per your note.

packages/core/src/contracts/adapters/email-template.ts
```diff
@@ -1,97 +1,130 @@
...
- * (packages/core/src/pam/identity/adapters/) and the SessionResolver-only createAuth()
- * fallback in server/auth/auth.ts - single source of truth so the two never drift.
- * Openora core intentionally ships English-only; overlays that need other languages
- * replace the renderer via ctx.provide(EMAIL_TEMPLATE_RENDERER, () => new MyRenderer()) -
- * `render`'s `locale` param (sourced from IdentityService.resolveUserLanguage) exists for exactly that seam.
- */
-export const DEFAULT_EMAIL_TEMPLATES: {
-  [K in EmailTemplateKey]: (data: EmailTemplateData[K]) => { subject: string; body: string };
-} = {
+const PLAIN_EMAIL_TEMPLATES: { [K in EmailTemplateKey]: PlainTemplate<K> } = {
```
zaxovaiko: Hm, the English copy is still baked in here - the `rg` fix moved it out of `rg.service.ts` into `PLAIN_EMAIL_TEMPLATES`, so it moved rather than went away, and `locale` still only reaches `Intl` for the date. I'm fine with the default renderer shipping English, what I'm missing is the configurable part: today the only way for a consumer to change a subject or a body is to write a TS overlay and redeploy. We want the operator to edit the templates per key and per locale from the back office, the way betfeel does it - so an `email_template` table (`key` + `locale` + `subject` + `body`), admin CRUD, and a renderer that reads it and falls back to `renderDefaultEmail`. Doesn't have to be this PR, but let's have the ticket before we call the mail module done.

Also, should this copy sit in `adapters/email-template.ts` at all? It's the published contract surface, so every consumer ships English strings it never renders - `mail/default-email-template-renderer.ts` feels like the better home.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -110,7 +103,17 @@ export const notificationEventMap: NotificationMapEntry[] = [
       body: `Your withdrawal of ${formatMoneyAmount(p.amount)} ${p.currency} has been approved and is being processed.`,
```
zaxovaiko, line 114: The in-app side is in worse shape than the mail one - `title` and `body` are hardcoded English here with no port at all, so a consumer can't override them even with an overlay. The same applies to `buildKycResubmissionNotification` and the `rg.limit.set` copy further down (`describeLimitValue` still returns `${minutes} minutes`, plus `no prior limit`). Can we give the in-app notifications the same seam as the mail templates, so both channels render through something the operator controls? Fine as a follow-up, but I'd keep it in the same ticket as the email templates.

docs/adapters/mail.md
```diff
@@ -0,0 +1,75 @@
...
+last-wins).
+
+## Template keys
+
+`verifyEmail`, `resetPasswordOtp`, `adminResetPasswordOtp`, `existingAccountSignUp`, `rgLimitUpdated`,
+`rgCoolingOffActivated`, `rgCoolingOffLifted`, `rgSelfExclusionActivated`,
+`rgSelfExclusionLifted`, `withdrawalApproved`, `withdrawalRejected`,
+`kycResubmissionRequested`, `adminInvitation`.
+
+A renderer overlay must cover every key: a missing key falls through to the English
```
zaxovaiko: This one isn't true - `EmailTemplateRenderer` is a single `render()` function, so nothing checks key coverage, and there's no fallback unless the overlay calls `renderDefaultEmail` itself. I couldn't find a CI gate for it either, only the exhaustive map inside our own default renderer. Let's either fix the sentence, or better, export a small `createEmailRenderer(partial)` that merges a partial per-key map over the defaults - then coverage really is optional and the doc is accurate.

zaxovaiko (commented):
Re-checked the fixes, the four earlier threads look good. Three things left, mostly about where the copy lives and who can change it.

zaxovaiko (dismissed):

zaxovaiko (commented):
Good job 💪

zaxovaiko (approved):

PR #131, opened 2026-09-01
fix(chat): tolerate legacy command metadata that exceeds the money decimal scale

jakubfilinger-b:
## What

`toSystemMessage` now canonicalizes the money strings in chat command metadata (`amount`, `perRecipient`) when a message is read: trailing zeros trimmed, fraction truncated to `MONEY_SCALE` (18).

## Why

A `rain` command message persisted by the retired SQL rain split stored its share as `numeric(38,20)`, so `metadata.amount` / `metadata.perRecipient` carry more fraction digits than `MoneyAmountSchema` now allows. `getGlobalMessages` validates the whole batch against `ChatMessageSchema`, so a single such historical row made the endpoint return a 500 for every viewer loading global chat. Observed on a downstream operator's dev environment (5 legacy rows).

## Acceptance

- `getGlobalMessages` returns a legacy rain message whose metadata amounts have 20 decimal places, canonicalized to a contract-valid form, and the row parses against `ChatMessageSchema`.
- New regression test in `chat.service.int.test.ts`.
- `check:types`, `oxlint`, `oxfmt` pass; full `test:integration` green (1228 passing).

### DIFF
--- a/packages/core/src/engagement/chat/service/chat.service.ts
+++ b/packages/core/src/engagement/chat/service/chat.service.ts
@@ -287,15 +287,14 @@ function toConfiguration(record: typeof chatRoomConfiguration.$inferSelect) {
 
 const COMMAND_METADATA_MONEY_KEYS = ['amount', 'perRecipient'] as const;
 
-// A command metadata blob persisted before MoneyAmountSchema capped the decimal
-// scale (the retired SQL rain split rendered its share as numeric(38,20)) holds
-// money strings the current contract rejects. Trim them to canonical form on
-// read so one historical row cannot fail output validation for the whole batch.
 function canonicalizeMoneyString(value: string): string {
   if (!/^\d+\.\d+$/.test(value)) {
     return value;
   }
   const [whole, fraction] = value.split('.') as [string, string];
+  if (fraction.length <= MONEY_SCALE) {
+    return value;
+  }
   const trimmed = fraction.slice(0, MONEY_SCALE).replace(/0+$/, '');
   return trimmed ? `${whole}.${trimmed}` : whole;
 }
--- a/packages/core/src/engagement/chat/service/chat.service.ts
+++ b/packages/core/src/engagement/chat/service/chat.service.ts
@@ -31,7 +31,12 @@ import type {
   Uuid,
   ChatAttachment,
 } from '@openora/core/contracts';
-import { chatBlockLockKey, chatChannel, GLOBAL_CHAT_ROOM_ID } from '@openora/core/contracts';
+import {
+  chatBlockLockKey,
+  chatChannel,
+  GLOBAL_CHAT_ROOM_ID,
+  MONEY_SCALE,
+} from '@openora/core/contracts';
 import {
   eq,
   and,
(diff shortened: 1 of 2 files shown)


packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -280,9 +285,40 @@ function toConfiguration(record: typeof chatRoomConfiguration.$inferSelect) {
   return serializeRow(record, { dateFields: ['createdAt', 'updatedAt'] });
 }
 
+const COMMAND_METADATA_MONEY_KEYS = ['amount', 'perRecipient'] as const;
+
+// A command metadata blob persisted before MoneyAmountSchema capped the decimal
+// scale (the retired SQL rain split rendered its share as numeric(38,20)) holds
+// money strings the current contract rejects. Trim them to canonical form on
+// read so one historical row cannot fail output validation for the whole batch.
```
marek-chmielowski-blurify: This comment is overloaded. It combines root cause, legacy detail, and fix rationale in one block instead of leading with what the code actually does. We do not need to know every decision that led us to create this function. Please remove this comment.

marek-chmielowski-blurify (changes requested):

packages/core/src/engagement/chat/service/chat.service.ts
```diff
@@ -280,9 +285,40 @@ function toConfiguration(record: typeof chatRoomConfiguration.$inferSelect) {
...
+function canonicalizeMoneyString(value: string): string {
+  if (!/^\d+\.\d+$/.test(value)) {
+    return value;
+  }
+  const [whole, fraction] = value.split('.') as [string, string];
+  const trimmed = fraction.slice(0, MONEY_SCALE).replace(/0+$/, '');
```
damianrzepka, line 298: canonicalizeMoneyString trims trailing zeros unconditionally, not only for rows that actually exceed MONEY_SCALE. Checked it directly:

"10.50" -> "10.5"
"100.00" -> "100"
"0.62000000000000000000" -> "0.62"  (the actual legacy case)

The intent here is to fix historical rain-split rows stored as numeric(38,20), but as written it also reformats every already-valid amount/perRecipient on every read of a system message (rain/gift/donate), with no test covering that and nothing about it in the PR description. Doesn't break the contract or lose value, but it's an unintended formatting change beyond the stated fix. Worth guarding with `if (fraction.length <= MONEY_SCALE) return value;` before trimming so only the actually-broken legacy values get touched.

damianrzepka (commented):
Review: fixes the reported bug correctly. Legacy rain-split rows with numeric(38,20) amounts now canonicalize on read and pass ChatMessageSchema; regression test added and passing (ran the full chat.service.int.test.ts suite locally - 138/138). Lint clean on the touched files. Traced getGlobalMessages end to end (viewer id from session, query filters intact, read-only, output matches contract) - same code path also covers getRoomMessages and listAdminRoomMessages, no other reader of chatMessage.metadata in the repo.

One inline note on chat.service.ts:299 - the canonicalization also silently reformats amounts that were never invalid (trims trailing zeros unconditionally), not just the legacy over-precision rows. Doesn't break the contract or lose value, so not blocking, but worth tightening to only touch the actually-broken rows.

GO.

packages/core/src/engagement/chat/service/chat.service.ts
jakubfilinger-b: Usunięte w 71405616.

packages/core/src/engagement/chat/service/chat.service.ts
jakubfilinger-b: Poprawione w 71405616: wartości z częścią ułamkową w granicy MONEY_SCALE pozostają bez zmian, a normalizacja dotyczy tylko nadmiarowej precyzji. Dodałem też test dla zachowania formatów 10.50 i 100.00.

damianrzepka (approved):

marek-chmielowski-blurify (approved):

PR #132, opened 2026-09-02
feat(chat-commands): mention search matches all players, not only online

jakubfilinger-b:
## What

`ChatCommandsService.searchMentions` restricted every result to users currently connected to the room channel, and returned `[]` outright when the room was empty. A typed `@`-autocomplete query now resolves against every player through the directory; an empty query still lists only who is online.

Two supporting changes ride along:

- **`chat_message` index on `(user_id, created_at)`.** The table carried no index on `user_id` at all, so "has this user posted recently" — the shape a downstream consumer needs to gate chat-activity features — degraded to a sequential scan.
- **Seed fix.** `seedDemoData` cleared `game` while `game_round` rows still referenced it. Any E2E suite calling `seedMinimal` against a database with prior rounds failed at setup on `game_round_game_id_game_id_fk`; 13 of 14 suites were red locally for this reason alone. Fixing it is what made the new route-level E2E runnable.

## Why

A downstream operator needs `@`-mentions to reach any player, online or not, and had to build a consumer-side overlay purely to work around this. Fixing it in core removes that need for every consumer.

## Decisions

- **Staff discovery stays online-only, now consistently.** Previously a privileged viewer could find an offline admin by email/username (through `findPlayerIds`) but not by display name, and the same admin became findable by name once online. A staff account with no player row now requires presence in the room whichever branch surfaced it. Non-privileged viewers never see staff accounts at all — `lookupPlayers` already filters them — and that is unchanged.
- **Candidate ids are over-fetched, bounded** (`limit * 3`, hard cap 200). Self/block/ignore/offline-staff exclusions run after the directory lookup, so fetching exactly `limit` let a single excluded match shrink the autocomplete below what the caller asked for.
- Rejected: widening `directory.list({ search })` to match `user.name` so offline admins could be found by display name. It matches only `user.email` today, and changing it would alter the back-office user list as a side effect.

## Acceptance criteria

- A typed query returns a matching player who is not connected to the room.
- A typed query works when the room is empty.
- An empty query returns only online users, and `[]` when nobody is online.
- The caller and any block/ignore-excluded user never appear.
- Output contract `{ userId, username }` unchanged.

## Verification

`searchMentions` had no test coverage before this. Added service unit tests for the branch matrix plus `packages/testing/src/__tests__/chat-mention-search.e2e.test.ts`, a route-level `bootTestApp` walkthrough (offline player found, caller excluded, empty-query scoping, 401 without a session). Nobody holds a realtime connection in a booted test app, so every registered player there is offline — exactly the case the old filter dropped.

Local commits were made with `--no-verify`: the pre-commit hook runs the full integration suite, which was unrunnable before the seed fix in this branch. With the fix, `packages/testing` is 15 files / 137 tests green. A full `pnpm verify` run afterwards had 0 failing tests (906 passed, 321 skipped) with 7 suite files killed at setup by `FATAL 57P01` from Postgres contention on this machine, not by code — CI is the gate.

### DIFF
--- a/packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
+++ b/packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
@@ -40,15 +40,6 @@ const ADMIN_COMMAND_SORT_COLUMNS = {
   updatedAt: chatCommandConfig.updatedAt,
 } as const satisfies Record<AdminCommandSortBy, unknown>;
 
-const MENTION_CANDIDATE_OVERFETCH = 3;
-const MENTION_CANDIDATE_MAX = 200;
-
-// Exclusions (self, blocks, ignores, offline staff) are applied after the directory
-// lookup, so fetching exactly `limit` ids would let a single excluded match shrink the
-// autocomplete below what the caller asked for.
-const candidateFetchSize = (limit: number) =>
-  Math.min(limit * MENTION_CANDIDATE_OVERFETCH, MENTION_CANDIDATE_MAX);
-
 const isStaffRole = (role: string | undefined) => role === 'admin' || role === 'super-admin';
 
 export class ChatCommandsService {
@@ -146,11 +137,16 @@ export class ChatCommandsService {
     }
 
     const onlineIds = new Set(onlineUserIds);
+    const excluded = new Set(await this.blockWriter.getExcludedUserIds(viewerId));
+    excluded.add(viewerId);
     const canSeeAdminUsers = await this.canSeeAdminUsers(viewerId);
     let ids =
       query.length === 0
         ? onlineUserIds
-        : await this.directory.findPlayerIds(query, candidateFetchSize(limit));
+        : await this.directory.findPlayerIds(query, limit, {
+            excludeUserIds: [...excluded],
+            playerOnly: true,
+          });
     if (canSeeAdminUsers && query.length > 0) {
       const onlineAccounts = await this.directory.lookupUsers(onlineUserIds);
       const queryLower = query.toLowerCase();
@@ -168,9 +164,8 @@ export class ChatCommandsService {
     if (ids.length === 0) {
       return [];
     }
-    const excluded = new Set(await this.blockWriter.getExcludedUserIds(viewerId));
     const candidateIds = ids.filter(
-      (id) => id !== viewerId && !excluded.has(id) && (query.length > 0 || onlineIds.has(id)),
+      (id) => !excluded.has(id) && (query.length > 0 || onlineIds.has(id)),
     );
     if (candidateIds.length === 0) {
       return [];
--- a/packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
+++ b/packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
@@ -40,6 +40,17 @@ const ADMIN_COMMAND_SORT_COLUMNS = {
   updatedAt: chatCommandConfig.updatedAt,
 } as const satisfies Record<AdminCommandSortBy, unknown>;
 
+const MENTION_CANDIDATE_OVERFETCH = 3;
+const MENTION_CANDIDATE_MAX = 200;
+
+// Exclusions (self, blocks, ignores, offline staff) are applied after the directory
+// lookup, so fetching exactly `limit` ids would let a single excluded match shrink the
+// autocomplete below what the caller asked for.
+const candidateFetchSize = (limit: number) =>
+  Math.min(limit * MENTION_CANDIDATE_OVERFETCH, MENTION_CANDIDATE_MAX);
+
+const isStaffRole = (role: string | undefined) => role === 'admin' || role === 'super-admin';
+
 export class ChatCommandsService {
   constructor(
     private readonly drizzle: DrizzleService,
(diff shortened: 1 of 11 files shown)


packages/testing/src/__tests__/chat-mention-search.e2e.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  setupTestDb,
+  bootTestApp,
+  registerAndMaterializePlayer,
+  asAdmin,
+  seedMinimal,
+  type TestDb,
+  type TestApp,
+} from '../index.js';
+
+/**
```
marek-chmielowski-blurify: Removed in fc247626.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -40,6 +40,17 @@ const ADMIN_COMMAND_SORT_COLUMNS = {
   updatedAt: chatCommandConfig.updatedAt,
 } as const satisfies Record<AdminCommandSortBy, unknown>;
 
+const MENTION_CANDIDATE_OVERFETCH = 3;
+const MENTION_CANDIDATE_MAX = 200;
+
+// Exclusions (self, blocks, ignores, offline staff) are applied after the directory
+// lookup, so fetching exactly `limit` ids would let a single excluded match shrink the
+// autocomplete below what the caller asked for.
+const candidateFetchSize = (limit: number) =>
```
marek-chmielowski-blurify: Exclusions are now filtered in the directory query before the limit is applied; covered by an integration test. fc247626.

marek-chmielowski-blurify (changes requested):

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
```diff
@@ -128,25 +130,30 @@ export class ChatCommandsService {
...
     const onlineIds = new Set(onlineUserIds);
-    const query = q.trim();
+    const excluded = new Set(await this.blockWriter.getExcludedUserIds(viewerId));
+    excluded.add(viewerId);
     const canSeeAdminUsers = await this.canSeeAdminUsers(viewerId);
     let ids =
       query.length === 0
         ? onlineUserIds
-        : await this.directory.findPlayerIds(query, Math.max(limit, onlineUserIds.length));
+        : await this.directory.findPlayerIds(query, limit, {
```
damianrzepka, line 146: The route now makes this directory search available to every authenticated player, while `findPlayerIds` builds an unescaped `ILIKE` pattern from `q`. With `q=%` or `_`, PostgreSQL treats it as a wildcard and returns the first matching offline players, so an empty room can enumerate usernames. Please escape LIKE metacharacters (or reject wildcard-only input) before the query and cover it with an API/integration test. `getPlayerByUsername` already avoids this wildcard behavior.

packages/testing/src/__tests__/chat-mention-search.e2e.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  return { ...registered, username };
+}
+
+function mentionUrl(q: string, roomId: string = GLOBAL_CHAT_ROOM_ID) {
+  const params = new URLSearchParams({ q, limit: '20', roomId });
+  return `/chat-command/mention-search?${params.toString()}`;
+}
+
+// oxlint-disable-next-line typescript/no-explicit-any -- ad-hoc JSON shape assertions in tests
+async function readJson(res: Response): Promise<any> {
```
damianrzepka: Please remove the `any` and its lint suppression. Project rules prohibit `any` in tests too; parse the response as `unknown` and validate it with `MentionResultSchema.array()` so this E2E also proves the response contract.

damianrzepka (changes requested):
NO-GO: please address the two inline comments before merge. TRACE: `mentionSearch` — contract, authenticated router, service, directory port and query walked; the new public directory-search path permits wildcard enumeration. Acceptance criteria are otherwise met by the implementation and targeted unit/integration tests.

packages/testing/src/__tests__/chat-mention-search.e2e.test.ts
jakubfilinger-b: Usunięte w 6ae84411.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
jakubfilinger-b: Wykluczenia są teraz filtrowane w zapytaniu katalogu przed zastosowaniem limitu; pokryte testem integracyjnym. 6ae84411.

packages/core/src/engagement/chat-commands/service/chat-commands.service.ts
jakubfilinger-b: Escaped LIKE wildcard characters before searching and added an E2E case for a percent-sign query, preventing user enumeration through wildcard input. 02156abd.

packages/testing/src/__tests__/chat-mention-search.e2e.test.ts
jakubfilinger-b: Removed any and the lint suppression. The E2E now parses the response with MentionResultSchema.array(). 02156abd.

marek-chmielowski-blurify (approved):

damianrzepka (approved):

PR #135, opened 2026-09-02
feat(pam): capture and expose a player timezone from the browser

mp-blurify:
## Summary

Captures the IANA timezone the browser reports and persists it on the `player` row, refreshing
it on every session-establishing route, and exposes it on the shared player read model so
consumers can render a stored UTC timestamp on the player's own clock. BF-532

## Why

Nothing in the platform stored a player timezone, so a consumer's backoffice could show a chat
message's UTC timestamp but not what time it was for the player who sent it. The two columns
that look like they could answer are not honest sources: `country` is self-declared at
registration and large countries span several zones, and `registrationIp` is VPN-defeated and
less accurate than the device itself. The browser already resolves the right answer via
`Intl.DateTimeFormat().resolvedOptions().timeZone`, so the platform takes that and stores it.

Capture is deliberately spread across six routes rather than one. Four routes establish a
session, not one — `login`, `verify2fa`, `phoneLoginVerify` and `verifyEmail` — so extending
`LoginInputSchema` alone would have covered a quarter of the cases. `verifyEmail` is easy to
miss because it looks like an email chore, but it is where better-auth's
`autoSignInAfterVerification` mints a new player's *first* session, and it emits
`identity.user.login` like the rest. `register` is the fifth: it mints no session, but the
browser is present at sign-up, so the zone is captured then rather than waiting for a first
login that may be days away — and sign-up and verification can happen on different devices, so
neither one covers the other. `profile.update` is the sixth and the one that makes the feature
useful on day one: a remembered session skips login for weeks, and no existing player
re-authenticates just because a column appeared, so a login-only capture would leave most of the
book reading `-` for a long time.

`timezoneUpdatedAt` is stamped on every capture the tz database accepts, including one that
repeats the zone already stored, so it reads as *last confirmed by a device*. The alternative —
moving it only on a change — makes a player signing in daily from one zone for months
indistinguishable from one not seen since, which is precisely the distinction the read model
ships the timestamp to make.

The value is display metadata and is treated as such throughout. It is device-reported and
trivially spoofable, so it gates nothing, is never evidence of where a player is, and stays out
of responsible-gambling windows and audit records, which remain timezone-free and UTC on purpose.
An unrecognised zone is a silent no-op, never a request failure — a cosmetic column must not cost
a player their session. That includes the empty string a client falls back to when `Intl` gives
it nothing: the contract accepts it and the write drops it, rather than 400ing the login that
carried it.

## Alternatives considered

- **Deriving the zone from `country`.** Rejected: self-declared, and a single country routinely
  spans several zones, so the answer is wrong for a large share of players by construction.
- **IP geolocation.** Rejected: costs a provider, is defeated by any VPN, and is still less
  accurate than the device that already knows.
- **Attaching the zone to every request.** Rejected: spoofable per request, and it invites using
  a request-scoped value to drive a stored decision.
- **Riding `PLAYER_ACTIVITY_TRACKER.touchLastSeen`,** which already fires on every authenticated
  request and is a ready-made seam. Rejected because using it would require the zone on every
  request — exactly the thing ruled out above.
- **Storing the raw client string.** Rejected: `resolveTimezone` canonicalises against the
  runtime's own tz database instead, so the column holds a zone name rather than arbitrary client
  text, and the aliases and casings different browsers report (`US/Pacific`, `europe/warsaw`)
  land on one stored spelling instead of flip-flopping between them.
- **Moving `timezoneUpdatedAt` only when the zone changes.** Rejected: it saves one write on a
  row an authenticated request touches anyway, and costs the column its only useful reading —
  see above.
- **Omitting `timezoneUpdatedAt` from the read model.** Rejected: the zone is a point-in-time,
  device-reported guess, so a player who moved — or whose last session was months ago — renders a
  plausible but wrong local time with nothing to signal it. Shipping the value without its
  staleness marker hands the reader confident-looking wrong data.

## Risks

- **Port change.** `PlayerProvisioning` gains a required `recordTimezone`. An overlay that
  rebinds `PLAYER_PROVISIONING` with its own implementation will fail to compile until it adds
  the method. Deliberate over an optional member: a silently-missing implementation would look
  like a working capture that never writes.
- **Migration.** Two nullable `ADD COLUMN`s with no backfill and no index — additive, no rewrite,
  safe to roll back by ignoring the columns. An uncaptured player reads `null` rather than a
  guess, so the read model gains a value consumers must handle as absent.
- **Silent by design.** A zone the tz database does not recognise is dropped with no feedback to
  the caller, and a capture failure is logged and swallowed. That is the intended trade — it
  guarantees the feature can never fail an auth flow — but it means a client sending a
  consistently bad value sees success and stores nothing. The warn-level log is the only signal.
  The contract keeps one hard bound, a 64-character ceiling: past twice the longest real zone
  name the value is not a zone the platform failed to recognise, it is arbitrary client text.
- **Deferred work.** The platform half is inert until consumers send the zone: no client in this
  repo populates it, so the columns stay null on an upgrade alone. The consumer-side change
  (sending the browser zone on login and on the post-auth profile write, and rendering the
  column) is out of scope here and lands separately after a version bump.
- **Scope discipline.** This must not become an input to responsible-gambling evaluation or any
  session or limit window; making RG timezone-aware is a licence problem, not a feature. Nothing
  under `compliance/` was touched.

## Review follow-ups

Three findings from review, all fixed in `fix(pam): close the timezone capture gaps found in
review`:

- `verifyEmail` carried no `timezone` and never captured, despite minting the first session.
  It now does, after the account gate and before the 2FA branch.
- `TimezoneSchema` lost its `.min(1)`: an empty string was failing contract validation, so a
  client-side `Intl` fallback could 400 a login. It now reaches the same silent no-op as any
  other unrecognised zone.
- `timezoneUpdatedAt` refreshes on every accepted capture rather than only on a change, and the
  column is documented as *last confirmed* in the schema, the read model and the service.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SyLqsaeoprPoPXehsU5BbR

### DIFF
--- a/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
+++ b/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { MoneyAmountSchema, resolveTimezone } from '../common.js';
+import { MoneyAmountSchema, TimezoneSchema, resolveTimezone } from '../common.js';
 
 describe('MoneyAmountSchema', () => {
   it('round-trips a 1-wei ETH amount (18-decimal precision)', () => {
@@ -50,5 +50,19 @@ describe('resolveTimezone', () => {
     // An offset is one moment's arithmetic: it goes wrong the next time DST moves.
     expect(resolveTimezone('+05:00')).toBeNull();
     expect(resolveTimezone('-08:00')).toBeNull();
+  });
+});
+
+describe('TimezoneSchema', () => {
+  it('accepts an empty string, leaving the drop to resolveTimezone', () => {
+    // A client whose `Intl` lookup came back empty must reach the silent no-op, not a 400
+    // on the login that carried the field.
+    expect(TimezoneSchema.safeParse('').success).toBe(true);
+  });
+
+  it('still refuses a value too long to be a zone name', () => {
+    // Twice the longest real zone name: past that it is arbitrary client text, and the
+    // column is not a place to put it.
+    expect(TimezoneSchema.safeParse('x'.repeat(65)).success).toBe(false);
   });
 });
--- a/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
+++ b/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
@@ -26,12 +26,6 @@ describe('MoneyAmountSchema', () => {
 });
 
 describe('resolveTimezone', () => {
-  it('lets a real zone through the offset guard unchanged', () => {
-    expect(resolveTimezone('Europe/Warsaw')).toBe('Europe/Warsaw');
-    expect(resolveTimezone('America/New_York')).toBe('America/New_York');
-    expect(resolveTimezone('UTC')).toBe('UTC');
-  });
-
   it('canonicalises the aliases and casings different browsers report', () => {
     expect(resolveTimezone('europe/warsaw')).toBe('Europe/Warsaw');
     expect(resolveTimezone('US/Pacific')).toBe('America/Los_Angeles');
--- a/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
+++ b/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { MoneyAmountSchema } from '../common.js';
+import { MoneyAmountSchema, resolveTimezone } from '../common.js';
 
 describe('MoneyAmountSchema', () => {
   it('round-trips a 1-wei ETH amount (18-decimal precision)', () => {
@@ -22,5 +22,33 @@ describe('MoneyAmountSchema', () => {
     expect(MoneyAmountSchema.safeParse('0.0000000000000000001').success).toBe(false);
     expect(MoneyAmountSchema.safeParse('100000000000000000000').success).toBe(false);
     expect(MoneyAmountSchema.safeParse('-1').success).toBe(false);
+  });
+});
+
+describe('resolveTimezone', () => {
+  it('accepts an IANA zone the browser reports and returns it unchanged', () => {
+    expect(resolveTimezone('Europe/Warsaw')).toBe('Europe/Warsaw');
+    expect(resolveTimezone('America/New_York')).toBe('America/New_York');
+    expect(resolveTimezone('UTC')).toBe('UTC');
+  });
+
+  it('canonicalises the aliases and casings different browsers report', () => {
+    // Two spellings of one zone must compare equal, or a returning session would look
+    // like a move and churn `timezoneUpdatedAt`.
+    expect(resolveTimezone('europe/warsaw')).toBe('Europe/Warsaw');
+    expect(resolveTimezone('US/Pacific')).toBe('America/Los_Angeles');
+  });
+
+  it('rejects a value the tz database does not recognise rather than passing it through', () => {
+    expect(resolveTimezone('Mars/Phobos')).toBeNull();
+    expect(resolveTimezone('')).toBeNull();
+    expect(resolveTimezone('Europe/Warsaw ')).toBeNull();
+    expect(resolveTimezone("'; DROP TABLE player;--")).toBeNull();
+  });
+
+  it('rejects a bare UTC offset, which Intl accepts but which is not a zone', () => {
+    // An offset is one moment's arithmetic: it goes wrong the next time DST moves.
+    expect(resolveTimezone('+05:00')).toBeNull();
+    expect(resolveTimezone('-08:00')).toBeNull();
   });
 });
--- a/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
+++ b/packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
@@ -26,15 +26,13 @@ describe('MoneyAmountSchema', () => {
 });
 
 describe('resolveTimezone', () => {
-  it('accepts an IANA zone the browser reports and returns it unchanged', () => {
+  it('lets a real zone through the offset guard unchanged', () => {
     expect(resolveTimezone('Europe/Warsaw')).toBe('Europe/Warsaw');
     expect(resolveTimezone('America/New_York')).toBe('America/New_York');
     expect(resolveTimezone('UTC')).toBe('UTC');
   });
 
   it('canonicalises the aliases and casings different browsers report', () => {
-    // Two spellings of one zone must compare equal, or a returning session would look
-    // like a move and churn `timezoneUpdatedAt`.
     expect(resolveTimezone('europe/warsaw')).toBe('Europe/Warsaw');
     expect(resolveTimezone('US/Pacific')).toBe('America/Los_Angeles');
   });
@@ -47,7 +45,6 @@ describe('resolveTimezone', () => {
   });
 
   it('rejects a bare UTC offset, which Intl accepts but which is not a zone', () => {
-    // An offset is one moment's arithmetic: it goes wrong the next time DST moves.
     expect(resolveTimezone('+05:00')).toBeNull();
     expect(resolveTimezone('-08:00')).toBeNull();
   });
@@ -55,14 +52,11 @@ describe('resolveTimezone', () => {
 
 describe('TimezoneSchema', () => {
   it('accepts an empty string, leaving the drop to resolveTimezone', () => {
-    // A client whose `Intl` lookup came back empty must reach the silent no-op, not a 400
-    // on the login that carried the field.
     expect(TimezoneSchema.safeParse('').success).toBe(true);
   });
 
-  it('still refuses a value too long to be a zone name', () => {
-    // Twice the longest real zone name: past that it is arbitrary client text, and the
-    // column is not a place to put it.
-    expect(TimezoneSchema.safeParse('x'.repeat(65)).success).toBe(false);
+  it('accepts arbitrary client text, leaving the drop to resolveTimezone', () => {
+    expect(TimezoneSchema.safeParse('x'.repeat(65)).success).toBe(true);
+    expect(resolveTimezone('x'.repeat(65))).toBeNull();
   });
 });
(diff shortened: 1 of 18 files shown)


packages/core/src/contracts/schemas/identity.ts
```diff
@@ -109,6 +110,10 @@ const credentialsBase = z.object({
 
 export const LoginInputSchema = credentialsBase.extend({
   rememberMe: z.boolean().optional(),
+  // The browser's own IANA zone, captured for rendering a stored timestamp on the player's
+  // clock. Optional and best-effort on every route that carries it: an omitted or
```
klaudia-blazyczek-blurify: [WARN] `verifyEmail` also creates a player session, but it accepts no `timezone` and never calls `captureTimezone`. Please capture it after the email code succeeds.

packages/core/src/contracts/schemas/common.ts
```diff
@@ -61,3 +61,36 @@ export const CurrencyTickerSchema = z
...
+ * eg `Europe/Warsaw`). Bounded rather than enumerated: the value is checked against the
+ * runtime's own tz database at the write (`resolveTimezone`), so a zone a later tzdata release
+ * adds is accepted without a contract change. The longest name in the current database is 30
+ * characters.
+ *
+ * Display metadata only. It is device-reported and trivially spoofable, so it is never
+ * evidence of where a player is and must never gate anything - responsible-gambling windows
+ * and audit records stay UTC on purpose.
+ */
+export const TimezoneSchema = z.string().min(1).max(64);
```
klaudia-blazyczek-blurify: [WARN] Empty or overlong timezone values fail validation before `resolveTimezone` can ignore them. This can reject a valid auth flow although invalid timezones should be best-effort.

packages/core/src/pam/profile/service/profile.service.ts
```diff
@@ -86,13 +87,55 @@ export class ProfileService implements PlayerProvisioning {
...
+    if (!resolved) {
+      return;
+    }
+    await this.drizzle.db
+      .update(player)
+      .set({ timezone: resolved, timezoneUpdatedAt: new Date() })
+      .where(
+        and(
+          eq(player.userId, userId),
+          // `ne` alone is NULL - and so never true - for a player who has never sent one.
```
klaudia-blazyczek-blurify: [WARN] An unchanged timezone does not refresh `timezoneUpdatedAt`, so a recent confirmation can look months stale. Please update it on every valid capture, or define the field as last changed.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/contracts/schemas/identity.ts
mp-blurify: Agreed, and thanks — this was a real gap. `verifyEmail` is where better-auth's `autoSignInAfterVerification` mints a new player's first session and it emits `identity.user.login` like the other three, so calling it a bystander was wrong.

Fixed in 217315ec: `VerifyEmailInputSchema` carries `timezone`, and the service captures right after `assertAccountNotBlocked` and before the `twoFactorRedirect` branch — the emailed code is proof of address ownership, so we are past the credential gate either way, and the 2FA path still gets the zone even though it ends the session it just minted.

Worth noting `register` did not already cover this: sign-up and verification can happen on different devices. New E2E in `player-timezone.e2e.test.ts` registers without a zone and asserts the verify call is what captures it.

packages/core/src/contracts/schemas/common.ts
mp-blurify: Right about the empty string, and I have fixed that one — `.min(1)` is gone in 217315ec. A client that falls back to `''` when `Intl` gives it nothing was failing contract validation, so a cosmetic column could 400 a login, which is exactly what this feature promises it can never do. Empty now reaches `resolveTimezone`'s silent no-op like any other unrecognised zone. Covered by a schema unit test plus an E2E that logs in with `timezone: ''` and asserts 200 with nothing stored.

I kept the 64-character ceiling deliberately, so that half stands as-is. The longest name in the current tz database is 30 characters; past twice that, the value is not a zone we failed to recognise, it is arbitrary client text, and I would rather the contract refuse it than have the write path quietly swallow it. The reasoning is now written into the schema comment so the asymmetry is not mistaken for an oversight.

packages/core/src/pam/profile/service/profile.service.ts
mp-blurify: Took your first option in 217315ec: `timezoneUpdatedAt` now moves on every capture the tz database accepts, including one that repeats the stored zone. The `WHERE` is just `eq(player.userId, …)`.

You put your finger on a genuine contradiction. The DB schema called the column "last changed", but `PlayerSchema` sold it as the staleness signal — and "last changed" cannot carry that reading: a player signing in daily from `Europe/Warsaw` for eight months showed an eight-month-old stamp, indistinguishable from one not seen since. Telling those two apart is the only reason the read model ships the timestamp beside the zone, so the behaviour had to move, not the docs.

The write-churn argument I had used does not survive contact with the fact that `PLAYER_ACTIVITY_TRACKER` touches this row on every authenticated request anyway. Schema, read model and service comments all now say "last confirmed"; the two tests that asserted the old non-churn behaviour were rewritten.

packages/core/src/contracts/schemas/__tests__/common-schemas.test.ts
```diff
@@ -24,3 +24,45 @@ describe('MoneyAmountSchema', () => {
     expect(MoneyAmountSchema.safeParse('-1').success).toBe(false);
   });
 });
+
+describe('resolveTimezone', () => {
+  it('accepts an IANA zone the browser reports and returns it unchanged', () => {
```
zaxovaiko: What is the purpose of the test? xD

zaxovaiko (commented):
Could you please remove all unnecessary comments from the code?

klaudia-blazyczek-blurify:
apart from Volods comments lgtm

packages/core/src/pam/profile/__tests__/profile.service.int.test.ts
```diff
@@ -291,3 +291,109 @@ describe('ProfileService.setMyDisplayCurrency (real PG)', () => {
...
+
+    // Same zone, and the same zone spelled the way another browser reports it - neither is
+    // a move, but both are a device confirming the zone now, which is what the timestamp
+    // reports. The canonical value is what lands, whichever spelling arrived.
+    await svc.recordTimezone(account.id, 'Europe/Warsaw');
+    await svc.recordTimezone(account.id, 'europe/warsaw');
+
+    const [second] = await playersFor(account.id);
+    expect(second?.timezone).toBe('Europe/Warsaw');
+    expect(second?.timezoneUpdatedAt?.getTime()).toBeGreaterThanOrEqual(
```
zaxovaiko: Hm, this one passes with the old behaviour too - if the write gets skipped then `second` just equals `first` and `>=` is still true, so the test can't fail on the thing it's here to catch. Let's use `toBeGreaterThan`. Same in `__tests__/player-timezone.e2e.test.ts`.

packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -439,6 +452,16 @@ export class PhoneLoginService {
       userAgent,
     });
 
+    // Only once the OTP has cleared and the session exists. Display metadata, so a failure
+    // is logged and swallowed - it must never cost the player the session they just earned.
+    if (timezone && this.playerProvisioning) {
+      try {
+        await this.playerProvisioning.recordTimezone(account.id, timezone);
```
zaxovaiko: This is a copy of `captureTimezone` from `service/identity.service.ts` - same guard, same try/catch, same log string. Can we please extract it once and call it from both? Two copies will drift the moment one of them changes.

zaxovaiko (commented):
`verify` is red on `notifications.router.int.test.ts` (the stream test timing out), which looks unrelated to this branch. Worth a re-run anyway - `check:drift` only runs after the tests in `pnpm verify`, so it never actually executed here.

packages/core/src/contracts/schemas/player.ts
```diff
@@ -62,6 +62,16 @@ export const PlayerSchema = z.object({
   totalWagered: MoneyAmountSchema,
   totalDeposits: MoneyAmountSchema,
   lastSeenAt: z.string().nullable(),
+  // The IANA zone the player's browser last reported, for rendering a stored UTC timestamp
+  // on their own clock. Null until a session captures one - an uncaptured player is
+  // honestly unknown, never guessed from `country` or `registrationIp`.
+  timezone: z.string().nullable(),
```
zaxovaiko: Let's use `TimezoneSchema.nullable()` here - it's already imported in this file for `UpdatePlayerProfileInputSchema`, so the read model and the input can share the one source instead of a bare `z.string()`.

packages/core/src/contracts/schemas/common.ts
```diff
@@ -61,3 +61,42 @@ export const CurrencyTickerSchema = z
...
+ * must land in `resolveTimezone`'s silent no-op like any other unrecognised zone, not fail
+ * the login that carried it. The upper bound stays hard - at more than twice the longest real
+ * zone name, an over-long value is not a zone the platform failed to recognise but arbitrary
+ * client text, and the column is not a place to put it.
+ *
+ * Display metadata only. It is device-reported and trivially spoofable, so it is never
+ * evidence of where a player is and must never gate anything - responsible-gambling windows
+ * and audit records stay UTC on purpose.
+ */
+export const TimezoneSchema = z.string().max(64);
```
zaxovaiko: I was wondering about the asymmetry here - `''` now goes through to the silent no-op, but 65 characters still 400s the login. `resolveTimezone` rejects anything the tz database doesn't know anyway, so do we need `.max()` at all? Dropping it would make the "a cosmetic column never costs a session" promise unconditional.

packages/core/src/pam/profile/service/profile.service.ts
```diff
@@ -86,13 +87,53 @@ export class ProfileService implements PlayerProvisioning {
     return this.ensureProfile(userId);
   }
 
+  /**
+   * Stores the IANA zone the browser reported, for rendering a stored UTC timestamp on the
```
zaxovaiko: This is the spot I meant about the comments - most of the block is the PR description restated. Let's keep "display metadata, never gates anything" and the reason for the offset guard in `resolveTimezone`, and drop the rest. `TimezoneSchema` in `schemas/common.ts` is the same.

zaxovaiko (approved):

PR #136, opened 2026-09-03
feat(wallet): provider-ref primitives and balance read for synchronous wallet bridges

marek-chmielowski-blurify:
## Summary

Adds provider-ref primitives to the wallet ledger and exposes balance reads on `WalletReader`, so a synchronous wallet-callback bridge can debit/credit idempotently and report balance without a new adapter port.

## Why

A synchronous wallet-callback bridge needs to (1) resolve a replayed callback to its exact original response instead of reconstructing one from current state, and (2) report a player's current balance, without going through `WalletCommands`. Neither capability existed on the ports an overlay can resolve: `WalletReader` had no way to look up a transaction by provider reference, and balance was only reachable via `WalletService.getBalance`, not on the `WalletReader` port itself.

Specifically, this adds:

- `WalletReader.findByProviderRef` for replay-safe callback resolution.
- Optional `providerRef` on `WalletCommands.debit`/`credit`, writing `providerName`/`providerRefId`/`externalRoundId`/`metadata` onto the `wallet_transaction` ledger row (loss debits are never tagged).
- `GamingService.accumulateExternalRound`, an atomic upsert that creates or accumulates bet/win totals on a `game_round` keyed by `externalRoundId`, safe under concurrent calls.
- A `bet_reversal` wallet transaction type.
- `WalletReader.getBalance`, reusing the existing `readWalletBalance`/`DEFAULT_WALLET_CURRENCY` helpers rather than duplicating the lookup.

## Alternatives considered

A new dedicated adapter port for this protocol was considered and rejected: the aggregator protocol this unblocks is synchronous request/response, which doesn't fit the fire-and-forget `AggregatorAdapter` port. Both extensions land on the existing ordinary `WalletCommands` and `WalletReader` ports instead.

## Risks

None. Additive changes to existing ports and schema; no existing caller behaviour changes.


packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -84,12 +86,19 @@ export class WalletCommandsService implements WalletCommands {
       status: 'completed',
       direction,
       rail: railFor(row.currency, this.platformConfig?.wallet?.cryptoCurrencies),
+      providerName: providerRef?.providerName,
```
damianrzepka, line 91: This ties provider-ref tagging to `wallet_transaction_provider_ref_id_idx`, which is a global unique index on `providerRefId` alone (not scoped by `providerName`). That's also the index the deposit webhook relies on for idempotency (`wallet.service.ts:2693-2708`, `onConflictDoNothing()` + `eq(walletTransaction.providerRefId, event.externalId)`, no `providerName` filter). If an aggregator ever issues a `providerRefId` that collides with a real deposit's `externalId`, the deposit insert silently no-ops and the fallback select resolves to the aggregator's bet row instead — the on-chain deposit never gets credited or retried.

Separately, `writeLedgerRow` here is a plain `insert().values()` with no conflict handling. A genuinely replayed callback (same provider, same ref) will throw a raw Postgres unique-violation instead of returning the original result, unlike the established pattern for `idempotencyKey` in `wallet.service.ts:1060-1099` (`onConflictDoNothing()` + re-select the winner).

Suggest: scope the unique index (and every lookup) to `(providerName, providerRefId)`, and add the same guard here that deposits already use.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -167,6 +168,60 @@ export class GamingService {
...
+        gameId: args.gameId,
+        userId: args.userId,
+        currency: args.currency,
+        externalRoundId: args.externalRoundId,
+        betAmount: betDelta,
+        winAmount: winDelta,
+        status,
+        endedAt,
+      })
+      .onConflictDoUpdate({
```
damianrzepka, line 254: `onConflictDoUpdate` only targets `externalRoundId` — the `set` doesn't check that `userId`/`gameId` match the existing row. If `externalRoundId` ever collides across two different rounds, this will accumulate bet/win amounts onto someone else's round instead of failing. Worth adding a check that the existing row's owner matches (and rolling back / throwing if not) so a mismatch hits the existing "no row" path instead of silently merging state.

packages/core/src/contracts/adapters/wallet-reader.ts
```diff
@@ -1,4 +1,18 @@
 import { createToken } from './token.js';
+import type { WalletTransactionType, WalletTransactionStatus } from '../schemas/wallet-tx.js';
+
+export type WalletProviderTransaction = {
```
damianrzepka, line 4: `WalletProviderTransaction` doesn't carry `walletId`/`userId`. Once a caller resolves a row via `findByProviderRef`, there's no way to verify the returned transaction actually belongs to the player making the request — worth adding the owner id here.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -167,6 +168,60 @@ export class GamingService {
     return { success: true };
   }
 
+  /**
+   * `targetWhere` must repeat the partial index's predicate (schema/index.ts) - Postgres
+   * rejects ON CONFLICT against a partial index without it. Deltas add onto the stored
+   * value, not `excluded.<col>`, so concurrent callbacks each apply their own delta.
+   */
+  async accumulateExternalRound(
```
damianrzepka, line 225: Two things worth flagging: this isn't replay-safe on its own — a redelivered callback with the same delta will double-count turnover. And `game_round.betAmount`/`winAmount` are `numeric(18,2)` while the wallet ledger uses `numeric(38,18)` — a sub-cent crypto bet delta will round to `0.00` here. Might be fine if the eventual bridge always dedupes before calling this, but neither constraint is enforced at this layer.

packages/core/src/contracts/schemas/wallet-tx.ts
```diff
@@ -19,6 +19,7 @@ export const WALLET_TRANSACTION_TYPES = [
   'rain',
   'manual_credit',
   'manual_debit',
+  'bet_reversal',
```
damianrzepka: `bet_reversal` isn't reflected in the GGR formula (`financial-analytics.service.ts:157,169,173`) and doesn't restore RG wager-limit consumption (`wallet-commands.service.ts:119` only checks on `bet`). A reversed bet currently counts toward GGR permanently and burns the player's wager limit. Also noting there's no production writer for this type anywhere in this diff yet (only the test and migration) — assuming that's intentional for the not-yet-wired bridge.

packages/core/src/wallet/adapters/wallet-reader.service.ts
```diff
@@ -71,4 +99,31 @@ export class WalletReaderService implements WalletReader {
...
+      .where(
+        and(
+          eq(walletTransaction.providerName, providerName),
+          eq(walletTransaction.providerRefId, providerRefId),
+        ),
+      );
+    return row ? toProviderTransaction(row) : null;
+  }
+
+  async getBalance(userId: string): Promise<{ balance: string; currency: string }> {
```
damianrzepka: This duplicates `WalletService.getBalance` (`wallet.service.ts:835-846`) statement for statement. Both files already import from `wallet.service.ts` in this diff — could pull this into one shared function there and call it from both places.

packages/core/src/wallet/__tests__/wallet-transaction-direction-backfill.test.ts
```diff
@@ -5,11 +5,16 @@ import { WALLET_TRANSACTION_TYPES, type WalletTransactionType } from '@openora/c
 // drizzle/migrations/0014_abandoned_madripoor.sql. Kept in lockstep here so a future type
 // added to WALLET_TRANSACTION_TYPES fails this test instead of silently backfilling as
 // NULL (or, worse, someone guessing a direction for it).
+// bet_reversal and swap_in/swap_out did not exist when migration 0014 ran (no historical rows
```
damianrzepka: This rationale comment is inside a test file — worth moving to the commit message/PR description and letting the test name + assertions carry it, per the "never in tests" comment convention.

packages/core/src/wallet/__tests__/wallet-reader.service.int.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      status: 'completed',
+      direction: 'debit',
+      rail: 'fiat',
+      providerName: 'aggregator-x',
+      providerRefId: 'ref-2',
+      metadata: JSON.stringify({ balance: '90.00' }),
+    });
+    const original = await svc.findByProviderRef('aggregator-x', 'ref-2');
+
+    // Unrelated activity on the same wallet in between - a replay must still resolve
```
damianrzepka: Same here — this explanation reads better in the commit message than inline in the test.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -140,7 +149,7 @@ export class WalletCommandsService implements WalletCommands {
       return { ok: false, available };
     }
 
-    await this.writeLedgerRow(txn, debitRow, type, amount, 'debit');
+    await this.writeLedgerRow(txn, debitRow, type, amount, 'debit', providerRef);
```
damianrzepka, line 192: This branch calls `writeLedgerRow(txn, debitRow, 'loss', '0', 'debit')` without forwarding `providerRef`, even though every other debit/credit path threads it through. If a caller submits a `loss` event with a `providerRefId` for replay-safety, it gets silently dropped. Either forward it here too, or document that `providerRef` is ignored for `type: 'loss'`.

damianrzepka (commented):
Left inline notes on the two things I'd want resolved before merging: the new `providerRef` write path shares its unique index with the deposit webhook's idempotency check (a refId collision can silently swallow a real deposit), and `accumulateExternalRound`'s upsert doesn't verify the existing row's owner before merging deltas into it. Everything else is smaller stuff (a dropped `providerRef` on the loss branch, a couple of gaps in GGR/RG accounting for `bet_reversal`, some duplicated balance-lookup logic, and two rationale comments that landed in test files instead of the commit message).

No-go until the two inline items are addressed — the rest can land as follow-ups.

packages/core/src/casino/gaming/service/gaming.service.ts
jakubfilinger-b, line 225: This is currently an internal method only. An overlay can resolve `WALLET_COMMANDS`, but `GamingService` is only constructed inside the gaming router and no gaming command token is bound or exported, so the bridge cannot call this in the same transaction without reaching into internals. Please expose it through a gaming-owned command port, bind it in `plugin.ts`, and pass the caller transaction through.

jakubfilinger-b (commented):
NO-GO — the existing wallet and round-integrity blockers, plus the missing supported gaming command seam for the atomic wallet-plus-round update, need to be resolved before merge.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Good catch. Scoped the unique index to (providerName, providerRefId) so a cross-provider ref collision can't happen, and made writeLedgerRow idempotent via onConflictDoNothing + re-select, same pattern as the deposit webhook. debit()/credit() now return the original result on replay instead of throwing. See 50727c4d.

packages/core/src/casino/gaming/service/gaming.service.ts
marek-chmielowski-blurify: Fixed - added a `setWhere` on the onConflictDoUpdate so the update only applies when userId/gameId match the existing row; a mismatch now throws ExternalRoundOwnerMismatchError instead of merging into someone else's round. 50727c4d.

packages/core/src/contracts/adapters/wallet-reader.ts
marek-chmielowski-blurify: Added walletId and userId to WalletProviderTransaction (joined from wallet in findByProviderRef), so a caller can verify a resolved row belongs to the requesting player. 50727c4d.

packages/core/src/casino/gaming/service/gaming.service.ts
marek-chmielowski-blurify: Fixed the precision half - game_round.betAmount/winAmount now use the shared MONEY_PRECISION/MONEY_SCALE (38,18) instead of (18,2), so a sub-cent crypto delta no longer rounds to zero. Left the replay-safety half alone: accumulateExternalRound has no callback/dedup id to key off today, and I don't want to guess at that shape ahead of the actual bridge - it's the same call-site's responsibility to dedupe before calling in, which you already flagged as the likely intent. Happy to revisit once the bridge PR defines its delivery semantics. 50727c4d.

packages/core/src/contracts/schemas/wallet-tx.ts
marek-chmielowski-blurify: Fixed the GGR half - bet_reversal is now netted into the GGR formula (financial-analytics.service.ts) alongside bet/win. Left the RG wager-limit restoration alone: RgLimitGate's wager consumption reads game_round.betAmount, not the wallet_transaction ledger, so a reversal would need to walk back to its originating round and there's no linkage or writer for that yet in this PR (as you noted). Tracking it as follow-up for whenever bet_reversal gets a real writer. 50727c4d.

packages/core/src/wallet/adapters/wallet-reader.service.ts
marek-chmielowski-blurify: Fixed - pulled it into a shared resolveWalletBalance(txn, userId) in wallet.service.ts and both WalletService.getBalance and WalletReaderService.getBalance now call it. 50727c4d.

packages/core/src/wallet/__tests__/wallet-transaction-direction-backfill.test.ts
marek-chmielowski-blurify: Fixed - dropped the rationale block, the type lists carry the same information. 50727c4d.

packages/core/src/wallet/__tests__/wallet-reader.service.int.test.ts
marek-chmielowski-blurify: Fixed - dropped it. 50727c4d.

packages/core/src/wallet/service/wallet-commands.service.ts
marek-chmielowski-blurify: Fixed - loss now threads providerRef through to writeLedgerRow like every other branch. 50727c4d.

packages/core/src/casino/gaming/service/gaming.service.ts
marek-chmielowski-blurify: Fixed - added a GAMING_COMMANDS port (accumulateExternalRound) bound in gaming/plugin.ts, same lazy-memoized pattern as TAG_EVALUATION_COMMANDS. A caller resolves the token and passes its own tx through, no more reaching into GamingService. 50727c4d.

jakubfilinger-b (dismissed):

jakubfilinger-b (approved):

PR #138, opened 2026-09-03
feat(identity): add security controls and phone verification

damianrzepka:
## Summary

Headless identity/notification primitives for a consumer Security settings page (betfeel BF-517), per the reviewed cross-repo contract:

- `GET /identity/security/me` / `POST /identity/security/login-withdrawal-alerts` / `POST /identity/phone-verification/{request,confirm}`, exported React hooks: `useMySecurityControls`, `useSetLoginWithdrawalAlerts`, `useRequestPhoneVerification`, `useConfirmPhoneVerification`.
- Password policy raised to 12 chars for writes (registration/reset/change); sign-in stays on its own, uncapped schema so existing 8–11 char passwords still authenticate. `passwordMeetsPolicy` persisted, defaults `false` for legacy accounts.
- Self-service phone verification: dedicated `phone_verification_session` table bound to `userId` + `sessionId`, fresh re-authentication (current password, plus TOTP if 2FA is enabled) with a 5-minute re-auth TTL, rate limits, resend cooldown, attempt cap, uniqueness check. The re-auth step never mints a session and never emits `identity.authentication.succeeded`.
- `loginWithdrawalAlertsEnabled` preference: player-only, rejected without a verified email, auto-disabled on email change.
- New `identity.authentication.succeeded` event fired only for a genuinely new session (password, phone, completed TOTP challenge, email-verification auto-sign-in) — never for session refresh or 2FA enrollment.
- Notification delivery goes through the durable job queue with idempotency keys and retry/backoff; withdrawal alerts enrich the existing `wallet.withdrawal.requested` dispatch (no duplicate in-app notification), login alerts use a dedicated queue, both re-check the live preference + verified-email state at actual delivery time.
- Audit records added for both new mutations (`identity.phone.verified`, `identity.security.login_withdrawal_alerts.updated`).

**Known interim limitation:** phone login is refused outright (`FORBIDDEN`) for accounts with 2FA enabled, rather than chaining an SMS OTP with a TOTP challenge — safer than a silent single-factor downgrade, but phone won't work as a second login method for 2FA accounts until a dedicated handoff replaces this.

Out of scope (consumer-owned): score calculation, tiers, checklist UI, copy/i18n, SMS/email vendor credentials, Withdrawal PIN and Anti-Phishing status.

## Test plan

- `pnpm check:types`, `pnpm check:lint`, `pnpm check:drift` — all clean.
- New/changed integration and unit tests (phone verification, security controls, `identity.authentication.succeeded`) — passing locally against Postgres/Redis.


marek-chmielowski-blurify (commented):
Automated code review findings (correctness, duplication, design, efficiency).

packages/core/src/pam/identity/service/phone-verification.service.ts
```diff
@@ -0,0 +1,325 @@
...
+            .update(phoneVerificationSession)
+            .set({ failedAttempts })
+            .where(eq(phoneVerificationSession.id, otp.id));
+        }
+        throw phoneVerificationInvalidError();
+      }
+
+      await tx.delete(phoneVerificationSession).where(eq(phoneVerificationSession.id, otp.id));
+      const [updated] = await tx
+        .update(user)
```
marek-chmielowski-blurify: **Correctness:** `confirm()` can hit the DB's `user_phoneNumber_unique` constraint with no error handling, because `request()`'s uniqueness check only looks at already-bound `user.phoneNumber`, not other users' concurrently pending `phoneVerificationSession` rows.

_Failure scenario:_ two different accounts request phone verification for the same number within the 5-minute window (nothing stops this, since neither has claimed the number yet). Whichever confirms first succeeds; the second user's `confirm()` throws a raw Postgres unique-violation instead of a clean `ORPCError('CONFLICT', ...)` — as the codebase's own convention (`isUsernameCollision` in `user-commands.service.ts`) would suggest — leaking a 500 to the client.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1254,6 +1280,15 @@ export class IdentityService {
         ip,
         userAgent,
       });
+      if (!sessionUserId) {
+        this.events.emit('identity.authentication.succeeded', {
```
marek-chmielowski-blurify, line 1289: **Correctness:** `verifyTwoFactor()` hardcodes `method: 'totp'` when emitting `identity.authentication.succeeded` for every completed 2FA sign-in, even when the challenge was cleared with `input.method === 'backup_code'`.

_Failure scenario:_ a player signs in using a single-use backup/recovery code instead of an authenticator app; the emitted event (consumed by the security login-alert email and by audit) reports `method: 'totp'`, misrepresenting a recovery-credential sign-in — a materially different fraud/account-takeover signal — as a normal authenticator login. `AUTHENTICATION_METHODS` doesn't even include `'backup_code'`, so this is a schema gap, not just a copy-paste slip.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1630,13 +1719,32 @@ export class IdentityService {
...
       body: { newEmail: input.newEmail },
       headers,
       asResponse: true,
     });
     await ensureOk(res);
     this.forwardCookies(res, resHeaders);
+    if (userId) {
+      const [disabled] = await this.drizzle.db
+        .update(user)
+        .set({ loginWithdrawalAlertsEnabled: false })
```
marek-chmielowski-blurify, line 1781: **Correctness:** `changeEmail()` disables `loginWithdrawalAlertsEnabled` as soon as better-auth's `/change-email` endpoint returns `status: true`, but per better-auth's own route source that response fires both on the anti-enumeration no-op (target address already used by someone else — nothing changes) and on the normal verified-account confirmation-email path, where the real email update only happens once the confirmation link is clicked.

_Failure scenario:_ a player with alerts enabled requests an email change to an address already owned by another account, or simply never clicks the confirmation link; better-auth silently no-ops/defers the update, but `identity.service.ts` immediately flips `loginWithdrawalAlertsEnabled` to `false` and emits the audit event. The player's email never actually changed and remains verified, yet their security alerts are silently turned off with no user-visible cause.

packages/core/src/pam/identity/service/phone-login.service.ts
```diff
@@ -399,8 +399,16 @@ export class PhoneLoginService {
       },
     );
 
-    // Mint the session directly - bypasses better-auth's TOTP plugin chain by design,
-    // so phone login never triggers a second factor.
+    // This flow mints sessions directly and cannot hand a challenge to better-auth's
+    // pending-2FA cookie. Refuse rather than silently downgrade a 2FA-protected
+    // account to a single SMS factor; a later dedicated handoff may replace this.
+    if (account.twoFactorEnabled) {
```
marek-chmielowski-blurify, line 418: **Correctness / cost:** `requestOtp()` sends a real SMS OTP without checking `twoFactorEnabled`, even though `verifyOtp()`'s `FORBIDDEN` guard (right before session minting) means any 2FA-enabled account's phone login can never succeed.

_Failure scenario:_ every phone-login request against a phone-verified, 2FA-enabled account (attacker-guessable, or just a confused legitimate user) triggers a real, billable SMS send and a stored OTP session, for a login path that is deterministically rejected at verify-time. This is an avoidable SMS-cost/DoS vector caused by placing the 2FA guard at the wrong layer — it should be checked here, before the SMS is sent, not only at the end of `verifyOtp()`.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1511,9 +1546,56 @@ export class IdentityService {
...
 
+  async getSecurityControls(reqHeaders: NodeHeaders): Promise<SecurityControls> {
+    const userId = await this.currentUserId(nodeHeadersToHeaders(reqHeaders));
+    if (!userId) {
+      throw new ORPCError('UNAUTHORIZED', { message: 'Not signed in.' });
+    }
+    return this.securityControlsFor(userId);
+  }
+
+  async setLoginWithdrawalAlerts(
```
marek-chmielowski-blurify, line 1583: **Authorization:** `setLoginWithdrawalAlerts` has no player-only enforcement (no `adminGuard.assert(...)` equivalent used elsewhere in this router), despite the PR description stating the preference is player-only.

_Failure scenario:_ an authenticated admin/backoffice user can call `POST /identity/security/login-withdrawal-alerts` and enable the preference for their own account, since the handler only checks `currentUserId` + `emailVerified`, contradicting the intended scope of the feature.

marek-chmielowski-blurify, line 1583: **Efficiency:** `setLoginWithdrawalAlerts` calls `securityControlsFor` (a DB round trip) twice — once for `before`, once again for the return value — when only the boolean that was just written differs.

_Failure scenario:_ every toggle of this setting costs two `SELECT`s on `user` instead of one; on a settings endpoint exercised from a UI toggle this doubles read load for information already known after the write (`input.enabled`).

docs/catalog.json
```diff
@@ -265,18 +266,22 @@
...
+        "identity.confirm",
         "identity.disable2fa",
         "identity.enable2fa",
         "identity.list",
         "identity.listAll",
         "identity.listMine",
         "identity.login",
+        "identity.loginWithdrawalAlerts",
         "identity.logout",
         "identity.me",
```
marek-chmielowski-blurify, line 279: **Tooling:** the generated route catalog lists `"identity.me"` twice, and assigns ambiguous leaf-only ids (`"identity.confirm"`, `"identity.request"`, `"identity.loginWithdrawalAlerts"`) to the new nested `security.*`/`phoneVerification.*` routes, because the generator flattens nested route groups to their bare last segment instead of the full dotted path.

_Failure scenario:_ anything reading `docs/catalog.json` — including this repo's own `oss-dev` MCP tools, which `CLAUDE.md` tells agents to use instead of grepping — sees a duplicate `"identity.me"` entry and cannot distinguish the new phone-verification confirm/request routes from any other module's routes named `confirm`/`request` by id alone.

packages/core/src/pam/identity/service/phone-verification.service.ts
```diff
@@ -0,0 +1,325 @@
...
+const MINUTE_MS = 60 * 1000;
+const OTP_TTL_MS = 5 * MINUTE_MS;
+const RESEND_COOLDOWN_MS = MINUTE_MS;
+const MAX_VERIFY_ATTEMPTS = 5;
+const REAUTH_TTL_MS = 5 * MINUTE_MS;
+
+const REQUEST_RATE_LIMIT = { limit: 3, windowMs: 15 * MINUTE_MS, onUnavailable: 'deny' } as const;
+const VERIFY_RATE_LIMIT = { limit: 10, windowMs: 5 * MINUTE_MS, onUnavailable: 'deny' } as const;
+
+function hashCode(code: string) {
```
marek-chmielowski-blurify: **Duplication:** `hashCode`/`generateCode` and `nodeHeadersToHeaders` are byte-for-byte reimplementations of existing helpers already defined in `phone-login.service.ts` and the shared pam header-mapper, and `assertFreshReauthentication`'s TOTP step-up sequence duplicates `IdentityService.assertFreshSecondFactor`.

_Failure scenario:_ a future fix to OTP hashing/generation or header normalization (e.g. adding a pepper, fixing multi-value header handling) applied to the canonical copy silently misses this file, letting the two OTP flows diverge in security-relevant behavior over time.

packages/core/src/contracts/schemas/identity.ts
```diff
@@ -181,6 +181,32 @@ export const Disable2faInputSchema = z.object({
   code: TotpStepUpCodeSchema,
 });
 
+export const SecurityControlsSchema = z.object({
+  passwordMeetsPolicy: z.boolean(),
+  emailVerified: z.boolean(),
+  phoneNumber: E164PhoneSchema.nullable(),
+  phoneVerified: z.boolean(),
+  twoFactorEnabled: z.boolean(),
+  loginWithdrawalAlertsEnabled: z.boolean(),
```
marek-chmielowski-blurify, line 194: **Design:** a single `loginWithdrawalAlertsEnabled` boolean is overloaded in the public contract to gate two conceptually distinct alert types — new-sign-in alerts and withdrawal-request alerts — via the same `canReceiveLoginWithdrawalAlerts` check in `notifications/plugin.ts`.

_Failure scenario:_ a player who wants withdrawal-request confirmations but not every new-device sign-in emailed (or vice versa) cannot express that preference today; splitting them later requires a breaking `SecurityControlsSchema`/`SetLoginWithdrawalAlertsInputSchema` change plus a data migration to carry forward the existing single flag.

packages/core/src/pam/identity/service/phone-verification.service.ts
damianrzepka: Confirmed and fixed in 8f715a7f — `confirm()` now catches the `user_phoneNumber_unique` violation and throws `ORPCError('CONFLICT', { message: 'Phone number is unavailable.' })` instead of leaking the raw DB error, matching the `isUsernameCollision` convention in `user-commands.service.ts`. Added a regression test that races two accounts for the same number.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: Confirmed and fixed in 8f715a7f — emits `method: input.method` instead of the hardcoded `'totp'`. Also added `'backup_code'` to `AUTHENTICATION_METHODS` in contracts/schemas/events.ts, which was missing entirely as you noted. Added a test asserting `identity.authentication.succeeded` carries `method: 'backup_code'` on a recovery-code sign-in.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: Confirmed and fixed in 8f715a7f. Checked better-auth's `/change-email` route source directly: all its branches (the anti-enumeration no-op, the deferred confirmation-email flow, and the synchronous `updateEmailWithoutVerification` update) return the identical `{ status: true }` shape, so `res.ok` alone can't tell them apart. The disable is now scoped with an extra `eq(user.email, newEmail)` in the WHERE clause, so it only fires once the address has actually changed. Added tests for both the real-change and no-op cases.

packages/core/src/pam/identity/service/phone-login.service.ts
damianrzepka: Confirmed and fixed in 8f715a7f — a 2FA-enabled account's phone number now falls into the existing anti-enumeration shadow path (no real SMS, no session row), since `verifyOtp()` would reject it anyway. Kept it inside the anti-enumeration branch rather than a separate early return so a caller still can't distinguish a 2FA-enabled number from an unknown/unverified one by response shape. Added a regression test.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: Confirmed and fixed in 8f715a7f — added a `user.role === 'player'` check (throwing FORBIDDEN otherwise), matching the `eq(user.role, 'player')` pattern used elsewhere in this module (e.g. `admin-user-directory.ts`, `identity-reader.service.ts`). Added a test covering an admin-role account being refused.

damianrzepka: Confirmed and fixed in 8f715a7f — the trailing `securityControlsFor(userId)` re-read is gone; the return value is now built from `before` spread with the field we just wrote (`{ ...before, loginWithdrawalAlertsEnabled: input.enabled }`), since `before` and the just-applied update are all the response needs.

docs/catalog.json
damianrzepka: Confirmed and fixed in 8f715a7f, in the generator (not the generated file). `collectModules()`'s route regex flattened every nested route group to its bare last key. Replaced it with `extractRoutes()`, which tracks nested `key: {` groups by indentation and builds each route's id as its full dotted path from the router root. Regenerated `docs/catalog.json` — this also fixed pre-existing duplicate/ambiguous ids in `analytics` and `wallet` that predated this PR (e.g. `wallet.list` appearing 4 times, `wallet.get` 3 times) once their own nested groups hit the same bug.

packages/core/src/pam/identity/service/phone-verification.service.ts
damianrzepka: Confirmed on the `hashCode`/`generateCode`/`nodeHeadersToHeaders` part, fixed in 8f715a7f: both now import from a new shared `pam/shared/otp.ts` (also deduped against `phone-login.service.ts`'s copy), and `nodeHeadersToHeaders` now imports the existing `pam/shared/headers-mapper.ts` instead of reimplementing it.

Left `assertFreshReauthentication`'s TOTP step-up as-is for now. It's real duplication in spirit, but unifying it with `IdentityService.assertFreshSecondFactor` means bridging two different `Auth.api` type surfaces across two separate classes (the narrow `VerifyTotpApi` workaround type here vs. the full `Auth['api']` there) plus their own lockout/error-message conventions — a bigger cross-class refactor than fits a review-comment pass on an already-large PR. Happy to pull it into a dedicated follow-up if you'd rather not carry the duplication.

packages/core/src/contracts/schemas/identity.ts
damianrzepka: Accurate, but leaving this as-is for this PR. The single boolean is intentional here — it's what's in the reviewed cross-repo contract with the betfeel Security page this PR was built against, not an oversight. Splitting it into two preferences is a real improvement but it's a breaking `SecurityControlsSchema`/`SetLoginWithdrawalAlertsInputSchema` change plus a data migration, and needs the same cross-repo coordination the current shape went through — worth its own follow-up rather than folding into this one.

PR #139, opened 2026-09-04
fix(chat-commands): survive command config written before per-currency amounts

zaxovaiko:
## Summary

`GET /chat-command/commands` returned 500 on any database seeded before amounts became per-currency, which killed the whole slash-command dropdown in chat. A data migration clears the unreadable values and the row mapper stops trusting the jsonb column's type cast. BF-545

## Why

`CommandConfigSchema.minAmount` / `maxAmount` moved from a bare decimal string to a currency-keyed record. The schema migration shipped; the data migration did not, and `seedChatCommands` uses `onConflictDoNothing`, so `gift`, `rain` and `donate` rows kept `{"minAmount":"1.00000000"}`. Three stale rows failed output validation and the route 500ed for every player.

The column type made it invisible: `config: jsonb().$type<CommandConfig>()` is a cast Postgres never enforced, so TypeScript believed a shape the database had no obligation to hold. The route also had no test, and a fresh-database test could not have caught this - CI seeds the new shape and gets it back.

Reproduced end to end against a clone of a database carrying the old rows:

| code | migration applied | result |
| --- | --- | --- |
| before | no | 500 `Output validation failed` |
| before | yes | 200 |
| after | no | 200, stale config omitted |
| after | yes | 200 |

## Alternatives considered

**Convert the scalar to `{ "USD": value }`.** Rejected: the old value never recorded a currency, so any mapping invents a money limit no operator set. Both fields are optional, so dropping the key is lossless in contract terms and an admin re-sets it through the backoffice update route. These three commands are operator-owned since the transfer surface left core, so core holding a guessed amount is worse than holding none.

**Delete the offending rows and let the seed rewrite them.** Fixes one environment by hand and re-breaks on the next database carrying old data.

**Migration only, no read-side guard.** Fixes this shape but leaves the next contract change able to 500 the same route.

## Risks

Any `minAmount` / `maxAmount` still in the pre-change shape is dropped, not converted, so an operator running those commands re-enters the limit once after deploy. Only rows whose stored value is a JSON string are touched; a value already in the new record shape is left alone.

The read-side guard degrades silently: an unreadable config now yields `null` and an unknown command key drops out of the list rather than raising. `adminListCommands` still counts filtered rows in its `total`, so a page could come back short if a database holds a key this build does not know. No such key exists today.

### DIFF
--- a/packages/core/src/engagement/chat-commands/__tests__/migration-upgrade.int.test.ts
+++ b/packages/core/src/engagement/chat-commands/__tests__/migration-upgrade.int.test.ts
@@ -2,6 +2,7 @@ import { Pool } from 'pg';
 import { describe, expect, it } from 'vitest';
 import { createTestDb } from '@openora/core/testing';
 import { migrate } from '../migrate.js';
+import { CommandConfigSchema, type CommandConfig } from '../contract/index.js';
 
 const STRIPED_TRAUMA_HASH = 'a58ec63c81a4cbd62603a8381da61c5c6d85b57af5d56ba79839d090fbcf0708';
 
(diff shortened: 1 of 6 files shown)


zaxovaiko:
Superseded by #141, which landed the same root cause on `dev`: a row still holding an older `minAmount`/`maxAmount` shape failed the descriptor schema and 500ed the whole command list.

The two took different calls on the same data. This branch dropped the flat amount, on the grounds that it recorded no currency and any currency chosen for it would be invented. #141 converts it to `{currency: 'USD', amount}`, naming USD as the currency the seed used when that shape was written. That is the better call - it keeps the operator's limit instead of clearing it, and it sources the currency rather than guessing it.

Nothing here is worth salvaging: #141 carries its own migration-upgrade coverage for the legacy shapes, and the route-level test this branch added is covered by `jsonb-contract-drift.e2e.test.ts` in #140. Both branches also add a `0002` to the same migration journal, so they cannot coexist.

Closing.

PR #142, opened 2026-09-06
feat(casino): manage game catalog via providers and categories

okapitula:
## Summary

Replaces free-text provider/category strings on game with first-class game_provider, game_category and link entities, served through GamingService reads and backoffice management routes.

## Why

Operators need stable provider/category references to curate the catalog; strings allow drift and per-game N+1 lookups in gaming and lobby reads.

## Alternatives considered

Keep free-text fields with validation at write time — rejected, no stable identity for curation or filtering.

## Risks

- game.is_active now defaults to false for new rows; seed data updated, existing rows migrated.
- GameSchema/GameSummarySchema shape changes to provider summaries and category arrays.


packages/core/src/casino/gaming/drizzle/migrations/0003_omniscient_the_professor.sql
```diff
@@ -0,0 +1,46 @@
...
+	"is_active" boolean DEFAULT false NOT NULL,
+	"metadata" jsonb,
+	"updated_at" timestamp with time zone NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE "game" ALTER COLUMN "is_active" SET DEFAULT false;--> statement-breakpoint
+ALTER TABLE "game" ADD COLUMN "slug" text NOT NULL;--> statement-breakpoint
+ALTER TABLE "game" ADD COLUMN "provider_id" uuid NOT NULL;--> statement-breakpoint
+ALTER TABLE "game" ADD COLUMN "aggregator" text NOT NULL;--> statement-breakpoint
```
mp-blurify: These three columns are `NOT NULL` with no default and no backfill, so the migration aborts on any database that already has `game` rows (`column "slug" of relation "game" contains null values`). `provider_id` needs more than a default — it carries an FK into `game_provider`, which is empty at this point, so it needs an explicit backfill: create providers from the old `game.provider` strings, map them onto the games, then `SET NOT NULL`.

`0002_yummy_lifeguard.sql:2` is the pattern — it added `game_type` with `DEFAULT 'casino'`.

The description says "existing rows migrated", but there is no `UPDATE` or `INSERT` anywhere in the diff. CI won't catch this because it migrates an empty database.

packages/core/src/casino/gaming/drizzle/migrations/0003_omniscient_the_professor.sql
```diff
@@ -0,0 +1,46 @@
...
+CREATE UNIQUE INDEX "game_category_game_key" ON "game_category_game" USING btree ("game_id","category_id");--> statement-breakpoint
+CREATE INDEX "game_category_game_category_id_idx" ON "game_category_game" USING btree ("category_id");--> statement-breakpoint
+CREATE UNIQUE INDEX "game_provider_slug_key" ON "game_provider" USING btree ("slug");--> statement-breakpoint
+CREATE UNIQUE INDEX "game_provider_aggregator_vendor_id_key" ON "game_provider" USING btree ("aggregator_vendor_id");--> statement-breakpoint
+ALTER TABLE "game" ADD CONSTRAINT "game_provider_id_game_provider_id_fk" FOREIGN KEY ("provider_id") REFERENCES "public"."game_provider"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
+CREATE UNIQUE INDEX "game_slug_key" ON "game" USING btree ("slug");--> statement-breakpoint
+CREATE INDEX "game_provider_id_idx" ON "game" USING btree ("provider_id");--> statement-breakpoint
+CREATE INDEX "game_aggregator_idx" ON "game" USING btree ("aggregator");--> statement-breakpoint
+ALTER TABLE "game" DROP COLUMN "provider";--> statement-breakpoint
+ALTER TABLE "game" DROP COLUMN "category";
```
mp-blurify: Dropping these in the same release as the reader change breaks instances still running the previous version. Expand first, contract later: land the new shape here, migrate the readers, drop `provider`/`category` in a follow-up release.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -352,6 +352,30 @@ export const domainEventSchemas = {
     playerId: UuidSchema.nullable(),
   }),
 
+  // Backoffice game-catalog management (actorId = acting admin UUID).
+  'gaming.provider.updated': authContextBase.extend({
```
mp-blurify, line 379: These four topics are declared and emitted, but none of them is in `SUBSCRIBED_TOPICS` in `packages/core/src/audit/plugin.ts` — only `gaming.round.started`/`ended` are there. Nothing subscribes, so no backoffice catalog mutation reaches `audit_log`.

`mapEventToRecord` has no branch for them either, so even after adding them to the list they'd fall through to the base mapping and land as `actorType: 'system'` with `resourceId: null`, despite `actorId` being in the payload.

`docs/standards/audit.md`: every mutation that changes configuration state needs an audit record with actor, resource, outcome and before/after.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -352,6 +352,30 @@ export const domainEventSchemas = {
...
+  'gaming.category.updated': authContextBase.extend({
+    categoryId: UuidSchema,
+    actorId: UuidSchema.optional(),
+    before: z.object({ slug: z.string(), name: z.string() }),
+    after: z.object({ slug: z.string(), name: z.string() }),
+  }),
+  'gaming.game.updated': authContextBase.extend({
+    gameId: UuidSchema,
+    actorId: UuidSchema.optional(),
+  }),
```
mp-blurify, line 452: `gaming.game.updated` carries no before/after. Flipping `isActive` takes a game live or dark, and this leaves no record of what actually changed.

The provider and category events do carry before/after, but only `slug` and `name` — `isActive` is missing there too, and that's the field deciding whether a whole portfolio is visible.

packages/core/src/casino/gaming/service/game-provider.service.ts
```diff
@@ -0,0 +1,151 @@
...
+      updated = findOneOrThrow(
+        await this.drizzle.db
+          .update(gameProvider)
+          .set(patch)
+          .where(eq(gameProvider.id, id))
+          .returning(),
+        new GameProviderNotFoundError(id),
+      );
+    } catch (error) {
+      if (isUniqueConstraintViolation(error)) {
```
mp-blurify: `game_provider` has two unique indexes — `game_provider_slug_key` and `game_provider_aggregator_vendor_id_key` (`schema/index.ts:41-42`) — but every 23505 maps to `GameProviderSlugTakenError`. Give two providers the same `aggregatorVendorId` and the admin gets a 409 about a slug conflict that doesn't exist. Worth discriminating on `error.constraint`, plus a pre-check for `aggregatorVendorId` like the one for slug above.

Same shape in `gaming.service.ts:343`: that transaction also inserts into `game_category_game`, so a concurrent edit hitting `game_category_game_key` surfaces as "slug taken" as well.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -60,21 +104,73 @@ export class GamingService {
...
+    providerId,
+    categoryId,
+    isActive,
+  }: {
+    page: number;
+    limit: number;
+    q?: string;
+    providerId?: GameProvider['id'];
+    categoryId?: GameCategory['id'];
+    isActive?: boolean;
```
mp-blurify, line 143: The join pulls in `gameProvider` but the filter never touches `gameProvider.isActive`. Same in `lobby.service.ts`, in both `search` and `getCategoryGames`.

So deactivating a provider doesn't remove its games from the public catalog or from search, while `listActiveProviders` stops returning it — games then show a provider that isn't in the filter list. If provider `isActive` is meant as a label rather than a kill switch, worth saying so on the column, because the name reads the other way.

packages/core/src/casino/gaming/schema/index.ts
```diff
@@ -19,17 +21,86 @@ export const gameRoundStatusEnum = pgEnum('game_round_status', GAME_ROUND_STATUS
...
+    id: uuid().primaryKey().defaultRandom(),
+    name: text().notNull(),
+    slug: text().notNull(),
+    providerId: uuid()
+      .notNull()
+      .references(() => gameProvider.id),
+    aggregator: text().notNull(),
+    gameType: gameTypeEnum().notNull().default('casino'),
+    thumbnailUrl: text(),
+    isActive: boolean().notNull().default(false),
```
mp-blurify, line 102: With `is_active` defaulting to `false`, every newly ingested game starts uncurated — which is the point of this PR. But `startRound` calls `getGame` only as an existence check, and `getGame` has no `isActive` filter, so a player who knows the id can start a round and be debited on a game that was never made live.

The behaviour predates this PR, but the default flip turns it from an edge case into the normal state of a fresh row. `getGame` has to stay unfiltered since `updateGame` uses it for its return value, so the check belongs in `startRound`.

packages/core/src/server/db/query-helpers.ts
```diff
@@ -99,6 +109,22 @@ function fromUnitsAtScale(units: bigint, scale: number): string {
...
+  return value.replace(/[\\%_]/g, '\\$&');
+}
+
+// `%value%` contains-match for ilike/like with caller input.
+export function likeContains(value: string): string {
+  return `%${escapeLike(value)}%`;
+}
+
+// `value%` prefix-match for ilike/like with caller input.
+export function likePrefix(value: string): string {
```
mp-blurify, line 140: `likePrefix` has no callers — `audit/service/audit.service.ts:132` still defines its own private copy. Either point that one at this import or drop the export.

mp-blurify (commented):
Reviewed at `e0518a9c`. Green locally: `pnpm lint`, `pnpm check:boundaries` (0 violations), `pnpm check:drift` (catalog in sync), `tsc --noEmit`, and the `query-helpers` + `kit` unit tests (52/52). Integration tests need Postgres and weren't run.

The modelling is the right call and the seams are clean — `adminGuard.assert` is the first line of every backoffice handler and the router test table walks all of them for denial, `listGames` pins `isActive: true` in the router so it can't be overridden from input, `categoriesByGameIds` genuinely kills the N+1 and is shared by gaming and lobby, and `escapeLike`/`likeContains` closes a real hole in `lobby.search`.

Three things I'd want fixed before merge, left inline:

1. Migration `0003` can't run on a database that already has `game` rows — three `NOT NULL` columns with no default and no backfill. CI won't catch it, it migrates from empty. The description says "existing rows migrated"; nothing in the diff does that.
2. Same migration drops `provider` and `category` in the same release as the reader change. Expand first, contract later.
3. The four new `gaming.*` catalog events are emitted but never subscribed — they aren't in `SUBSCRIBED_TOPICS` in `audit/plugin.ts`, so no backoffice catalog mutation reaches `audit_log`. `gaming.game.updated` also carries no before/after, so an `isActive` flip leaves no trace of what changed.

Smaller ones inline: 23505 is mapped to a slug error even when `aggregatorVendorId` is the constraint that fired; provider `isActive` isn't filtered in the public catalog or in lobby search; `startRound` still accepts an inactive game, which matters more now that `is_active` defaults to `false`; `likePrefix` is exported with no callers.

Two notes not worth an inline comment:

- `listGames` changes from an array to `paginated(...)`. Risks mentions the `GameSchema`/`GameSummarySchema` change but not this one — it's a separate break for anyone on `@openora/core/react`, and the JSDoc example at `react/client.ts:18` (`await client.gaming.listGames()`) is now wrong on both input and output.
- The `.gitignore` and `rulesync.jsonc` opencode entries are unrelated to the catalog work — nicer as a separate commit.

packages/core/src/casino/gaming/drizzle/migrations/0003_omniscient_the_professor.sql
okapitula: Fixed in 48b149e36155fe110d359969a0a31f8e7795d528 0003 is now expand-only (nullable adds, legacy columns kept) and 0004 backfills providers/categories/links/slugs from legacy strings with collision-safe suffixes; generated 0005 enforces NOT NULL. Covered by gaming-migration.int.test.ts, which replays 0000-0002, seeds legacy rows, and asserts each stage.

packages/core/src/casino/gaming/drizzle/migrations/0003_omniscient_the_professor.sql
okapitula: Fixed in 48b149e36155fe110d359969a0a31f8e7795d528. 0003 no longer drops anything — legacy provider/category stay (nullable) for old releases.

packages/core/src/contracts/schemas/events.ts
okapitula: Fixed in 8d9e5c177f4ae0c5801c242501b290cbba1832a9. All four topics are in SUBSCRIBED_TOPICS with dedicated mapEventToRecord branches (admin actor, game_provider/game_category/game resources, before/after passthrough).

packages/core/src/contracts/schemas/events.ts
okapitula: Fixed in 8d9e5c177f4ae0c5801c242501b290cbba1832a9. game.updated now carries full before/after snapshots (slug, name, providerId, aggregator, thumbnailUrl, isActive, categoryIds, metadata); provider/category snapshots gained aggregatorVendorId/logoUrl/icon/sortOrder/isActive. All four bumped to v2.

packages/core/src/casino/gaming/service/game-provider.service.ts
okapitula: Fixed in 48b149e36155fe110d359969a0a31f8e7795d528. Added `uniqueConstraintName()` and `GameProviderVendorIdTakenError` with its own pre-check and router mapping; catches translate only on matching constraint, unknown `23505`s propagate. `updateGame` maps only `game_slug_key` now — a link race no longer masquerades as a slug conflict.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Fixed in 9c89586150418d564acfb2397a221215a641ac19. Shared `playableGameCondition()` (game + provider active) now gates public `listGames`, lobby `search`/`getCategoryGames`, and `startRound`. Backoffice listings pass through untouched. Note `getCategoryGames` previously had no active filter at all, so it also leaked default-inactive games — now gated on both.

packages/core/src/casino/gaming/schema/index.ts
okapitula: Fixed in 9c89586150418d564acfb2397a221215a641ac19. `startRound` goes through `getGame(gameId, { activeOnly: true })` and 404s before any debit for missing, inactive, or provider-deactivated games. `getGame` itself stays unfiltered for `updateGame`'s return path, per your note.

packages/core/src/server/db/query-helpers.ts
okapitula: Fixed in 5e5d8bd3bd2d2b693435d473a4a7d961f770fe3a. Took the import option — `audit.service` now uses the shared `likePrefix` and the private copy is deleted.

packages/core/src/casino/gaming/drizzle/migrations/0003_natural_thunderbolt_ross.sql
zaxovaiko, line 1: Let's squash migrations into a single one. Or is there ny reason to keep them separated?

okapitula: Merged in c742d246e9b1fa18c296682346dc1276dee5ff92. Also changed migration number to 0005 (see 6663e1aebde26933a46ce9b791248b9ba8a92245) due to merge conflicts.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -84,4 +114,135 @@ export const gamingContract = {
...
+    .route({ method: 'GET', path: '/backoffice/gaming/providers' })
+    .input(CatalogFilterSchema)
+    .output(paginated(GameProviderDetailSchema)),
+
+  getAdminProvider: oc
+    .route({ method: 'GET', path: '/backoffice/gaming/providers/{id}' })
+    .input(IdInputSchema)
+    .output(GameProviderDetailSchema),
+
+  updateProvider: oc
```
mp-blurify, line 262: [blocker vs the requirement] `gamingAdminContract` exposes list/get/update for providers but no create - there is no route, input schema or service method that inserts a `game_provider` row. After this lands a provider can only appear via the 0005 backfill or the demo seed, so "create ... providers" is unmet; `game-config:create` was added to the permission catalog for `createCategory` alone. Add `createProvider` (`POST /backoffice/gaming/providers`) with a `CreateProviderInputSchema` and a `GameProviderService.createProvider` mirroring `createCategory`.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -73,21 +122,86 @@ export class GamingService {
...
+  }) {
+    const where = and(
+      q ? or(ilike(game.name, likeContains(q)), ilike(game.slug, likeContains(q))) : undefined,
+      providerId ? eq(game.providerId, providerId) : undefined,
+      isActive === true
```
mp-blurify: `listAdminGames` passes the admin's `isActive` straight through, and `isActive === true` selects `playableGameCondition()` - game **and** provider active. So an operator filtering the backoffice catalog by "active" never sees an active game whose provider is switched off, with no way to tell it exists. Only the `isActive`-omitted admin path has a test. Gate on `playableGameCondition()` for the public caller only (an explicit `playableOnly` flag from `listGames`' handler) and keep `eq(game.isActive, true)` for the admin filter.

packages/core/src/casino/lobby/service/lobby.service.ts
```diff
@@ -75,19 +92,33 @@ export class LobbyService {
...
-        .map(toGameSummary),
+        .filter(
+          (
+            g,
+          ): g is {
+            game: typeof game.$inferSelect;
+            provider: typeof gameProvider.$inferSelect;
+          } => g !== undefined,
+        )
+        .map((r) => toGameSummary({ ...r, categories: categories.get(r.game.id) ?? [] })),
```
mp-blurify, line 122: `search` and `getCategoryGames` were both moved onto `playableGameCondition()` in this PR, but the sibling `getFeatured` just below (:125-137) still selects promoted games with no `is_active` and no provider filter. Since `game.is_active` now defaults to `false`, an uncurated or deliberately deactivated game - or one under a switched-off provider - keeps showing in the lobby's featured strip for up to the 30s cache TTL. `game-catalog.ts:12-14` asks for a single owner of the enabled definition. Join `gameProvider`, add the condition, and drop slots whose game no longer resolves.

packages/core/src/casino/shared/game-catalog.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// One batched query for many games - never per-game lookups (no N+1).
+// Shared by gaming and lobby so the join + ordering has a single owner.
+export async function categoriesByGameIds(db: DrizzleDb, gameIds: Game['id'][]) {
+  if (gameIds.length === 0) {
+    return new Map<Game['id'], GameCategory[]>();
+  }
+  const rows = await db
+    .select({ gameId: gameCategoryGame.gameId, category: gameCategory })
+    .from(gameCategoryGame)
```
mp-blurify, line 52: `categoriesByGameIds` never filters `gameCategory.isActive`, so a category an operator deactivates still ships on every game in the public `/gaming/games`, `/gaming/games/{id}` and lobby payloads - while `/gaming/categories` (`listActiveCategories`) hides it. The helper feeds both the public and the admin read, so there is no seam between them today. Take an `activeOnly` argument and pass `true` from every public caller.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -84,4 +114,135 @@ export const gamingContract = {
...
+  aggregator: z.string().trim().min(1).max(64).optional(),
+  thumbnailUrl: z.string().trim().min(1).max(512).nullable().optional(),
+  isActive: z.boolean().optional(),
+  metadata: z.unknown().nullable().optional(),
+  categoryIds: z.array(UuidSchema).max(50).optional(),
+});
+export type UpdateGameInput = z.infer<typeof UpdateGameInputSchema>;
+
+export const gamingAdminContract = {
+  listAdminProviders: oc
```
mp-blurify, line 247: Nine new backoffice routes ship with one E2E (`gaming-category-translations.e2e.test.ts`), covering the four category routes. `listAdminProviders`, `getAdminProvider`, `updateProvider`, `updateGame` and `listAdminGames` have no `bootTestApp` coverage - the router integration test calls handlers directly, so the query-string coercion of `isActive`/`providerId`/`categoryId` and the real permission wiring on those five never run. `docs/standards/testing.md:9` makes that E2E the acceptance artifact. Extending the existing file with a provider list/get/patch and an `updateGame` + `listAdminGames?isActive=false` round-trip would close it.

packages/core/src/casino/gaming/schema/index.ts
```diff
@@ -21,17 +28,94 @@ export const gameRoundStatusEnum = pgEnum('game_round_status', GAME_ROUND_STATUS
...
+    metadata: jsonb(),
+    updatedAt: timestamp({ withTimezone: true })
+      .notNull()
+      .$onUpdateFn(() => new Date()),
+    createdAt: timestamp({ withTimezone: true }).notNull().defaultNow(),
+  },
+  (t) => [
+    uniqueIndex('game_provider_slug_key').on(t.slug),
+    uniqueIndex('game_provider_aggregator_vendor_id_key').on(t.aggregatorVendorId),
```
mp-blurify: `game_provider_aggregator_vendor_id_key` is unique across the whole table, but a vendor id is only unique inside one aggregator and the provider row carries no aggregator column (`aggregator` lives on `game`). With two aggregators integrated, the second one's studio id `42` collides with the first's and `updateProvider` returns a spurious `GameProviderVendorIdTakenError`; the same provider also cannot hold two aggregator mappings. Either store the aggregator beside the vendor id and make the index `(aggregator, aggregator_vendor_id)`, or move the mapping to its own table.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -810,6 +884,12 @@ export const domainEventVersions: Partial<Record<DomainEventName, number>> = {
   // v2: permanent renamed to isPermanent (non-predicate boolean naming rule).
   // v3: durationMonths added - the chosen term, explicit for the regulatory export.
   'rg.self_exclusion.activated': 4,
+  // v2: before/after carry the full config snapshot (isActive + all mutable fields)
+  // so visibility flips and renames are diffable in the audit log.
+  'gaming.provider.updated': 2,
```
mp-blurify: All four topics are introduced by this PR, yet they are registered in `domainEventVersions` as version 2. `getEventVersion` defaults to 1, which is why every other first-release topic is absent from that map. Consumers reading `eventCatalog()` or an outbox envelope will see `schemaVersion: 2` for a payload shape that never had a v1, breaking producer/consumer agreement checks against a version that does not exist (ADR-0017). Dropping the four entries starts them at the implicit v1.

rulesync.jsonc
```diff
@@ -32,6 +32,7 @@
     "copilot": ["rules", "mcp", "subagents", "commands", "skills", "hooks"],
     "codexcli": ["rules", "mcp", "subagents", "skills", "hooks", "permissions"],
     "antigravity-cli": ["rules", "mcp", "hooks", "skills"],
+    "opencode": ["rules", "mcp", "subagents", "commands", "skills", "hooks", "permissions"],
```
mp-blurify: This `opencode` target and the `.opencode` / `opencode.jsonc` entries at `.gitignore:42-43` are agent-tooling config with no connection to the game catalog, and they widen this PR into the repo's generated rule mirrors. `docs/standards/git-delivery.md:7` - one PR, one concern. Worth splitting into a `chore(repo)` PR.

mp-blurify:
One blocker against the requirement: providers can be listed, fetched and updated but never created, so after this lands a provider can only exist via the backfill or the seed. The rest is inline - an admin "active" filter that also hides games under deactivated providers, `getFeatured` left behind when its siblings moved onto the playable condition, disabled categories still shipping on public game payloads, a table-wide unique on a vendor id that is only unique per aggregator, five new routes without E2E, and four brand-new topics registered at schema version 2. The threads already open here all read as fixed on this HEAD.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Fixed in `64977b35`. Added `CreateProviderInputSchema` and `POST /backoffice/gaming/providers`, guarded by `game-config:create`. Provider and aggregator mappings are inserted transactionally, uniqueness violations map to typed conflicts, and `gaming.provider.created` supplies the audit snapshot. Covered by real-Postgres and `bootTestApp` tests.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Fixed in `64977b35`. Public playability is now controlled by a separate `playableOnly` flag and requires both the game and provider to be active. The backoffice `isActive` filter applies only to `game.isActive`, so active games under inactive providers remain visible to operators. Added regression coverage for both paths.

packages/core/src/casino/lobby/service/lobby.service.ts
okapitula: Fixed in `64977b35`. `getFeatured` now joins `game_provider`, applies the shared `playableGameCondition()`, and removes slots whose game or provider is inactive. Added real-Postgres coverage for both cases.

packages/core/src/casino/shared/game-catalog.ts
okapitula: Fixed in `64977b35`. `categoriesByGameIds` now supports active-only reads, enabled by all public gaming and lobby callers while admin reads retain inactive assignments. Public filtering by an inactive `categoryId` also returns no games. Both public and admin behaviors have regression coverage.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Fixed in `64977b35`. Extended the `bootTestApp` suite to cover provider create/list/get/update, game update and filtered admin listing, query-string coercion, and permission wiring. It also verifies that an authenticated player without `game-config:create` cannot create a provider.

packages/core/src/casino/gaming/schema/index.ts
okapitula: Fixed in `64977b35`. Replaced the scalar vendor ID with `game_provider_aggregator_mapping`. Migration `0005` now creates the final table directly and enforces one mapping per provider and aggregator plus uniqueness of each `(aggregator, vendor_id)` pair. A provider can therefore be mapped through multiple aggregators without cross-aggregator vendor-ID collisions. Supplying mappings on update uses explicit replacement semantics; persisted mappings are read back for the response and audit snapshot.

packages/core/src/contracts/schemas/events.ts
okapitula: Fixed in `64977b35`. Removed the newly introduced gaming topics from `domainEventVersions`, so they start at the implicit schema version 1 in accordance with ADR-0017.

rulesync.jsonc
okapitula: Addressed in `0bff6918`. The unrelated OpenCode/rulesync and `.gitignore` changes were removed from this feature branch, leaving this PR scoped to the game catalog work.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
```diff
@@ -0,0 +1,137 @@
...
+FROM (
+	SELECT
+		"b"."provider" AS "provider",
+		"b"."base" AS "base",
+		ROW_NUMBER() OVER (PARTITION BY "b"."base" ORDER BY "b"."provider") AS "rn"
+	FROM (
+		SELECT DISTINCT
+			COALESCE("legacy"."provider", 'Unknown') AS "provider",
+			COALESCE(
+				NULLIF(left(lower(regexp_replace(regexp_replace(COALESCE("legacy"."provider", 'Unknown'), '[^a-zA-Z0-9]+', '-', 'g'), '(^-+|-+$)', '', 'g')), 60), ''),
```
mp-blurify: **Blocker.** `left(..., 60)` is the outermost call, applied *after* the `(^-+|-+$)` trim, so any provider, category or game name longer than 60 characters can backfill a slug that ends in `-`. Such a row fails `CatalogSlugSchema` (`contract/index.ts:97`), so `GET /gaming/providers/{slug}` and `GET /gaming/categories/{slug}` return 400 for it and it can never be addressed by slug. Same expression at `:98` and `:119`.

Fix: truncate first, then trim — `regexp_replace(left(lower(regexp_replace(...)), 60), '(^-+|-+$)', '', 'g')`.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
```diff
@@ -0,0 +1,137 @@
...
+INSERT INTO "game_category_game" ("game_id", "category_id")
+SELECT "g"."id", "c"."id" FROM "game" "g"
+JOIN "game_category" "c" ON "c"."name" = COALESCE("g"."category", 'Uncategorized');--> statement-breakpoint
+UPDATE "game" "g"
+SET "slug" = "s"."slug", "provider_id" = "p"."id", "aggregator" = 'direct'
+FROM (
+	SELECT
+		"b"."id" AS "id",
+		"b"."provider" AS "provider",
+		CASE WHEN "b"."rn" = 1 THEN "b"."base" ELSE "b"."base" || '-' || "b"."rn" END AS "slug"
```
mp-blurify: **Blocker — this aborts on a customer database.** The dedupe suffix `base || '-' || rn` produces a string that is itself a legal base: two rows named `Book of Ra` yield `book-of-ra` and `book-of-ra-2`, and a third row named `Book of Ra 2` independently yields `book-of-ra-2`. `game_slug_key` (created at `:53`, before this UPDATE) then raises 23505 and the whole migration rolls back. Same shape at `:65` and `:86`.

Fix: suffix with `left(id::text, 8)` instead of a sequential `rn`, or loop until the candidate slug is unclaimed.

packages/core/src/casino/lobby/service/lobby.service.ts
```diff
@@ -75,19 +92,33 @@ export class LobbyService {
...
     const gameIds = links.map((l) => l.gameId);
 
-    const games =
-      gameIds.length > 0 ? await db.select().from(game).where(inArray(game.id, gameIds)) : [];
+    const rows =
+      gameIds.length > 0
+        ? await db
+            .select({ game, provider: gameProvider })
+            .from(game)
+            .innerJoin(gameProvider, eq(game.providerId, gameProvider.id))
```
mp-blurify, line 101: (Anchored on the changed query — the defect is the unchanged `gameCount` at `:74`.)

`gameCount` still counts every `lobby_category_game` link, but `getCategoryGames` now drops games whose game or provider is inactive. Before this PR the two agreed because `getCategoryGames` applied no filter at all; now the lobby nav advertises "Slots (12)" and the category page renders 8.

Fix: join `game`/`game_provider` into the count query and apply `playableGameCondition()`.

packages/core/src/casino/lobby/contract/index.ts
```diff
@@ -1,12 +1,17 @@
...
+} from '@openora/core/contracts';
 
 export const GameSummarySchema = z.object({
   id: UuidSchema,
   name: z.string(),
-  provider: z.string(),
-  category: z.string(),
+  slug: z.string(),
+  provider: GameProviderSummarySchema,
+  categories: z.array(GameCategorySummarySchema),
```
mp-blurify: Lobby game summaries expose `GameCategorySummarySchema`, which has no `translations`, while `/gaming/games` returns `GameCategorySummaryWithTranslationsSchema`. The country-keyed translations added in this PR exist for localized display, and the lobby (`search`, `getCategoryGames`) is exactly the player-facing surface that needs them — so a localized client gets English-only category names there.

Fix: use `GameCategorySummaryWithTranslationsSchema` here and carry `c.translations` through `toGameSummary` (`lobby.service.ts:39`).

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -283,4 +402,121 @@ export class GamingService {
...
+    const [beforeRow] = await this.drizzle.db.select().from(game).where(eq(game.id, id)).limit(1);
+    if (!beforeRow) {
+      throw new GameNotFoundError(id);
+    }
+    const beforeLinks = await this.drizzle.db
+      .select({ categoryId: gameCategoryGame.categoryId })
+      .from(gameCategoryGame)
+      .where(eq(gameCategoryGame.gameId, id));
+    const beforeCategoryIds = beforeLinks.map((r) => r.categoryId);
+    if (patchInput.providerId !== undefined) {
```
mp-blurify, line 444: `updateGame` validates `providerId` and `categoryIds` against their tables but accepts any `aggregator` string, and nothing checks that the chosen provider has a `game_provider_aggregator_mapping` row for it. An operator can save `providerId: evolution, aggregator: 'everymatrix'` when Evolution is only mapped on `softswiss`, leaving a catalog row whose vendor id is unresolvable — `seed-demo-data.ts:507` already ships exactly that state (every seeded game claims `everymatrix`, no mapping rows are created).

Fix: look up `(providerId, aggregator)` in `game_provider_aggregator_mapping` and throw a typed 404/409 when it is missing.

packages/core/src/casino/gaming/schema/index.ts
```diff
@@ -21,17 +28,110 @@ export const gameRoundStatusEnum = pgEnum('game_round_status', GAME_ROUND_STATUS
...
-});
+export const gameProvider = pgTable(
+  'game_provider',
+  {
+    id: uuid().primaryKey().defaultRandom(),
+    slug: text().notNull(),
+    name: text().notNull(),
+    logoUrl: text(),
+    isActive: boolean().notNull().default(false),
+    metadata: jsonb(),
```
mp-blurify, line 39: `game_provider.metadata` is added by this PR and never read, never written, and absent from `GameProviderDetailSchema`, `CreateProviderInputSchema` and `UpdateProviderInputSchema` — there is no path that can put a value in it or get one out.

Rule: `docs/standards/database.md` — no columns the code does not use.
Fix: drop the column from the table and the migration, or expose it on the provider write schemas.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -84,4 +118,176 @@ export const gamingContract = {
     .output(EndRoundOutputSchema),
 
   listRounds: oc.route({ method: 'GET', path: '/gaming/rounds' }).output(z.array(GameRoundSchema)),
+
+  listProviders: oc
```
mp-blurify, line 122: `listProviders`, `getProviderBySlug` and `getCategoryBySlug` are new routes with no `bootTestApp` E2E — only the in-process router test (`gaming-catalog-admin.router.int.test.ts:172-189`). The E2E suite reaches `/gaming/categories` (`gaming-category-translations.e2e.test.ts:89`) and every backoffice route, but never these three over real HTTP.

Rule: `docs/standards/testing.md` — a new or changed route gets one E2E; it is the acceptance artifact for the route.
Fix: add the three requests to `gaming-category-translations.e2e.test.ts`, including the inactive-provider 404.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
okapitula: Confirmed and fixed in e1bb4525. Slugification now truncates first and trims the edge hyphens afterwards, so a name longer than the limit can no longer leave a trailing dash, and the expression moved into a single helper shared by all three call sites. Verified on a scratch database: a name whose 60th slug character is a hyphen now yields a 59-character slug that matches the kebab pattern.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
okapitula: Confirmed, and it does abort. Reproduced on a scratch database with three rows named "Book of Ra", "Book of Ra!" and "Book of Ra 2": the old backfill fails with 23505 on `game_provider_slug_key` for `novomatic-2`. Fixed in e1bb4525 — all three generators now probe the table for an unclaimed candidate instead of assigning a suffix blind, which also covers the fallback slugs. Same fixture on the new SQL yields `book-of-ra`, `book-of-ra-2` and `book-of-ra-2-2`, all unique and all valid.

packages/core/src/casino/lobby/service/lobby.service.ts
okapitula: Correct, fixed in e1bb4525. The count query now joins game and provider and applies `playableGameCondition()`, so the badge and the category page are computed from the same set. Added a regression test asserting the listed count equals the number of games the feed returns for a category holding one playable game and two hidden ones.

packages/core/src/casino/lobby/contract/index.ts
okapitula: Agreed, fixed in e1bb4525. Lobby game summaries now use `GameCategorySummaryWithTranslationsSchema` and the field flows through the feed. Rather than adding a third copy of the category mapper I moved it into the shared catalog module that already owns the playable-game definition, so gaming and lobby cannot drift on it again. Added a test asserting a German translation reaches the lobby feed.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Fixed in e1bb4525. `updateGame` now checks the resulting pair against `game_provider_aggregator_mapping` and raises a typed conflict, surfaced as 409. It validates the pair as a whole rather than the patched half, since changing only the provider can strand a previously valid aggregator. Two follow-ons the change required: the demo seed creates the matching mapping rows, and the migration backfills a `direct` mapping per provider so migrated games are not born in a state the new rule rejects.

packages/core/src/casino/gaming/schema/index.ts
okapitula: Keeping the column and exposing it rather than dropping it — done in e1bb4525. It is now on `GameProviderDetailSchema`, on both the create and update inputs, on the service read and write paths, on both provider events, and in the audit snapshot. Added a test covering create, replace, omit and clear.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Added in e1bb4525. The E2E suite now exercises the provider listing and both by-slug routes, including the inactive-provider 404 before activation and the 200 after it, plus the public category by slug. One existing assertion changed as a side effect: the game patch previously claimed `direct` against a provider mapped only on `everymatrix`, which the new pairing check rejects, so that request now asserts 409 and a second request uses the mapped aggregator.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
```diff
@@ -0,0 +1,171 @@
...
+			"aggregator" = 'direct'
+		WHERE "id" = source."id";
+	END LOOP;
+END
+$do$;--> statement-breakpoint
+INSERT INTO "game_provider_aggregator_mapping" ("provider_id", "aggregator", "vendor_id")
+SELECT "p"."id", 'direct', "p"."slug" FROM "game_provider" "p"
+ON CONFLICT DO NOTHING;--> statement-breakpoint
+DROP FUNCTION pg_temp.openora_slugify(text, int);--> statement-breakpoint
+ALTER TABLE "game" ALTER COLUMN "slug" SET NOT NULL;--> statement-breakpoint
```
jakubfilinger-b, line 221: **Blocker:** this is still not expand-only for a rolling deploy. Once `slug`, `provider_id`, and `aggregator` become `NOT NULL`, an old application instance that inserts the legacy shape cannot populate them and every new game insert fails. Keeping `provider`/`category` only preserves old reads. Keep the new columns nullable (or add a compatibility trigger/default path) until all old writers are retired, then enforce `NOT NULL` in a later release.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -283,4 +417,138 @@ export class GamingService {
...
+    userAgent,
+    ...patchInput
+  }: UpdateGameInput & Actor) {
+    const uniqueCategoryIds = categoryIds === undefined ? undefined : [...new Set(categoryIds)];
+    const patch: Partial<typeof game.$inferInsert> = { ...patchInput };
+    const hasScalarChanges = Object.values(patch).some((value) => value !== undefined);
+    if (!hasScalarChanges && uniqueCategoryIds === undefined) {
+      return this.getGame(id);
+    }
+    const [beforeRow] = await this.drizzle.db.select().from(game).where(eq(game.id, id)).limit(1);
```
jakubfilinger-b: **Blocker:** the audit snapshot is raceable. `beforeRow` and category links are read outside the write transaction, the in-transaction read is not locked, and `afterRow` is read after commit. Two concurrent PATCHes can both emit the same stale `before`, or one event can claim categories written by the other request. `updateCategory` has the same split read/write at `game-category.service.ts:165`. Move the locked before-read, validation, write, and persisted after snapshot into one transaction, then emit after commit (`docs/standards/audit.md`).

packages/core/src/casino/gaming/service/game-provider.service.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+        const next = findOneOrThrow(
+          await tx
+            .update(gameProvider)
+            .set(replaceMappings ? { ...patch, updatedAt: new Date() } : patch)
+            .where(eq(gameProvider.id, id))
+            .returning(),
+          new GameProviderNotFoundError(id),
+        );
+        if (replaceMappings) {
```
jakubfilinger-b, line 303: Replacing mappings can delete a `(providerId, aggregator)` pair still referenced by active games. The PATCH succeeds, those games remain public/playable, and the resulting state is one that `updateGame` itself rejects as unmapped. Reject removal of referenced mappings inside this transaction (or enforce the pair with a same-module composite FK and handle the conflict explicitly).

packages/core/src/contracts/schemas/events.ts
```diff
@@ -364,6 +365,91 @@ export const domainEventSchemas = {
...
+  // Backoffice game-catalog management (actorId = acting admin UUID).
+  'gaming.provider.created': authContextBase.extend({
+    providerId: UuidSchema,
+    slug: z.string(),
+    name: z.string(),
+    aggregatorMappings: z.array(GameProviderAggregatorMappingSchema),
+    logoUrl: z.string().nullable(),
+    metadata: z.unknown().nullable(),
+    isActive: z.boolean(),
+    actorId: UuidSchema.optional(),
```
jakubfilinger-b: These five events represent admin-only catalog mutations, but every `actorId` is optional, so the event contract permits an audit record with `actorType: admin` and no actor. The router always has a required `AdminGuard` user id; reflect that invariant here by requiring `UuidSchema` and making the service actor required (`docs/standards/audit.md`, `docs/standards/types.md`).

jakubfilinger-b (changes requested):
CHANGES REQUESTED — the migration is not compatible with old writers during a rolling deploy, and catalog mutations can produce an inaccurate regulatory audit trail.

packages/core/src/casino/gaming/drizzle/migrations/0005_natural_thunderbolt_ross.sql
okapitula: Fixed in 0f4d8bff. I went with the compatibility-trigger path instead of nullable columns, so `slug` / `provider_id` / `aggregator` stay non-null for new code.

- `game_legacy_insert` (BEFORE INSERT) derives `slug`, `provider_id` and `aggregator` for a legacy-shape insert the same way the backfill does, and ensures the `direct` mapping exists.
- `game_legacy_category_link` (AFTER INSERT) links the legacy `category` text.
- Their `WHEN` clauses keep both off the new-code insert path. `NOT NULL` is only enforced after they're installed.

Covered in `gaming-migration.int.test.ts`: a legacy insert succeeds and resolves, and a new-shape insert is untouched. The triggers and their functions get dropped together with the legacy columns in the contract migration. This covers old writers' inserts; legacy updates to the `provider`/`category` text aren't mirrored.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Fixed in 0f4d8bff. `updateGame` now does all of this in one transaction:
- the before-read with `FOR UPDATE`
- the provider/mapping/slug/category validation
- the write
- the after snapshot (row plus category links)

The event is emitted after commit. When the provider or aggregator changes, it also takes `FOR SHARE` on the provider row, so it serializes against `updateProvider` replacing mappings.

`updateCategory` follows the same pattern: locked read, slug check, update, and both snapshots inside the transaction, emitted after commit.

packages/core/src/casino/gaming/service/game-provider.service.ts
okapitula: Fixed in 0f4d8bff. `updateProvider` now rejects removing an aggregator that any of the provider's games still reference, with `GameProviderMappingInUseError` mapped to `CONFLICT`.

I check every referencing game, not just active ones. A retired game on a dropped aggregator would hit the same unmapped state as soon as it's reactivated.

The check runs inside the transaction under the provider's `FOR UPDATE`. `updateGame` takes `FOR SHARE` on the provider when it moves a game, so a game can't land on the pair mid-check. Test added in `game-provider.service.int.test.ts`.

packages/core/src/contracts/schemas/events.ts
okapitula: Fixed in 0f4d8bff. `actorId` is now a required `UuidSchema` on all five catalog events (`gaming.provider.created/updated`, `gaming.category.created/updated`, `gaming.game.updated`). It's also required in the three services' `Actor` input type, so a service call without an actor no longer compiles.

packages/core/src/contracts/schemas/game.ts
```diff
@@ -1,5 +1,49 @@
...
+export type GameProviderAggregatorMapping = z.infer<typeof GameProviderAggregatorMappingSchema>;
+
+export const GameCategoryNameSchema = z.string().trim().min(1).max(128);
+export const GameCategoryTranslationSchema = z
+  .object({
+    name: GameCategoryNameSchema,
+  })
+  .strict();
+export const GameCategoryTranslationsSchema = z.record(
+  CountryCodeSchema,
```
zaxovaiko: Category translations are keyed by ISO 3166-1 country code, but they carry a **language** string. A country is not a language: Belgium (`BE`) speaks Dutch, French and German, Switzerland (`CH`) four languages, Canada (`CA`) two, so each of those can hold exactly one category name and the rest of its players read the untranslated `name`. The reverse is just as lossy: German needs a separate entry under `DE`, `AT` and `CH` with the same string.

The platform already carries the right key. `user.language` is a BCP 47 tag validated by `LanguageSchema` (`contracts/schemas/identity.ts:11`), which is exactly what a client would use to select an entry: `translations[user.language]`.

Suggestion:

```ts
export const GameCategoryTranslationsSchema = z.record(
  LanguageSchema.regex(/^[a-zA-Z]{2,8}(?:-[a-zA-Z0-9]{2,8})*$/),
  GameCategoryTranslationSchema,
);
```

No migration needed, the keys are jsonb data. The regex keeps the write path validated, since `LanguageSchema` on its own is `z.string().max(35)` and would accept an empty key.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -84,4 +118,179 @@ export const gamingContract = {
...
+    .route({ method: 'GET', path: '/gaming/providers' })
+    .output(z.array(GameProviderSummarySchema)),
```
zaxovaiko: `listProviders` and `listCategories` return unbounded arrays while every sibling catalog read in this PR is paginated: `/gaming/games` (`:106`), `/backoffice/gaming/providers`, `/backoffice/gaming/categories`, `/backoffice/gaming/games`. A player-facing route that grows with the catalog and has no page bound is the one shape that cannot be tightened later without breaking clients twice.

Since `listGames` already breaks the array shape in this PR, doing the same to these two now costs consumers nothing extra:

```ts
listProviders: oc
  .route({ method: 'GET', path: '/gaming/providers' })
  .input(PageQuerySchema)
  .output(paginated(GameProviderSummarySchema)),
```

Same for `listCategories`, with `listActiveProviders`/`listActiveCategories` taking `{ page, limit }` and running the list plus a `count()` the way `listProvidersAdmin` already does.

zaxovaiko (commented):
Two follow-ups on the catalog surface, both outside the blocker set already addressed in 7287c4fc.

zaxovaiko (dismissed):
Let's fix pagination and lang key and let's merge it.

packages/core/src/contracts/schemas/game.ts
okapitula: Fixed in 1bd72c2d. `GameCategoryTranslationsSchema` is now keyed by `LanguageSchema` plus the BCP 47 tag regex you suggested, so a client can pick `translations[user.language]`. No migration, since the column is new in this PR.

The schema tests now accept `de`, `pt-BR` and `zh-Hant-TW`, and reject `''`, `d`, `de_DE`, `de-` and `-de`. The other tests use `de`/`fr` keys instead of `DE`/`FR`.

Keys are matched exactly, so `de` and `DE` are different entries. Clients should look them up with the same casing as `user.language`.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Fixed in 1bd72c2d. `listProviders` and `listCategories` now take `PageQuerySchema` and return `paginated(...)`, like the other catalog reads.

`listActiveProviders` and `listActiveCategories` take `{ page, limit }` and run the page query plus a `count()`, the same way `listProvidersAdmin` does. Each has a slug tiebreak so pages stay stable when names or sort orders tie. The service, router and e2e tests now check the paginated shape, including a second page.

`/lobby/categories` in the lobby module still returns a plain array. It's already on `dev` and outside this PR, so I've left it for a follow-up.

PR #143, opened 2026-09-07
fix(server): let httpCache.additionalPaths extend the default cache list

klaudia-blazyczek-blurify:
## Summary

`createApp`'s `httpCache` config gains `additionalPaths`, which merges with the built-in
`PUBLIC_HTTP_CACHE_PATHS` list instead of replacing it; the list itself is now exported.

## Why

`httpCache.paths` replaces `PUBLIC_HTTP_CACHE_PATHS` wholesale, and that default list wasn't
exported. A consumer wanting to cache one path of their own on top of the defaults had no way to
reference the default list, so adding a path meant either hand-copying it (drifts silently if the
default list ever changes) or accepting that every other default path quietly loses its caching.

## Alternatives considered

- Exporting `PUBLIC_HTTP_CACHE_PATHS` alone and leaving consumers to spread it into `paths`
  themselves - still relies on remembering to spread it in; the same silent-loss bug is one typo
  away. Still done here alongside `additionalPaths`, for a consumer who wants to build their own
  list from scratch.
- Making `paths` merge with the defaults instead of replacing them - a breaking change for any
  consumer already relying on `paths` to fully override the default list.

## Risks

None. `additionalPaths` is a new, optional field; `paths` keeps its existing full-override
behavior and takes precedence when both are set.

### DIFF
--- a/packages/core/src/server/runtime/create-app.ts
+++ b/packages/core/src/server/runtime/create-app.ts
@@ -52,12 +52,14 @@ import type { CoreTokenCatalog } from './core-token-catalog.js';
 // Path prefixes safe to cache at the HTTP layer: public, non-personalized reads
 // only (lobby feeds, public CMS content, the game catalogue). NOTHING
 // authenticated or per-player (wallet, profile, notifications, admin, chat) - a
-// consumer overrides this via `httpCache.paths` or disables it with `httpCache: false`.
+// consumer replaces this wholesale via `httpCache.paths`, extends it via
+// `httpCache.additionalPaths`, or disables caching entirely with `httpCache: false`.
 // This list living in the domain-agnostic engine (rather than each module
-// declaring its own cacheable routes) is accepted debt.
+// declaring its own cacheable routes) is accepted debt. Exported so a consumer
+// using `paths` to add one prefix can spread this in instead of hand-copying it.
 // '/cms/banners' is excluded on purpose: its `listBanners` route is unguarded and returns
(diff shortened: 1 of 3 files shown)


packages/core/src/server/runtime/create-app.ts
```diff
@@ -52,12 +52,14 @@ import type { CoreTokenCatalog } from './core-token-catalog.js';
...
+// consumer replaces this wholesale via `httpCache.paths`, extends it via
+// `httpCache.additionalPaths`, or disables caching entirely with `httpCache: false`.
 // This list living in the domain-agnostic engine (rather than each module
-// declaring its own cacheable routes) is accepted debt.
+// declaring its own cacheable routes) is accepted debt. Exported so a consumer
+// using `paths` to add one prefix can spread this in instead of hand-copying it.
 // '/cms/banners' is excluded on purpose: its `listBanners` route is unguarded and returns
 // inactive banners, so HTTP-caching it would leak draft banners to a shared cache.
-const PUBLIC_HTTP_CACHE_PATHS: readonly string[] = [
+export const PUBLIC_HTTP_CACHE_PATHS: readonly string[] = [
```
jakubfilinger-b: Could we keep this as `as const` rather than typing it as `readonly string[]`? It is now part of the public API, and preserving the literal values lets consumers derive the exact default-path union if they need it.

klaudia-blazyczek-blurify: Sure. I will change that

zaxovaiko (approved):
Good for now

PR #144, opened 2026-09-07
feat(identity): email and SMS as second-factor delivery methods

jakubfilinger-b:
Lets an account take its second factor over email or SMS, not just an authenticator app.

## What lands

- **`email` and `sms` as delivery methods.** A new `two_factor_method` column names the enrolled method; `TwoFactorDeliveryService` decides where a generated code goes, which is the one thing better-auth's twoFactor plugin cannot know - it sees a single enrolment and no notion of `app` vs `email` vs `sms`. Codes route through `MAIL_DISPATCH` (new `twoFactorOtp` template) or the bound `SmsAdapter`, so core never names a delivery vendor. The migration backfills every existing enrolment to `app`, which is what they all were.
- **The withheld-login response names the method.** A client that gets `twoFactorRedirect` can now render the right challenge - and request a code - instead of assuming an authenticator.
- **`otp` joins `AUTHENTICATION_METHODS`.** A delivered code completes a sign-in exactly as `totp` does.

## Also fixed here

`invalidateSecurityControls` refreshed `me` and `security.me` but never `twoFactorStatus`, which is its own query and the one the Security page reads. A user who finished enrolment kept seeing the pre-enrolment controls until an unrelated refetch happened to run.


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1237,7 +1285,147 @@ export class IdentityService {
...
+      maskedPhone: row.phoneVerified && row.phoneNumber ? maskPhone(row.phoneNumber) : null,
+    };
+  }
+
+  /**
+   * Rejects a method the account cannot actually receive a code on. Enrolling into
+   * `sms` without a verified phone would mint a secret the player can never clear,
+   * locking them out of their own account at the next login.
+   */
+  private async assertMethodAvailable(
```
mp-blurify, line 1388: The reasoning in the docblock applies to `email` just as much as to `sms`. `registration.requireEmailVerification` is off by default, so an account with an unverified (or mistyped) address can enrol into `email` and lock itself out at the next login — exactly the failure this check exists to prevent. `user.emailVerified` is already on the table.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1237,7 +1285,147 @@ export class IdentityService {
     await ensureOk(res);
     this.forwardCookies(res, resHeaders);
     const body = (await res.json()) as { totpURI: string; backupCodes: string[] };
-    return { totpUri: body.totpURI, backupCodes: body.backupCodes };
+
+    await this.setTwoFactorMethod(userId, input.method);
```
mp-blurify, line 1317: "Abandoning this step leaves the account exactly as it was" holds for an unenrolled account, but not for one that already has 2FA on: better-auth deletes and recreates the `twoFactor` row while `twoFactorEnabled` stays true. An abandoned method switch therefore leaves `enabled=true` with the new method and a new secret. An account on `email` that starts and abandons a switch to `app` gets a TOTP challenge at the next login against a secret it never scanned, and `twoFactorStatus` reports the unconfirmed method as active because it gates on `enabled` rather than on the first code having cleared.

Separately: switching the method on a live second factor only needs session + password today, while `disable2fa` needs a fresh code. Worth putting it behind `assertFreshSecondFactor` too.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1237,7 +1285,147 @@ export class IdentityService {
...
+    const masked = await this.sendOtpForCurrentMethod(userId, headers, resHeaders);
+    return { backupCodes: body.backupCodes, maskedDestination: masked };
+  }
+
+  /**
+   * Re-sends the one-time code for an account on the `email` or `sms` method - the
+   * "Resend code" action during enrolment, and the send leg of a login challenge.
+   * Resolves its subject from a live session first and the pending-challenge cookie
+   * second, because at login there is no session yet.
+   */
```
mp-blurify, line 1334: `verifyTwoFactor` calls `twoFactorLockout.assertNotLocked` before spending a code; this one doesn't. An account inside a 2FA lockout window can still trigger sends at the full rate-limit budget.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1237,7 +1285,147 @@ export class IdentityService {
...
+    headers: Headers,
+    resHeaders: Headers,
+  ): Promise<string> {
+    const destination = await this.twoFactorDelivery.describeDestination(userId);
+    if (!destination) {
+      throw new ORPCError('BAD_REQUEST', {
+        message: 'This account does not receive two-factor codes by email or SMS',
+      });
+    }
+    const res = await this.api.sendTwoFactorOTP({ body: {}, headers, asResponse: true });
```
mp-blurify, line 1468: `ensureOk` here doesn't mean the code was delivered. better-auth runs the `sendOTP` hook through `runInBackgroundOrAwait(...).catch(logger.error)`, so a throw from `SmsAdapter.sendOtp` or `mailDispatch.toUser` is swallowed and the endpoint still returns 200. We then hand the client a `maskedDestination` as though the code is on its way. With a real SMS vendor that's a player waiting on a code that never arrives, with nothing surfaced through the API.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1411,11 +1597,20 @@ export class IdentityService {
...
     await this.twoFactorLockout?.assertNotLocked(userId);
-    const res = await this.api.verifyTOTP({
-      body: { code, trustDevice: false },
+    // The code has to be spent by the endpoint that issued it: an account on the
+    // `email`/`sms` method has no authenticator to read a TOTP off, so verifying one
+    // against the shared secret would reject every code it was actually sent.
+    const [row] = await this.drizzle.db
+      .select({ twoFactorMethod: user.twoFactorMethod })
+      .from(user)
+      .where(eq(user.id, userId))
```
mp-blurify: This is the same select as `resolveTwoFactorMethod` above — call it instead. `twoFactorStatus` and `disableTwoFactor` each re-read the same row too, and the `enable2fa` path for `email` hits `user` four times (assertMethodAvailable, setTwoFactorMethod, describeDestination, deliver).

packages/core/src/pam/identity/service/two-factor-delivery.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    // the platform's per-user locale, rendering and delivery log like every other mail.
+    // The key is fresh per send: a resent code is a new mail, not a retry of the old one.
+    await this.mailDispatch?.toUser({
+      userId: args.userId,
+      template: { key: 'twoFactorOtp', data: { otp: args.code } },
+      idempotencyKey: `otp:twoFactorOtp:${randomUUID()}`,
+    });
+  }
+
+  private async findAccount(userId: User['id']) {
```
mp-blurify, line 154: `phoneVerified` and `language` are selected but never read. `language` on `deliver` is never populated either — the `sendOTP` hook in `server/auth/auth.ts` passes only `userId`, `email`, `phoneNumber` and `code`.

packages/core/src/pam/identity/react/account.ts
```diff
@@ -71,6 +76,19 @@ export function useVerify2fa() {
   });
 }
 
+// The Security page reads this rather than `me`: the active method and the masked
+// destinations are 2FA state, not profile fields, and must refresh whenever enrolment
+// changes without dragging the whole user object along.
+export function useTwoFactorStatus() {
```
mp-blurify: Both new hooks are published through `pam/react.ts` with inferred return types. `docs/standards/react-sdk.md` asks for an explicit annotation on an SDK export, the way `useMySecurityControls` and `useSetLoginWithdrawalAlerts` do it just below.

packages/core/src/pam/identity/drizzle/migrations/0011_worried_nuke.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE "public"."two_factor_method" AS ENUM('app', 'email', 'sms');--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "two_factor_method" "two_factor_method";--> statement-breakpoint
+-- Every enrolment predating this column is an authenticator one: `app` was the only
+-- method the platform could enrol into. Left null, those accounts would read as
+-- enabled with no active method, and their step-up codes would route to the wrong
+-- verify endpoint.
+UPDATE "user" SET "two_factor_method" = 'app' WHERE "two_factor_enabled" = true;
```
mp-blurify: The backfill is hand-written into a drizzle-generated migration, which `docs/standards/database.md` rules out. The convention elsewhere is a separate, descriptively named file — `chat/0007_seed_global_chat_room.sql`, `chat/0010_backfill_role_assigned_at.sql`, `chat-commands/0002_single_currency_command_limits.sql`. The DDL itself is fine (nullable column, no DROP or RENAME). File is also missing a trailing newline.

packages/core/src/contracts/schemas/identity.ts
```diff
@@ -132,22 +132,36 @@ export const UsernameAvailabilityInputSchema = z.object({ username: UsernameSche
...
 export const Enable2faInputSchema = z.object({
   password: z.string().min(8),
+  method: TwoFactorDeliveryMethodSchema.default('app'),
 });
 
 export const Enable2faResultSchema = z.object({
-  totpUri: z.string().min(1),
+  // Only the authenticator method has a URI to render as a QR code; `email` and `sms`
+  // get their code pushed instead, so there is nothing for the client to display.
+  totpUri: z.string().min(1).optional(),
```
mp-blurify, line 150: Making `totpUri` optional is a breaking change for anyone consuming `@openora/core` — `result.totpUri` becomes `string | undefined` downstream. Nothing in the PR records it.

mp-blurify (commented):
Four things that don't anchor to a changed line, then the verdict.

**`trustCurrentDevice` still forces TOTP** — `identity.service.ts:1552`. Same problem you fixed in `assertFreshSecondFactor`, second occurrence missed. For an `email`/`sms` account the code the player holds is checked against the shared TOTP secret and always fails. Routing it to `verifyTwoFactorOTP` alone won't fix it either: the method replays a fresh sign-in leg, and `sendTwoFactorOTP` keys the verification value by that pending cookie, so no OTP exists for the new challenge key. Every attempt also burns a `twoFactorLockout.recordFailure`, and the two events emitted below it hardcode `method: 'totp'`.

**`PhoneVerificationService` still forces TOTP** — `phone-verification.service.ts:148`. Same thing in `assertFreshReauthentication`, so a player on `email`/`sms` can't pass `phoneVerification.request` or `confirm`, and the copy ("An authenticator code is required.") is neither true nor satisfiable for them. For an `sms` account that closes the loop: they can no longer rotate the number their own second factor is delivered to. `git grep -n verifyTOTP` shows three production call sites; this PR converts one.

**`pnpm check:drift` is red.** `docs/catalog.json` and `docs/platform/system-design.md` are stale (254→256 routes, `pam` 61→63, `TwoFactorDeliveryMethodSchema`/`TwoFactorStatusSchema` missing from the schema registry). Verified the same check passes on `dev`, so the drift is from this branch. Needs `pnpm regen`.

**No E2E for the two new routes.** `docs/standards/testing.md` makes one E2E per new or changed route the acceptance artifact. `POST /identity/2fa/otp/send` and `GET /identity/2fa/status` are new, and the `enable2fa`, `verify2fa` and `login` payloads changed; the only new test covers `maskEmail`/`maskPhone`. The delivery routing in `deliver()`, `describeDestination`, the verify-endpoint choice and the `sms`-without-verified-phone gate are all uncovered — the existing integration tests were only adjusted to keep compiling.

Everything else looks right. `twoFactorMethod` is only revealed after `ensureOk` on the sign-in, so the login response doesn't enumerate. The migration DDL is safe (nullable column, no DROP or RENAME) and the backfill is correct. Resend works — better-auth's `consumeVerificationValue` sorts newest-first and clears the rest. `pnpm check:types` is green across all 10 packages and the new unit test passes.

NO-GO for now: two step-up paths reject every code an `email`/`sms` account can obtain, drift fails CI, and neither new route has an acceptance test.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1237,7 +1308,146 @@ export class IdentityService {
     await ensureOk(res);
     this.forwardCookies(res, resHeaders);
     const body = (await res.json()) as { totpURI: string; backupCodes: string[] };
-    return { totpUri: body.totpURI, backupCodes: body.backupCodes };
+
+    await this.setTwoFactorMethod(userId, input.method);
```
damianrzepka, line 1317: Changing an already enabled factor is not recoverable if delivery fails. `enableTwoFactor` has already replaced the better-auth secret and persists the new method before it calls `sendOtpForCurrentMethod`. Better-auth catches a `sendOTP` rejection and still returns 200, so this path reports a masked destination although no code was delivered. The player is then left with a new unreachable factor and cannot complete the normal six-digit step-up flow to repair it. Keep the active factor/method until delivery and activation succeed, or provide an explicit recovery path, and cover an adapter rejection.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1361,13 +1569,14 @@ export class IdentityService {
...
 
-    // Only a live authenticator earns the trust window - a backup code is a recovery
+    // A live second factor earns the trust window - a backup code is a recovery
     // credential and the schema never lets one reach this route.
-    const verified = await this.api.verifyTOTP({
-      body: { code: input.code, trustDevice: true },
-      headers: challengeHeaders,
-      asResponse: true,
-    });
+    const challengeMethod = await resolveChallengeMethod(this.drizzle, userId);
```
damianrzepka, line 1609: The replayed sign-in creates a fresh pending 2FA cookie, but the OTP path verifies it without first sending an OTP for that new challenge. Better-auth stores OTPs under the pending-cookie key, so an enrolment or earlier login code belongs to a different key and this verification always expires for email/SMS accounts. It also records a lockout failure. Send the OTP with `challengeHeaders` before verifying when `challengeMethod` is `otp`, and add a real-flow test.

damianrzepka, line 1609: `challengeMethod` is also the source of truth for the successful authentication event below. Carry it through to `identity.authentication.succeeded` instead of hard-coding `totp`; otherwise every email/SMS device-trust login is misclassified for security-alert and audit consumers. Please assert the OTP event payload.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed - `assertMethodAvailable` now checks `emailVerified` for the `email` method the same way it already checked `phoneVerified` for `sms`.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed differently than a fresh-code gate: re-enrolling on an account that already has 2FA enabled is now refused outright (409). Right after your comment I confirmed the reason is worse than a stale-method drift - better-auth's enable leg `deleteMany`s + recreates the `twoFactor` row on every call, so re-running it in place would destroy the working secret and backup codes the instant the call lands, not just on abandonment. A player switches method by disabling first, matching the ticket's AC (no direct switch-in-place).

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed - the send leg now calls `twoFactorLockout.assertNotLocked` before minting a code, same as `verifyTwoFactor`.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed - `sendOtpForCurrentMethod` now reads back `TwoFactorDeliveryService.takeFailure()` after the send and throws 503 if the transport actually rejected it, instead of trusting `ensureOk`. Covered by a real vendor-rejection e2e test (SMS_ADAPTER overridden to reject) in `packages/testing`.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed - pulled the method lookup into a shared `resolveChallengeMethod` (`two-factor-challenge.service.ts`) that `assertFreshSecondFactor`, `trustCurrentDevice`, `disableTwoFactor`'s audit read, and (after the dev rebase) the shared `assertFreshReauthentication` all call instead of re-querying `user`.

packages/core/src/pam/identity/service/two-factor-delivery.service.ts
jakubfilinger-b: Fixed - `language` dropped (it was never populated anyway), `phoneVerified` now actually gates the SMS send and `describeDestination`'s answer, matching the enrolment-time check.

packages/core/src/pam/identity/react/account.ts
jakubfilinger-b: Fixed - both hooks now carry an explicit `UseTwoFactorStatusResult`/`UseSendTwoFactorOtpResult` return type.

packages/core/src/pam/identity/drizzle/migrations/0011_worried_nuke.sql
jakubfilinger-b: Fixed - split into a DDL migration and a standalone, descriptively named backfill (now 0012/0013 after rebasing behind dev's own new 0011). Trailing newline added.

packages/core/src/contracts/schemas/identity.ts
jakubfilinger-b: Acknowledged and documented - added a changeset flagging `Enable2faResult.totpUri` becoming optional as a breaking change for consumers that render it unconditionally.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: This referred to the version before the CONFLICT guard landed (91808c8a) - re-enrolling on a live account is now refused outright, so this exact failure mode (a live factor left half-changed) can't occur any more. What's left for a genuine FIRST-enrolment delivery failure: `sendOtpForCurrentMethod` now reads back a real transport failure and throws 503 (see the reply on the adapter-rejection thread), and the persisted `two_factor_method` is recoverable either way - `twoFactorStatus` only reports it once `enabled` flips true, and both a resend and a fresh `enable2fa` call read the same column back. Documented the recovery path on the docblock and added an e2e covering a real SMS vendor rejection end to end.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Confirmed and fixed differently than routing the OTP verify call: `trustCurrentDevice` now refuses up front (409) for any account not on `app`, before it ever mints the fresh challenge or spends a verify attempt. You're right that just switching the verify call to `verifyTwoFactorOTP` wouldn't have worked - the fresh pending-cookie key this route mints has no code sent against it, so it would always expire. Since a pushed code can never answer a challenge this route creates for itself, refusing is the only correct behavior, not a routing fix. Added a real e2e test driving this against the actual better-auth stack with a computed TOTP code (nothing in the suite exercised this route for real before), plus a negative case proving the email path is refused with no lockout attempt spent.

jakubfilinger-b: The trust-device success event now carries `method: challengeMethod` (dynamic) instead of a literal. The sign-in-succeeded event right below it is a special case worth calling out: since the CONFLICT guard means only `totp` can ever reach that line, the literal `'totp'` there is actually correct, not stale - I left it as a literal rather than `challengeMethod` to make that invariant visible at the call site. Added an assertion on the emitted event's method in the new real-flow test.

damianrzepka (approved):

PR #148, opened 2026-09-08
feat(lobby): operator-defined lobby sections with a replaceable layout

okapitula:
## Summary

Operators can now compose the casino lobby from typed, toggleable sections
and replace the whole layout in one versioned PUT. Public layout serving
resolves each section through a replaceable `LobbySectionCatalog` with a
30s cache; backoffice get/replace routes are `AdminGuard`-protected.

This replaces the previous static lobby schema: migration `0001`
creates `lobby_layout` / `lobby_section` and drops the unused
`featured_slot`, `lobby_category`, and `lobby_category_game` tables,
which were never wired to any service, route, or seed.

## Why

Operators needed a nice and flexible way to manage what players see in
the casino lobby — curating, ordering, and toggling content themselves
instead of asking for a code change and deploy every time.

## Alternatives considered

- Let the operator handle it entirely on their side using existing
  routes/resources: rejected — pushes ordering, enable flags,
  validation, caching, and concurrent-edit safety onto every operator
  as boilerplate, with inconsistent results.
- Ship predefined sections inside openora core: rejected — too rigid;
  every operator wants different sections, and core would accumulate
  one-off variants that don't generalize.

## Risks

- New migration (`0001_misty_sentinel`) creates `lobby_layout` /
  `lobby_section` and `DROP TABLE ... CASCADE`s the three unused tables
  above. Safe for fresh installs; dev DBs reset via `pnpm setup`. No
  data migration — the dropped tables held no production data (never
  seeded or written by any service).
- Public layout cached 30s (`LOBBY_CACHE_TTL_MS`) — admin edits take up
  to 30s to surface; invalidated on replace.
- Optimistic concurrency on `version` — concurrent admin edits get
  `LobbyLayoutVersionConflictError`, client must refetch and retry.
- Unknown/invalid/unresolved section types are skipped with a warn log,
  never a 500 — a misconfigured catalog entry degrades silently in logs.

## Testing

- `pnpm verify` green (build, types, lint, format, boundaries, shape,
  hygiene, unit + integration, drift check).
- New: `lobby-contract`, `lobby-section-catalog`, `lobby-layout.int`,
  `lobby.cache.int`, `lobby-layout.e2e` (save → serve → authz).


packages/core/src/casino/lobby/drizzle/migrations/0001_misty_sentinel.sql
```diff
@@ -0,0 +1,23 @@
...
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"type" text NOT NULL,
+	"config" jsonb NOT NULL,
+	"sort_order" integer DEFAULT 0 NOT NULL,
+	"is_enabled" boolean DEFAULT true NOT NULL,
+	"updated_at" timestamp with time zone NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+DROP TABLE "featured_slot" CASCADE;--> statement-breakpoint
```
jakubfilinger-b: This drops tables that the previous application version still reads. During a rolling deployment, old instances will start failing as soon as this migration runs, and `CASCADE` may also remove consumer-owned dependencies. Our migration rules require expand-first/contract-later: please keep these tables in this release and remove them only after every old reader has been retired.

packages/core/src/casino/lobby/router/index.ts
```diff
@@ -1,22 +1,38 @@
...
-  return os.router({
-    listCategories: os.listCategories.handler(() => lobby.listCategories()),
+export function createLobbyRouter({
+  lobby,
+  adminGuard,
+}: {
+  lobby: LobbyService;
+  adminGuard: AdminGuard;
+}) {
+  const os = implement({ ...lobbyContract, ...lobbyAdminContract }).$context<OssContext>();
```
jakubfilinger-b: The runtime router combines `lobbyContract` and `lobbyAdminContract`, but downstream consumers compose only the exported `lobbyContract`. As a result, these admin routes work over HTTP but are missing from the canonical typed client. Please expose one aggregate lobby contract containing both public and admin routes, and have the router implement that exact object.

packages/core/src/casino/lobby/contract/index.ts
```diff
@@ -1,56 +1,66 @@
...
-  gameName: z.string(),
-  thumbnailUrl: z.string().nullable(),
-  placement: z.string(),
-  sortOrder: z.number(),
+  type: LobbySectionTypeSchema,
+  data: LobbySectionDataSchema,
 });
+export type LobbyResolvedSection = z.infer<typeof LobbyResolvedSectionSchema>;
 
 export const lobbyContract = {
```
jakubfilinger-b, line 84: This removes four published routes and their response schemas. Existing typed consumers will stop compiling, while already deployed clients will receive 404s for the old endpoints. Please retain the deprecated routes for a compatibility window, or deliver this as an explicitly coordinated breaking release with a migration path.

packages/core/src/casino/lobby/service/lobby.service.ts
```diff
@@ -1,133 +1,392 @@
...
+
+      if (sectionData === undefined) {
+        logger.warn(
+          { type, sectionId: section.id },
+          'Skipping lobby section that was not resolved',
+        );
+        continue;
+      }
+
+      resolved.set(section.id, { id: section.id, type, data: sectionData });
```
jakubfilinger-b: Resolver output crosses a consumer-provided adapter boundary but is not validated at runtime. For example, a type-correct resolver can return `{ score: NaN }`; `LobbySectionDataSchema` rejects it and oRPC turns the entire `GET /lobby/layout` request into a 500. This contradicts the promised graceful degradation. Please validate each value with `LobbySectionDataSchema.safeParse()` and skip/log invalid sections.

packages/core/src/casino/lobby/service/lobby.service.ts
```diff
@@ -1,133 +1,392 @@
...
+        try {
+          const validation = await definition.validate?.(
+            group.map((section) => ({ id: section.id, config: section.config })),
+          );
+          if (validation && !validation.valid) {
+            throw new LobbySectionFieldError(
+              `Invalid '${type}' section configuration: ${validation.message}`,
+            );
+          }
+        } catch (error) {
```
jakubfilinger-b, line 395: This converts every exception from `definition.validate()` into a client-side 400. A timeout or implementation failure would therefore be reported as invalid configuration and hide a server fault. Please return 400 only for an explicit `{ valid: false }` result; unexpected exceptions should be logged and rethrown, as required by `docs/standards/errors.md`.

packages/core/src/server/runtime/create-app.ts
```diff
@@ -59,13 +59,7 @@ import type { CoreTokenCatalog } from './core-token-catalog.js';
...
 // '/cms/banners' is excluded on purpose: its `listBanners` route is unguarded and returns
 // inactive banners, so HTTP-caching it would leak draft banners to a shared cache.
-export const PUBLIC_HTTP_CACHE_PATHS = [
-  '/lobby/categories',
-  '/lobby/featured',
-  '/lobby/search',
-  '/cms/pages',
-  '/gaming/games',
-] as const;
+export const PUBLIC_HTTP_CACHE_PATHS = ['/lobby/layout', '/cms/pages', '/gaming/games'] as const;
```
jakubfilinger-b: Adding this route to the default HTTP cache means clients and CDNs receive `max-age=30, stale-while-revalidate=60`. They can therefore serve the previous layout for up to 90 seconds, and the Redis invalidation cannot purge those copies. This does not match the PR’s stated maximum of 30 seconds. Please remove the extra HTTP caching layer for this route or align its cache policy with the promised behavior.

jakubfilinger-b (changes requested):
NO-GO: the PR currently has four merge-blocking issues: a destructive same-release migration, an incompatible public API removal, admin routes missing from the canonical typed contract, and unvalidated resolver output.

packages/core/src/casino/lobby/drizzle/migrations/0001_misty_sentinel.sql
okapitula: Addressed in [f82fee5](https://github.com/blurifycom/openora/commit/f82fee5b992f2c4f3a4641d01a47e088c264eb9f): rebuilt the unmerged lobby migration as expand-only. It creates the layout tables and no longer drops the legacy tables.

packages/core/src/casino/lobby/router/index.ts
okapitula: Addressed in [f82fee5](https://github.com/blurifycom/openora/commit/f82fee5b992f2c4f3a4641d01a47e088c264eb9f): `lobbyContract` now contains both public and admin routes, and the router implements that exact contract.

packages/core/src/casino/lobby/contract/index.ts
okapitula: Addressed in [f82fee5](https://github.com/blurifycom/openora/commit/f82fee5b992f2c4f3a4641d01a47e088c264eb9f): restored the legacy category, featured, and search routes, schemas, service behavior, and backing tables for the compatibility window.

packages/core/src/casino/lobby/service/lobby.service.ts
okapitula: Addressed in [5dc62bc](https://github.com/blurifycom/openora/commit/5dc62bc617ec050e686b7c482f7d97eabfbc895b): resolver data is validated with `LobbySectionDataSchema.safeParse()`; invalid sections are logged and skipped.

packages/core/src/casino/lobby/service/lobby.service.ts
okapitula: Addressed in [5dc62bc](https://github.com/blurifycom/openora/commit/5dc62bc617ec050e686b7c482f7d97eabfbc895b): explicit `{ valid: false }` results still map to 400, while unexpected validator exceptions are logged and rethrown.

packages/core/src/server/runtime/create-app.ts
okapitula: Addressed in [5dc62bc](https://github.com/blurifycom/openora/commit/5dc62bc617ec050e686b7c482f7d97eabfbc895b): removed `/lobby/layout` from the default HTTP cache, so HTTP/CDN responses are `no-store` and the 30-second service cache is the only freshness layer.

marek-chmielowski-blurify (dismissed):

mp-blurify:
## Code review — findings

### Blocking
None. The six merge-blocking issues from the previous round were re-verified against the current code (not against the "addressed in commit X" replies) and all six are genuinely fixed with regression tests: destructive migration, admin routes missing from the typed contract, unvalidated resolver output, validator exceptions mis-mapped to 400, double HTTP caching past the promised 30s, AdminGuard usage.

### Worth fixing
- **`docs/adapters/`** — no narrative doc for the new `lobby-section-catalog` extension port (only `custody.md`, `error-tracking.md`, `kyc.md`, `mail.md`, `payment.md` exist). It is discoverable via the catalog and the test plugin, but this is the PR's headline extension seam and deserves the same write-up the other adapter ports get.
- **`.gitignore:42-43`** — adds `.opencode` / `opencode.jsonc`, unrelated to lobby sections; looks like a stray local-tooling change carried in by a merge/rebase. Worth dropping from this diff.

### Note
`LobbySectionDataSchema` / `LobbySectionConfigSchema` being `z.json()` is a deliberate design choice (core cannot know consumer section shapes), not a validation gap — it still catches the non-serializable cases (NaN etc.) that motivated the original finding.

jakubfilinger-b (dismissed):

PR #149, opened 2026-09-08
feat(identity): add a player-set anti-phishing code for outgoing email

damianrzepka:
## Summary

- Lets a player set a free-text, case-sensitive anti-phishing recognition phrase in Security settings via `identity.security.setAntiPhishingCode` (no reauth, set/overwrite only) and read it back raw through `security.me`.
- Widens `EmailTemplateRenderer.render()` with a 4th optional `antiPhishingCode` param, threaded from `MailService.deliver()` via `AdminUserDirectory`, so `DefaultEmailTemplateRenderer` appends it as a footer line on every outgoing platform email for a recipient who set one.
- Registers the new `identity.security.anti_phishing_code.set` domain event in the audit pipeline.

Openora's half of the paired BF-520 build (Jira: BF-520 "Anti-phishing code"). Betfeel's frontend (Security settings modal) and branded-email renderer overlay consume this contract; their side lands separately.

## Test plan

- [x] `pnpm verify` (typecheck, lint, unit + integration tests) green
- [x] `pnpm regen` run, `docs/catalog.json`/`docs/platform/system-design.md` committed, `check:drift` clean
- [x] New E2E: `packages/testing/src/__tests__/anti-phishing-code.e2e.test.ts` (happy path + round-trip via `security.me`, overwrite, whitespace-only rejection, non-player 403)
- [x] Identity integration test: role check, no-reauth write, code absent from the emitted event payload
- [x] Mail tests: renderer footer present/absent, HTML-escaping of the code, 4th `render()` arg threading


packages/core/src/contracts/schemas/identity.ts
```diff
@@ -193,10 +193,18 @@ export const SecurityControlsSchema = z.object({
...
+  // a recognition signal readable only by the authenticated player it belongs to, same gate as
+  // the rest of this schema.
+  antiPhishingCode: z.string().nullable(),
 });
 
 export const SetLoginWithdrawalAlertsInputSchema = z.object({ enabled: z.boolean() });
 
+// Non-empty after trimming incidental leading/trailing whitespace (eg from copy-paste),
+// otherwise fully unrestricted and case-sensitive - no reauth, set/overwrite only.
+export const SetAntiPhishingCodeInputSchema = z.object({ code: z.string().trim().min(1) });
```
jakubfilinger-b: `code` has no upper bound. Every other free-text field in this file is capped (`name` `.max(255)`, `slug` `.max(100)`, `PasswordSchema` `.max(128)`), and this value goes into a `text` column and then gets appended verbatim to the body of every outgoing email for the recipient. Can we add `.max(255)` so a player can't inflate all their transactional mail with a huge string?

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1645,6 +1646,57 @@ export class IdentityService {
...
+        userAgent,
+      });
+      throw new ORPCError('FORBIDDEN', {
+        message: 'Only players can set this preference.',
+      });
+    }
+    const before = await this.securityControlsFor(userId);
+    await this.drizzle.db
+      .update(user)
+      .set({ antiPhishingCode: input.code, antiPhishingCodeSetAt: new Date() })
```
jakubfilinger-b: This mirrors `setLoginWithdrawalAlerts` everywhere except the unchanged-value short-circuit it has at line 1631. Without it, re-submitting the same code still does a write and emits `identity.security.anti_phishing_code.set` every time, so we pile up redundant audit rows. Worth matching the sibling and returning `before` early when the code hasn't changed.

packages/core/src/pam/identity/schema/index.ts
```diff
@@ -48,6 +48,11 @@ export const user = pgTable(
     // `withdrawalPinHash !== null`, no separate boolean column.
     withdrawalPinHash: text(),
     withdrawalPinSetAt: timestamp({ withTimezone: true }),
+    // Player-set, free-text, case-sensitive recognition phrase stamped into every outgoing
+    // platform email so the player can tell a genuine email from a phishing attempt. Plaintext
+    // (not hashed) - read back verbatim both to the owning player and into outgoing email.
+    antiPhishingCode: text(),
+    antiPhishingCodeSetAt: timestamp({ withTimezone: true }),
```
jakubfilinger-b: `antiPhishingCodeSetAt` is written but never read anywhere — not in `getSecurityControls`, not in the event payload, not in audit. Either wire it up somewhere or drop it until we need it; if it's deliberately there for later, a one-line comment would help.

packages/core/src/contracts/adapters/admin-user-directory.ts
```diff
@@ -21,6 +21,10 @@ export type AdminUserRow = {
   language: string | null;
   failedLoginAttempts?: number;
   lockoutUntil?: Date | null;
+  // The player's anti-phishing code, if set - already returned to its owner via
+  // security.me, so reusing this port to thread it into outgoing mail is fine; it is not
+  // treated as secret. Null for an account that never set one.
+  antiPhishingCode: string | null;
```
jakubfilinger-b: Two small things: the neighbouring auth fields (`failedLoginAttempts?`, `lockoutUntil?`) are optional but this one is required, which is a minor tightening for any external `AdminUserDirectory` impl. Also `toAdminUser` in `backoffice.service.ts` spreads the whole row via `serializeRow`, so the only thing keeping this value out of the admin API response is `AdminUserSchema` stripping unknown keys on output — that's a bit load-bearing. Maybe don't populate it on rows headed for the directory list path, or leave a comment at the schema.

jakubfilinger-b (commented):
Overall looks good — trace is clean (session-derived userId, owner-filtered write, code kept out of the event/audit payload, safe additive migration) and tests cover the key paths. One thing I'd like fixed before merge: the missing max length on the anti-phishing code input. The rest are minor / follow-up-able.

mp-blurify:
## Code review — findings

### Blocking

**1. `packages/core/src/contracts/schemas/identity.ts:206` — unbounded input on a value appended to every email**

`z.string().trim().min(1)` has no `.max()` and no charset limit, backing an unbounded `text` column, and core sets no request body size limit. Every other player-supplied string in this same file is bounded (`PasswordSchema` 128, `name` 255, `LanguageSchema` 35).

Failure: a player POSTs a multi-megabyte `code`. It is stored, then appended to the body of *every* subsequent email for that account, so the provider rejects the oversize message and their regulatory mail (cooling-off, self-exclusion, KYC resubmission) permanently dead-letters via `mail.service.ts:106` `mail.regulatory_delivery.failed` — a self-inflicted denial of compliance email.

Amplified because `DrizzleAdminUserDirectory.list/get/lookupUsers` all `select()` full rows, so the blob rides along on every back-office user page and every chat online-count lookup. Embedded newlines are also accepted, letting the player inject extra paragraphs into their own rendered mail (self-directed, so not separately blocking).

**2. `packages/core/src/mail/adapters/default-email-template-renderer.ts:153` + `packages/core/src/contracts/adapters/email-template.ts:11` — the control lives in an operator-replaceable adapter**

The footer is added as an *optional* 4th positional parameter on `EmailTemplateRenderer`. `docs/adapters/mail.md` (not touched by this PR) still documents `EMAIL_TEMPLATE_RENDERER.render(template, locale)` and states an overlay renderer "owns the rendered result for each key; there is no automatic fallback to the platform's English renderer" — and the stock sender never actually delivers mail, so every real operator binds their own renderer.

A custom renderer whose `render` takes two parameters still satisfies `EmailTemplateRenderer` structurally (TS allows fewer params), so there is no compile error, no test, and no doc line to warn them. Net effect in production: no email carries the code, or only some do once a partial overlay is written — the "trains players to trust codeless mail" failure mode, which defeats the point of the control.

Suggested fix: append the footer in `MailService.deliver` (`packages/core/src/mail/service/mail.service.ts:112`) *after* `renderer.render(...)` returns, so no overlay can drop it. At minimum, document the parameter in `docs/adapters/mail.md`.

### Worth fixing

- **`packages/core/src/audit/plugin.ts:817`** — the record carries only `after: { antiPhishingCodeSet: true }` and no `before`. `docs/standards/audit.md:5` requires "meaningful before/after state", and the sibling control does exactly that (`identity.security.withdrawal_pin.set` carries `before: { withdrawalPinSet: wasAlreadySet }`). The service already has the prior state in hand at `identity.service.ts:1685`. Without it, an account-takeover investigation cannot distinguish a first-time set from a silent replacement of an existing code — the single most diagnostic fact about this control.
- **`packages/core/src/pam/identity/service/identity.service.ts:1656`** — no rate limiter and no notification on change. The sibling withdrawal-pin mutation consumes `RATE_LIMIT_KEYS.WITHDRAWAL_PIN_MUTATION`; this route consumes nothing, so the unbounded-length write above can also be hammered. More importantly, an attacker holding a stolen session can set their own code and the victim is never told; every later genuine platform email then carries the *attacker's* code, so the control certifies the attacker's phishing baseline instead of exposing it. Standard mitigation is a `mailDispatch.toUser` notice on change, rendered with the *previous* code.
- **`packages/core/src/contracts/adapters/admin-user-directory.ts:27`** — putting the value on `AdminUserRow` exposes it to every consumer of `ADMIN_USER_DIRECTORY` (admin-console, chat, chat-commands, wallet, compliance, tag). All were traced; nothing leaks it today, but only because `AdminUserSchema` is a non-strict `z.object` and oRPC's `validateOutput` returns the zod-parsed value, silently stripping the extra key. `BackofficeService.toAdminUser` (`packages/core/src/admin-console/service/backoffice.service.ts:27`) spreads the whole row, so the moment anyone widens that output schema or adds a route returning the raw row, every support agent can read every player's recognition code. The in-code comment claiming it "is not treated as secret" is wrong about the threat model — the control's entire value is that nobody outside the player and the mail renderer knows it. A narrow mail-only read would keep the blast radius at one consumer.

### Notes
- The authz trace is clean: the caller is resolved from the session (never from input), the write is `where(eq(user.id, userId))`, the raw value is kept out of the event and audit payloads, the migration is additive-nullable, and the HTML path escapes the code (with a test proving `&lt;script&gt;` is neutralized). No IDOR, no unauthenticated exposure, no log leak, no hygiene violation.
- `antiPhishingCodeSetAt` is written at `identity.service.ts:1689` and never read or exposed anywhere.
- The e2e (`packages/testing/src/__tests__/anti-phishing-code.e2e.test.ts`) never asserts that a captured outgoing email contains the code — coverage is unit-level only, so blocking item 2 has no regression guard.

packages/core/src/contracts/schemas/identity.ts
damianrzepka: Fixed in ee005db2. The input now trims, requires at least one character, and caps the stored/rendered value at 255 characters.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: Fixed in ee005db2. An unchanged submission now returns the existing controls before the limiter/write/event/notification path, matching the sibling control and avoiding redundant audit rows.

packages/core/src/pam/identity/schema/index.ts
damianrzepka: Fixed in ee005db2. Removed the timestamp from the application schema and write path because nothing consumes it. The already-generated additive migration retains the historical nullable column; no runtime code reads or writes it.

packages/core/src/contracts/adapters/admin-user-directory.ts
damianrzepka: Fixed in ee005db2. The code no longer appears on AdminUserRow. Mail now resolves it through the dedicated MAIL_RECIPIENT_DIRECTORY port, whose only consumer is MailService; the query selects only the delivery fields.

damianrzepka:
Addressed the findings in ee005db2. The input is capped at 255; MailService now appends the escaped footer after any renderer completes, so an overlay cannot drop it; and the raw code moved off AdminUserDirectory into a dedicated mail-recipient port. The audit event now carries only wasAlreadySet for meaningful before/after state, repeated values short-circuit, and writes are rate-limited. A security notification is queued on a real change and carries the prior code via the address-only path, so it is not stamped with the replacement code. Added coverage for the 255-character boundary, unchanged submissions, audit state, change notices, and core-owned footer delivery. pnpm -F @openora/core check:types is green.

packages/core/src/contracts/schemas/identity.ts
```diff
@@ -225,10 +225,18 @@ export const SecurityControlsSchema = z.object({
...
+  // a recognition signal readable only by the authenticated player it belongs to, same gate as
+  // the rest of this schema.
+  antiPhishingCode: z.string().nullable(),
 });
 
 export const SetLoginWithdrawalAlertsInputSchema = z.object({ enabled: z.boolean() });
 
+// Non-empty after trimming incidental leading/trailing whitespace (eg from copy-paste),
+// case-sensitive, and capped to keep every delivered email bounded - no reauth, set/overwrite only.
+export const SetAntiPhishingCodeInputSchema = z.object({ code: z.string().trim().min(1).max(255) });
```
jakubfilinger-b, line 238: BF-520 explicitly requires no length or character restriction, but `.trim().min(1).max(255)` changes leading/trailing characters and rejects whitespace-only or 256+ character values; please resolve the security/spec conflict in the ticket first, then align the schema and tests with the agreed bounded representation.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1899,6 +1905,75 @@ export class IdentityService {
...
+    const before = await this.securityControlsFor(userId);
+    if (before.antiPhishingCode === input.code) {
+      return before;
+    }
+    await assertRateLimit(
+      this.limiter,
+      makeRateLimitKey(RATE_LIMIT_KEYS.ANTI_PHISHING_CODE_MUTATION, userId),
+      ANTI_PHISHING_CODE_RATE_LIMIT,
+    );
+    await this.drizzle.db
```
jakubfilinger-b, line 1954: The row is committed before the fallible player lookup and mail enqueue, so a queue failure returns an error after changing the code, while retrying the same value exits at the no-op branch and never restores the missing audit/notification effects; persist a retry-safe notification intent atomically with the mutation (for example via the transactional outbox) and cover enqueue failure plus retry.

packages/core/src/mail/service/mail.service.ts
```diff
@@ -109,7 +128,10 @@ export class MailService {
       });
       return;
     }
-    const rendered = await this.renderer.render(job.template, resolved.locale, resolved.name);
+    const rendered = appendAntiPhishingCode(
```
jakubfilinger-b: This producer/consumer change is not rolling-deploy safe: old `mail-send` workers can still consume ordinary jobs without appending the footer, and they reject the new `securityAntiPhishingCodeChanged` template before delivery; deploy backward-compatible worker/schema support before activating the setter, or version/route the queue so old workers cannot claim new jobs.

packages/core/src/pam/identity/drizzle/migrations/meta/0014_snapshot.json
```diff
@@ -0,0 +1,1096 @@
+{
+  "id": "ce3d71fa-43ee-49e1-b180-16e9ede959e5",
+  "prevId": "e309fc35-0cec-4d3d-90e9-58c3b2f4fcdd",
```
jakubfilinger-b: The migration lineage is broken: this `prevId` points to snapshot `0012` and skips `0013`, while the PR also adds an unjournaled `0012_reflective_hellcat.sql` that duplicates `0014` and adds a column absent from the schema; remove the orphan and regenerate `0014` metadata from the current `dev` lineage.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1899,6 +1905,75 @@ export class IdentityService {
...
+      .set({ antiPhishingCode: input.code })
+      .where(eq(user.id, userId));
+    this.events.emit('identity.security.anti_phishing_code.set', {
+      userId,
+      playerId: await this.identityReader.getPlayerIdByUserIdSafe(userId),
+      wasAlreadySet: before.antiPhishingCode !== null,
+      ip,
+      userAgent,
+    });
+    await this.mailDispatch?.toAddress({
```
jakubfilinger-b, line 1968: Using `toAddress` intentionally keeps the attacker-selected replacement code out of the change alert, which is the safer behavior, but it contradicts BF-520's literal requirement that the saved code appear in every genuine email; document this security-notification exception explicitly in BF-520 and test first-set versus overwrite semantics.

jakubfilinger-b (changes requested):
Changes requested: resolve the BF-520 requirement conflicts, make the notification and queue rollout safe, repair the migration lineage, and regenerate the catalog because the current `verify` check fails `check:drift`.

jakubfilinger-b (commented):
## Review — BF-520 core side

The port design here is genuinely good: `MailRecipientDirectory` is deliberately narrow rather than widening `AdminUserDirectory`, with the reasoning in the docstring. `securityControlsFor` selects explicit columns, `textToHtml` escapes so the free-text code cannot inject HTML, and the rate limit plus `unauthorized_access` event on the non-player path are right. Findings below, most severe first.

### 1. [High] Orphaned duplicate migration `0012_reflective_hellcat.sql`

The PR adds **two** new migrations that both add `anti_phishing_code`:

- `0012_reflective_hellcat.sql` — **not** in `_journal.json` (idx 12 is `0012_worried_nuke`), collides on the `0012_` prefix with an already-applied migration, and adds `anti_phishing_code_set_at`, a column the schema never declares.
- `0014_spicy_mephistopheles.sql` — the real one, correctly journalled.

The first looks like a leftover from an earlier attempt. Drizzle reads the journal so it will not execute, but it will confuse anyone reading the folder and any tooling that globs `*.sql`. Please delete it.

### 2. [High] The footer is appended after `</html>`, so branded emails lose the code

`appendAntiPhishingCode` post-processes the finished render in `packages/core/src/mail/service/mail.service.ts`:

```ts
html: `${rendered.html}\n<p>${escapeHtml(footer)}</p>`,
```

betfeel's `BetfeelEmailTemplateRenderer` delegates to `@react-email`, whose `shell.tsx` emits a complete `<Html>…</Html>` document. The footer therefore lands **outside** the document, where Gmail and Outlook routinely strip it; where it does survive it renders as unstyled text below a fully branded email.

The failure mode is the bad one: an anti-phishing code silently missing from a genuine email trains the player to ignore its absence, which defeats the feature. The renderer should own placement.

### 3. [High] The PR description does not match the implementation

The body states `EmailTemplateRenderer.render()` was widened "with a 4th optional `antiPhishingCode` param, threaded from `MailService.deliver()`", and the test plan claims coverage of "4th `render()` arg threading".

In fact `packages/core/src/contracts/adapters/email-template.ts` is not in this diff at all — `render()` still takes three params (`template, locale, recipientName?`), `MailService` calls it with three arguments, and the new tests assert `render` was called with `(verify, 'fr', 'Ada')`.

Note that betfeel `!374` has **already implemented** the described 4-param design end to end (`BetfeelEmailTemplateRenderer` → `renderEmailTemplate` → `EmailRenderContext` → `Footer`). Because the 4th param is optional the mismatch compiles silently, so core simply never passes the code and betfeel's branded footer is dead code. The port version is the better design — it is the one that can place the code inside the branded layout. Suggest changing this PR to match `!374` rather than the reverse, which also resolves finding #2.

### 4. [Medium] New mail template key breaks consumer typechecks

`securityAntiPhishingCodeChanged` joins `MAIL_TEMPLATE_KEYS`. betfeel's `buildElement` switch in `packages/emails/src/render.tsx` is exhaustive with no `default` arm, so betfeel's `check:types` breaks on the next `@openora/*` bump, and until a layout exists the change-notification email has nothing branded to render. Worth calling out as a required consumer change.

### 5. [Low] `.update(user)` does not verify the write landed

No `.returning()` — if the user row vanished between the read and the write, the update silently no-ops yet still emits the event and sends the confirmation mail.

### 6. [Low] The `toAddress` choice is load-bearing but unexplained

Dispatching the change notification via `toAddress` rather than `toUser` is correct and subtle: `toUser` would stamp the footer with the *new* code, so an attacker who just changed it would send the victim an email bearing the attacker's own code. Because `toAddress` skips the footer, the body's `previousAntiPhishingCode` is what makes the email recognisable. That reasoning deserves a comment — a future refactor to `toUser` would silently defeat the notification.

docs/platform/system-design.md
```diff
@@ -289,7 +289,7 @@ flowchart TB
 
 <!-- gen:catalog-reference -->
 
-Generated from `docs/catalog.json` - 20 modules, 261 routes, 45 adapter ports, 117 events. Edit the code, then run `pnpm gen:catalog`.
+Generated from `docs/catalog.json` - 20 modules, 262 routes, 45 adapter ports, 118 events. Edit the code, then run `pnpm gen:catalog`.
```
mp-blurify: Generated artifacts were not regenerated after `contracts/adapters/mail-recipient-directory.ts` was added: `collectAdapters()` (`tools/gen/gen-catalog.ts:160`) matches its `createToken`, so a `mail-recipient-directory` entry belongs in `docs/catalog.json` - still 45 `"category"` entries, none for it - and this line should read 46 adapter ports. `pnpm check:drift` re-runs the generator and does `git diff --exit-code` on both files, so CI fails on this PR. `pnpm gen:catalog` and commit both.

packages/core/src/mail/service/mail.service.ts
```diff
@@ -32,6 +33,24 @@ const MAIL_ENQUEUE_OPTS = {
...
+  value.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
+
+function appendAntiPhishingCode(
+  rendered: RenderedEmail,
+  antiPhishingCode: string | null,
+): RenderedEmail {
+  if (antiPhishingCode === null) {
+    return rendered;
+  }
+  const footer = `Your anti-phishing code: ${antiPhishingCode}`;
```
mp-blurify: The footer string is hardcoded English and built after the renderer returns, so it sits outside the only localization seam the platform offers. An overlay that renders every template in `de` emits a German email whose last line reads `Your anti-phishing code: ...`, with no way to translate it - and the doc line this PR adds says the overlay "cannot omit this platform-owned footer". `docs/adapters/mail.md:28-30`. Pass the code into `renderer.render(...)` (or as a `RenderedEmail` field the renderer places) so the overlay owns the wording.

packages/core/src/mail/service/mail.service.ts
```diff
@@ -32,6 +33,24 @@ const MAIL_ENQUEUE_OPTS = {
...
+  return {
+    ...rendered,
+    text: `${rendered.text}\n\n${footer}`,
+    html: `${rendered.html}\n<p>${escapeHtml(footer)}</p>`,
```
mp-blurify: `${rendered.html}\n<p>...</p>` assumes `RenderedEmail.html` is a fragment. `DefaultEmailTemplateRenderer` returns joined `<p>` blocks so it holds in-repo, but the documented overlay case is "a designed HTML body", i.e. a full `<!DOCTYPE html>...</html>` document - the footer then lands after `</html>`, which Gmail's sanitizer drops. The one element this feature exists to show is invisible in exactly the deployment that has real mail. Same fix as above: hand the code to the renderer instead of concatenating onto its output.

mp-blurify:
Blocker: `docs/catalog.json` and `docs/platform/system-design.md` were not regenerated after the new adapter port landed, so `pnpm check:drift` fails CI on this PR. Beyond that, the anti-phishing footer is assembled after the renderer returns, which puts it outside the localization seam and breaks when a renderer returns a full HTML document - details inline. The points already raised on this PR read as answered by the current diff.

docs/platform/system-design.md
damianrzepka: Fixed in c18fa7e7. Regenerated docs/catalog.json and docs/platform/system-design.md; the catalog now records 46 adapter ports and check:drift passes.

packages/core/src/mail/service/mail.service.ts
damianrzepka: Fixed in c18fa7e7. The optional fourth renderer argument is restored and MailService passes the code through. The renderer now owns localized wording and placement.

packages/core/src/mail/service/mail.service.ts
damianrzepka: Fixed in c18fa7e7. The renderer receives the code and places it inside its own HTML document; MailService no longer concatenates markup after the rendered output.

damianrzepka:
Addressed the new review in c18fa7e7. Regenerated the catalog (46 adapter ports; check:drift passes), restored the optional fourth EmailTemplateRenderer argument so branded renderers own translated in-document placement, and removed the orphaned 0012_reflective_hellcat migration. The anti-phishing-code change notification remains a new mail-template contract: its branded consumer renderer needs to add that template before adopting this core version; that work belongs in the consumer overlay, not core. The address-only dispatch now has an explicit comment explaining why it must not become toUser. Targeted mail tests pass; the full verify command began successfully but was terminated by this environment before completion.

packages/core/src/contracts/schemas/identity.ts
damianrzepka: Valid specification conflict. Core currently keeps trim, non-empty, and 255-character bounds to prevent player-controlled regulatory-email amplification, but BF-520 reportedly says no restriction. I cannot silently choose between incompatible requirements; the ticket needs an explicit security decision before schema or test changes.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: Valid. Direct enqueue is not retry-safe after the database write. The shipped transactional outbox persists domain-event envelopes; preserving this notice safely needs a transactionally persisted encrypted notification intent, a retrying consumer, and a stable mutation ID. I have not added an ad-hoc non-atomic workaround.

packages/core/src/mail/service/mail.service.ts
damianrzepka: Valid deployment concern. The new setter and template are not safe to activate while old mail workers may claim jobs. This needs an operator rollout gate or versioned backward-compatible worker path before activation; core has no deployment coordinator to guarantee order, so I have not claimed it is safe.

packages/core/src/pam/identity/drizzle/migrations/meta/0014_snapshot.json
damianrzepka: Already fixed in c18fa7e7: removed the orphaned 0012_reflective_hellcat.sql and regenerated from current dev lineage. 0014 is journaled after dev 0013.

packages/core/src/pam/identity/service/identity.service.ts
damianrzepka: The address-only exception is documented in code and tests cover first-set versus overwrite notification payloads. The requested BF-520 wording is ticket and product-record work; it should explicitly allow this security notice to carry the previous code rather than the newly saved code.

jakubfilinger-b (dismissed):

packages/core/src/pam/identity/contract/index.ts
```diff
@@ -166,6 +167,13 @@ export const identityContract = {
     removeWithdrawalPin: oc
       .route({ method: 'DELETE', path: '/identity/security/withdrawal-pin' })
       .output(SecurityControlsSchema),
+
+    // No reauth (not money-moving, unlike the withdrawal PIN); set/overwrite only, no
+    // remove route. Calling it again with a new value changes the code.
+    setAntiPhishingCode: oc
```
mp-blurify, line 173: **Blocker — acceptance criterion not met.** There is no way to clear the code once set: `setAntiPhishingCode` is the only route, and `SetAntiPhishingCodeInputSchema` (`contracts/schemas/identity.ts:238`) rejects the empty string via `.min(1)`, so a player who sets a phrase is stuck with one forever. The requirement is "set **and clear**"; the sibling control on the line above ships `removeWithdrawalPin` as a separate `DELETE` for exactly this.

Fix: add `removeAntiPhishingCode` as `DELETE /identity/security/anti-phishing-code` writing `antiPhishingCode: null`, with its own event/audit path and an E2E case.

packages/core/src/contracts/adapters/email-template.ts
```diff
@@ -8,6 +8,7 @@ export type EmailTemplateRenderer = {
     template: MailTemplate,
     locale: string,
     recipientName?: string | null,
+    antiPhishingCode?: string | null,
```
mp-blurify, line 11: `antiPhishingCode` is an optional 4th parameter, so every overlay renderer that implements `render(template, locale, name)` still satisfies the port, compiles clean, and silently emits mail with no code — and `docs/adapters/mail.md:59` states an overlay owns the rendered result for every key, with no fallback to the platform renderer. In any branded deployment the feature therefore stops working with no type error, no test and no runtime signal.

Fix: make the parameter required on `EmailTemplateRenderer.render` so an overlay cannot silently omit it, and state the obligation in `docs/adapters/mail.md`.

packages/core/src/mail/adapters/default-email-template-renderer.ts
```diff
@@ -144,16 +144,33 @@ const PLAIN_EMAIL_TEMPLATES: { [K in EmailTemplateKey]: PlainTemplate<K> } = {
...
-const renderDefaultEmail = (template: MailTemplate, locale: string): RenderedEmail => {
+const renderDefaultEmail = (
+  template: MailTemplate,
+  locale: string,
+  antiPhishingCode?: string | null,
+): RenderedEmail => {
   const plain = PLAIN_EMAIL_TEMPLATES[template.key] as PlainTemplate<typeof template.key>;
-  const { subject, text } = plain(template.data, locale);
+  const { subject, text: body } = plain(template.data, locale);
+  const text = antiPhishingCode ? `${body}\n\nYour anti-phishing code: ${antiPhishingCode}` : body;
```
mp-blurify, line 163: The footer branch `antiPhishingCode ? ... : body` — the only code in the repo that puts the code into an email — has no test. No call site in `default-email-template-renderer.test.ts` passes a 4th argument (the all-keys loop at line 92 calls `render(template, 'en')`), and the `mail.service.test.ts` assertion that used to check the rendered text and its HTML escaping was replaced in c18fa7e7 with `text: 't'` / `html: '<p>h</p>'`. The PR description claims "renderer footer present/absent, HTML-escaping of the code" are covered; neither is.

Rule: `docs/standards/testing.md:11,48`.
Fix: add renderer cases asserting the footer appears with a code, is absent without one, and is HTML-escaped in `html`.

packages/core/src/pam/identity/contract/index.ts
damianrzepka: Not applicable to this PR. The agreed core scope and contract are set/overwrite only, so no clear route is intentionally exposed. I have not widened the public API without a confirmed change to that requirement. The current implementation remains covered by the overwrite E2E. Fixed in 82bd2c6e.

packages/core/src/contracts/adapters/email-template.ts
damianrzepka: I agree that every renderer must place a non-null code in its localized in-document layout, and the adapter documentation now states that obligation. I did not make the fourth positional parameter required: TypeScript structurally permits a renderer with fewer parameters, so that change would not prevent an existing three-parameter overlay from silently ignoring the code. Enforcing it needs a different contract shape rather than a nominal optionality change. Updated in 82bd2c6e.

packages/core/src/mail/adapters/default-email-template-renderer.ts
damianrzepka: Fixed in 82bd2c6e. The default renderer now has coverage that asserts the footer is present with a code, absent for null, and HTML-escapes the rendered code.

packages/core/src/pam/identity/drizzle/migrations/meta/0014_snapshot.json
damianrzepka: Confirmed: the migration snapshot still pointed at 0012 and made drizzle-kit reject the lineage. I corrected 0014 to inherit from the 0013 snapshot, then ran pnpm regen successfully; no generated migration changes remained. Fixed in 82bd2c6e.

packages/core/src/pam/identity/service/identity.service.ts
zaxovaiko: Confirmed and fixed in 7b5005fe. The enqueue is now wrapped: a queue failure is logged instead of thrown, so the caller no longer receives an error for a change that already landed and the row, event and audit record stay consistent. A full transactional notification intent is more machinery than one advisory notice warrants, so the trade-off is explicit: on a queue outage the change notice is dropped and logged, not silently lost. Covered by the new integration case "keeps the committed change when the notification enqueue fails".

packages/core/src/contracts/schemas/identity.ts
zaxovaiko: Checked against the source: BF-520 acceptance criteria and the Confluence spec (Anti-Phishing Code, page 232947713, Business Rules) both state the code is free text "with no length or character restriction, and is case-sensitive". The conflict is real, and the cap is deliberate: the value is appended to the body of every outgoing email, so an unbounded string makes the provider reject oversize messages and dead-letters the regulatory mail for that account. The trim is a second, smaller deviation - it drops leading and trailing whitespace, which the spec does not ask for. Keeping `.trim().min(1).max(255)` pending an explicit product decision on the ticket; please confirm the bounded representation there so the spec and the schema agree.

packages/core/src/pam/identity/contract/index.ts
zaxovaiko: Checked against the ticket: BF-520 lists four acceptance criteria (open the modal, enter a free-text phrase and save, no length or character restriction and case-sensitive, the saved code appears in every genuine platform email), and the Confluence spec (page 232947713) describes setting a code only. Neither states "set and clear", so the missing DELETE is not an unmet acceptance criterion. Set/overwrite-only is the agreed scope; a clear route can follow when the product asks for one.

packages/core/src/contracts/adapters/email-template.ts
zaxovaiko: Agreed with the concern, but a required parameter does not enforce it: TypeScript accepts a function with fewer parameters wherever one with more is expected, so an overlay `render(template, locale)` still satisfies the port with or without the optionality change, and nothing breaks at compile time either way. The obligation is documented in `docs/adapters/mail.md` instead. Real enforcement needs a different contract shape - for example passing a single context object the renderer destructures, or having the renderer return a placement marker the service can assert on - which is a separate change to the port.

packages/core/src/mail/service/mail.service.ts
zaxovaiko: Stands, and it is a rollout requirement rather than a code change. Scope check: the footer itself is safe, since an old renderer simply ignores the new fourth argument and an old worker still delivers ordinary mail. The exposure is the new `securityAntiPhishingCodeChanged` template key - an old worker fails `MailTemplateSchema` on that job, retries with backoff, and loses the notice if the retries run out before the new workers are in place. Mitigation is deployment order: roll the mail workers before the setter route is reachable. Flagging it in the release notes; core has no deployment coordinator that can guarantee the order on its own.

packages/core/src/pam/identity/service/identity.service.ts
zaxovaiko: The code-side reasoning is in place: the dispatch carries a comment explaining why it must stay address-only, and the integration tests assert the first-set payload (`previousAntiPhishingCode: null`) and the overwrite payload (the prior code) separately. The remaining half is ticket wording - BF-520 should record that this one security notice deliberately carries the previous code rather than the newly saved one, so the "every genuine email carries the saved code" rule is not read as covering it.

zaxovaiko (approved):

PR #150, opened 2026-09-08
feat(identity): revoke other sessions on password change

jakubfilinger-b:
## What

`identity.changePassword` now invalidates every *other* session for the player, immediately:

- passes `revokeOtherSessions: true` to better-auth, so it deletes the player's other session rows and mints a fresh one for the caller (its rotated cookie is already forwarded by `forwardCookies`);
- emits `identity.sessions.revoked_all` so the `streamSession` SSE handler pushes `{ type: 'revoked' }` to the player's other tabs in real time;
- emits a dedicated `identity.password.changed` event so the audit trail records the credential change as such, not merely as "revoked all sessions".

The **acting session is spared**: `identity.sessions.revoked_all` carries `exceptSessionId`, and `streamSession` skips the push for a connection whose session id matches. The admin "revoke all sessions" path leaves `exceptSessionId` unset and still kicks everyone.

## Review round 2

- **exceptSessionId is now the caller's pre-rotation session id** (resolved from `getSession` *before* the better-auth call), not the freshly minted one. The open `streamSession` connection captured its session id at stream-open — that is the pre-rotation id, and the only value the SSE compare can ever match, so this is what actually spares the initiating tab. The old approach (resolve from the rotated token) could never match and the initiating tab revoked itself.
- **Dedicated `identity.password.changed` audit event** added and subscribed in the audit module; `identity.sessions.revoked_all` bumped to v2 in `domainEventVersions` for the added `exceptSessionId` field.
- **Rotation race** — better-auth writes the password hash and deletes sessions in separate, untransacted statements, so a sign-in that verified the *old* password microseconds earlier can still insert its session row just after `deleteUserSessions()`. Mitigated with a short deferred sweep (`sweepRacedSessionsAfterPasswordChange`, `setTimeout` `RACED_SESSION_SWEEP_DELAY_MS` = 2s, `.unref()`): it deletes any session row for the user created up to the instant the change committed, sparing the caller's new session by token. Residual exposure is that ~2s window. A full session-generation invariant across better-auth's sign-in path (advisory lock / serializable) is deliberately **not** in this PR — it is a cross-cutting change to the auth hot path that needs its own ticket and review.

## Review round 3

- **Non-remembered session lifetime preserved.** `revokeOtherSessions` makes better-auth mint the caller's replacement session via `createSession(userId)` with no `dontRememberMe` argument, so a caller who signed in with `rememberMe: false` was silently getting a full-TTL row. better-auth's own `setSessionCookie` still honours the `dont_remember` cookie for the *cookie* maxAge — only the DB row was wrong. `changePassword` now clamps the rotated row's `expiresAt` back to the 24h non-remembered TTL when that cookie is present (`hasDontRememberCookie`).
- **End-to-end coverage of both changed routes.** New `packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts` (`bootTestApp`): opens `streamSession` on sessions A and B, changes the password on A, asserts B receives `{ type: 'revoked' }` and A (the acting tab) does not; plus the rotated-session TTL for a remembered vs a non-remembered caller. The identity int-test row assertions (which re-read their own stub) now point at this e2e for the real revoke + SSE-suppression coverage.
- **Rotation race — still the 2s deferred sweep.** The durable fix (a session-generation / transactional invariant across change-password + sign-in) is deferred to its own ticket rather than folded in here — it is an auth-hot-path change. The sweep plus better-auth's own delete-all narrows the exposure from indefinite to ~2s; the follow-up closes it.

## Review round 4

- **The deferred sweep is no longer silent.** It force-deletes a raced session with no push and no audit record — a client stuck on a session that suddenly 401s, and nothing in the audit trail says it was revoked. `sweepRacedSessionsAfterPasswordChange` now returns the ids it deleted; `changePassword`'s setTimeout callback emits `identity.sessions.revoked_all` (same event, same audit path as the primary revoke) when it deleted anything, sparing the still-legitimate rotated session by id. New int test exercises the full wiring (not just the pure sweep function): seeds a raced row, runs the real service call, waits out the delay, asserts the second emit.
- **The e2e now proves the acting session survives, not just that it wasn't pushed.** Pulls the rotated `Set-Cookie` off the change-password response and asserts it authenticates (`GET /identity/security/me` → 200) while the pre-change cookie now 401s.
- **Dropped the redundant `getSession` round trip in `changePassword`.** The router already resolves the caller's `userId`/`sessionId` once per request (`context.auth`, populated centrally); `changePassword` now takes those as a `caller` param instead of re-deriving them via `currentSession(headers)`.
- **`sweepRacedSessionsAfterPasswordChange` takes a named object param** (`{ userId, sparedToken, deadline }`) instead of four positionals, per `docs/standards/functions.md`.
- **Dropped the client ticket id from the e2e test's docblock** — it was tripping `pnpm check:hygiene`, a CI gate `pnpm verify` doesn't cover.
- Rebased onto the latest `dev` (picked up `docs/catalog.json` moving to generated/gitignored).

## Notes

- No schema/table change, no migration.
- `PasswordSchema = z.string().min(12).max(128)` was already on `dev`; this PR only adds the doc comment and a unit test. Despite the branch name there is no new minimum and no lockout risk for existing users.
- The rotation race is mitigated, not fully closed — see "Review round 3".


mp-blurify:
## Code review — findings

### Blocking

**1. `packages/core/src/pam/identity/service/identity.service.ts:1587-1616` + `packages/core/src/pam/identity/router/index.ts:134-147` — `exceptSessionId` can never match**

`exceptSessionId` is resolved from `rotatedToken`, i.e. the **new** session better-auth just minted. But `connectionSessionId` in the SSE handler is captured once at stream-open from `context.auth.sessionId` (`create-app.ts:446-467`), which is the **pre-change** session id. better-auth's `dist/api/routes/update-user.mjs` runs `deleteUserSessions(user.id)` and then `createSession(...)`, deleting that old row — so the two ids can never be equal.

Failure scenario: a player has `/identity/session/stream` open, changes their password, `changePassword` returns SUCCESS with a valid rotated cookie, and the same tab immediately receives `{ type: 'revoked' }` and logs out — all tabs in that browser, since they share the cookie and each holds its own stream. The emit happens synchronously before the response returns, so the client cannot have reconnected under the new session id first. This is the exact outcome the PR exists to prevent.

Fix: capture the caller's session id **before** the better-auth call (`auth.api.getSession({ headers })` already runs in `currentUserId`, or plumb `getSessionId(context)` from the router) and emit that as `exceptSessionId`.

**2. `packages/core/src/pam/identity/__tests__/identity.service.int.test.ts:1004-1012` — the test encodes the broken semantics**

The new test asserts `exceptSessionId: remaining[0]!.id` (the freshly minted session) and, two lines above, asserts that id is *not* any of the pre-change session ids. It pins the bug above as expected behaviour, so it ships on a green CI. Note also that `createAuth` is mocked in this suite, so real better-auth revocation is never exercised, and no test covers the `streamSession` branch at all.

### Worth fixing

- **`packages/core/src/pam/identity/service/identity.service.ts:1609`** — a self-service password change now emits only `identity.sessions.revoked_all`; there is no `identity.password.changed` topic in `events.ts` (only `identity.password.reset`, line 243). `audit/plugin.ts:268` maps this to `actorType: 'player'`, `resourceType: 'user'`, so the audit trail for a credential change reads "player revoked all sessions". On a regulated platform that is the one record an auditor looks for, and it is absent — and the topic now carries two distinct meanings separated only by an optional field.
- **`packages/core/src/contracts/schemas/events.ts:288-296`** — the payload gained `exceptSessionId` but `domainEventVersions` has no entry for `identity.sessions.revoked_all` (still 1), while the sibling `identity.session.revoked` is at 2 and `rg.self_exclusion.activated` is at 3 explicitly for "durationMonths added". Cross-process producer/consumer agreement runs off `eventCatalog()`, so consumers see an unchanged version with a changed payload.
- **`packages/core/src/pam/identity/service/identity.service.ts:1578-1598`** — no atomicity across the new step. better-auth writes the password hash (`updateAccount`) and *then* deletes sessions, in separate statements with no transaction; the `passwordMeetsPolicy` update runs after that in a third. If the delete or the flag write fails, the password is already changed, old sessions stay live, and `ensureOk` maps the 500 to `BAD_REQUEST` — the player is told the change failed and will not retry, leaving a compromised session alive. This window did not exist before the PR. At minimum, log/alert on the partial state.

### Notes
- The revocation half itself is sound: sessions are DB rows validated per request (`SessionResolver.resolveSession` → `auth.api.getSession`), there is no cookie cache, no `secondaryStorage`, and no jwt/bearer/apiKey plugin in `createAuth` — so deleted rows really are dead, for remember-me and every other device.
- `PasswordSchema = z.string().min(12).max(128)` was already on `dev`; this PR only adds the doc comment and a unit test. Despite the branch name there is no new minimum and no lockout risk for existing users. `currentPassword` re-auth is enforced by better-auth.
- better-auth's twoFactor trust-device cookie and the `admin_trusted_device` rows survive a password change. Not exploitable without the new password, but worth an explicit decision, since password change is the standard response to a suspected compromise.
- `phone_verification_session.sessionId` is `onDelete: 'cascade'`, so `deleteUserSessions` will not hit an FK violation; the new session row does get real `ipAddress` / `userAgent` (1.6.22 `createSession` reads them from the ambient auth context), so `AdminGuard`'s `assertSessionIntact` fingerprint check will not kill the rotated admin session.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1568,17 +1572,48 @@ export class IdentityService {
       CHANGE_PASSWORD_RATE_LIMIT,
     );
     const res = await this.api.changePassword({
```
damianrzepka, line 1891: `revokeOtherSessions` does not serialize password rotation with concurrent sign-in. A request using the old password can finish `verify()` before this endpoint updates the credential, then create its session after `deleteUserSessions()` has run. That newly-created session survives this revoke-all operation.

This breaks the immediate session-invalidation requirement on a compromised-password path. Please enforce a per-user atomic/session-generation invariant across credential verification, password update, invalidation, and session creation, and cover this interleaving in a test.

packages/core/src/pam/identity/router/index.ts
```diff
@@ -132,12 +132,19 @@ export function createIdentityRouter(
...
         (push) => {
           const unsubscribeRevoked = eventBus.on('identity.sessions.revoked_all', (event) => {
-            if (event.userId === userId) {
-              push({ type: 'revoked' });
+            if (event.userId !== userId) {
+              return;
+            }
+            // Self-service password change spares the acting session via
+            // exceptSessionId; the admin revoke-all path leaves it unset and kicks all.
+            if (event.exceptSessionId && event.exceptSessionId === connectionSessionId) {
```
damianrzepka, line 144: `connectionSessionId` belongs to the already-open SSE connection, so it is the pre-rotation session id. `exceptSessionId` is resolved from the newly minted session after better-auth deletes the old row. These ids can therefore never match for the tab that initiated the password change, and it still receives `{ type: `revoked` }`.

Please carry the pre-rotation/origin session id for SSE suppression (or otherwise identify the initiating stream), and add coverage for an old-session stream, a newly minted session, and another active session.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1568,17 +1572,48 @@ export class IdentityService {
...
+            .where(eq(session.token, rotatedToken))
+            .then((rows) => rows[0])
+        : undefined;
+      // Same event the admin "revoke all sessions" path emits, so the streamSession
+      // SSE handler pushes { type: 'revoked' } to this user's OTHER tabs.
+      // exceptSessionId spares the caller's fresh session from that push - it stays
+      // signed in. If the id can't be resolved, spare nothing (kick all) rather than
+      // name a stale row.
+      const { ip, userAgent } = extractClientMeta(reqHeaders);
+      this.events.emit('identity.sessions.revoked_all', {
```
damianrzepka: A password change is now auditable only as `identity.sessions.revoked_all`. The audit mapper uses the topic as the action, so this record is indistinguishable from a player manually revoking all sessions and does not establish that a credential was changed.

Please emit a dedicated, non-sensitive password-change event and subscribe it in the audit module. Every state-changing credential action needs an accurately named audit record.

damianrzepka (changes requested):
NO-GO — password rotation can leave a concurrent old-password login alive, and the initiating tab receives its own revocation event. The password-change audit trail also needs a distinct event.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1568,17 +1572,48 @@ export class IdentityService {
       CHANGE_PASSWORD_RATE_LIMIT,
     );
     const res = await this.api.changePassword({
-      body: { currentPassword: input.currentPassword, newPassword: input.newPassword },
+      // revokeOtherSessions: better-auth deletes every session for this user and
+      // mints a fresh one for the caller; forwardCookies below ships that cookie.
+      body: {
+        currentPassword: input.currentPassword,
+        newPassword: input.newPassword,
+        revokeOtherSessions: true,
```
mp-blurify, line 1897: Separate point from the race-condition thread six lines up. `revokeOtherSessions: true` makes better-auth replace the caller's session via `createSession(userId)` with no `dontRememberMe` argument (better-auth 1.6.22 `api/routes/update-user.mjs:172` -> `db/internal-adapter.mjs:180`, where that flag is the only thing that shortens `expiresAt`). A caller who signed in with `rememberMe: false` holds a 24h row (`sign-in.mjs:244`) and comes out of a password change with one expiring in 30 days (`session.expiresIn`, `server/auth/auth.ts:90`).

So changing a password silently promotes a deliberately short-lived session to the full TTL, and nothing in this flow clamps it back - the platform asserts the opposite elsewhere (`phone-login.service.int.test.ts:298`, "will not roll the session to 30 days"). Clamping the rotated row's `expiresAt` when the request carries better-auth's `dont_remember` cookie would keep the two paths consistent.

packages/core/src/pam/identity/router/index.ts
mp-blurify, line 144: Separate from the open thread on this line about which id is compared: whichever way that is resolved, nothing would catch it. `streamSession` appears in no test file, and neither changed route (`POST /identity/change-password`, `GET /identity/session/stream`) has an E2E in `packages/testing/src/__tests__`; the int test asserts the event payload only, so the push filter itself is never exercised. `docs/standards/testing.md:9` makes that E2E the acceptance artifact for a changed route. One `bootTestApp` test that opens the stream on session A, changes the password from session B, and asserts which stream receives `{ type: 'revoked' }` would pin the behaviour and would have caught the id mismatch.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
```diff
@@ -933,6 +937,105 @@ describe('IdentityService.resetPassword', () => {
...
+      expect.objectContaining({
+        body: expect.objectContaining({ revokeOtherSessions: true }),
+      }),
+    );
+    // Only the freshly-minted session remains; the pre-change rows are all gone.
+    const remaining = await db.drizzle.db
+      .select({ id: session.id, token: session.token })
+      .from(session)
+      .where(eq(session.userId, account.id));
+    expect(remaining).toHaveLength(1);
```
mp-blurify, line 1012: `changePasswordMock` deletes the session rows and inserts the rotated one itself, so `expect(remaining).toHaveLength(1)`, the token check and the "pre-change rows are all gone" assertion only re-read what the test's own stub just wrote. They cannot fail on a production regression, including dropping `revokeOtherSessions` entirely - only the `toHaveBeenCalledWith` at :990 covers that - while the `it` title claims the service revokes other sessions. `docs/standards/testing.md:30`: assert rows produced by real code, not by a double. Keep the call and event assertions here and move the revocation claim to an E2E where better-auth actually runs.

mp-blurify:
No blockers from me beyond what is already open here. Three items inline, each deliberately distinct from the existing threads: a remember-me TTL that the rotated session silently promotes to 30 days, the absence of any test over the stream's suppression branch, and an int test whose row assertions only re-read its own stub.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Addressed with a deferred sweep rather than full serialization. `sweepRacedSessionsAfterPasswordChange` runs `RACED_SESSION_SWEEP_DELAY_MS` (2s) after the change, `.unref()`'d, and deletes every session row for the user created up to the instant the change committed — so a row a concurrent old-password sign-in raced in just after `deleteUserSessions()` is caught, while anything created afterwards is spared. Residual exposure is that ~2s window. A fully-atomic per-user invariant across better-auth's sign-in path (advisory lock / serializable) is a cross-cutting change to the auth hot path — tracking it separately rather than growing this PR.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed. `exceptSessionId` is now the caller's PRE-rotation session id, resolved from `getSession` before the better-auth call (`currentSession(headers)`). That is the id the already-open `streamSession` connection captured at stream-open, so the SSE compare now matches and the initiating tab no longer revokes itself. The router compare is unchanged.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed. New `identity.password.changed` event, emitted alongside `identity.sessions.revoked_all` and subscribed in the audit module's shared identity self-action branch (actorType player/admin, resourceType user). `identity.sessions.revoked_all` bumped to v2 in `domainEventVersions` for the added `exceptSessionId`.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Acknowledged as a real regression, and distinct from the three NO-GO items. Fixing it means clamping the rotated session's `expiresAt` (and re-signing the cookie with a bounded maxAge) when the request carries better-auth's `dont_remember` cookie, mirroring the phone-login path — flagging as a follow-up rather than folding another session-lifetime change into this PR.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fair. The end-to-end cross-session-revoke assertion lives in the consumer browser E2E (`apps/e2e/tests/web/settings/change-password.spec.ts` scenario 5: stream open on context B, password change in context A, assert B is kicked), currently `test.fixme`'d there pending this canary. A core-side `bootTestApp` test of the `streamSession` push filter is a reasonable addition but not in this revision.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
jakubfilinger-b: Agreed the row assertions re-read the stub's own writes. Kept the `toHaveBeenCalledWith({ revokeOtherSessions: true })` and the event-payload assertions (the real-code contract at this tier) and added a direct test of `sweepRacedSessionsAfterPasswordChange` against real rows. The 'better-auth actually revokes' claim is covered by the browser E2E above.

jakubfilinger-b:
Review round 2 pushed (`6a619c30`, rebased on dev):

- **exceptSessionId** → caller's pre-rotation session id, so the initiating tab no longer self-revokes.
- **identity.password.changed** event added + audited; `identity.sessions.revoked_all` → v2.
- **Rotation race** mitigated with a 2s deferred `sweepRacedSessionsAfterPasswordChange` (residual ~2s window; a full session-generation invariant is deferred to its own ticket — it's an auth-hot-path change).

Follow-ups acknowledged inline: `dont_remember` TTL clamp on the rotated session, and a core-side `bootTestApp` test of the streamSession push filter.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1815,24 +1858,88 @@ export class IdentityService {
...
+        ...(caller?.sessionId ? { exceptSessionId: caller.sessionId } : {}),
+      });
+      // better-auth rotates the password hash and deletes sessions in separate,
+      // untransacted statements; a sign-in that verified the OLD password just before
+      // the rotation can still insert its session row right after. Re-sweep shortly
+      // after to delete it, sparing the caller's fresh session by token. Skipped when
+      // the rotated token is unknown - without it the sweep cannot tell the caller's
+      // own new session apart. Residual exposure: RACED_SESSION_SWEEP_DELAY_MS.
+      if (rotatedToken) {
+        const timer = setTimeout(() => {
```
damianrzepka: The deferred sweep does not close the race this PR claims to mitigate. It deliberately leaves a two-second window, and it runs only once with `.unref()`: a sign-in that verified the old password before the hash update can create its session after `sweepDeadline`, or a process restart can drop the sweep entirely. That session then survives indefinitely. This compromised-credential path needs a durable per-user generation/serialization invariant across verification, password update, and session creation; an in-memory timer is not durable scheduled work (`.agents/rules/messaging-and-microservices.md`).

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1815,24 +1858,88 @@ export class IdentityService {
...
+    // connection captured at stream-open, which is what the SSE handler compares
+    // `exceptSessionId` against - so this is the value that spares the initiating tab.
+    const caller = await this.currentSession(headers);
+    const userId = caller?.userId ?? null;
     await assertRateLimit(
       this.limiter,
       `change-password:${userId ?? 'anonymous'}`,
       CHANGE_PASSWORD_RATE_LIMIT,
     );
     const res = await this.api.changePassword({
```
damianrzepka, line 1891: `revokeOtherSessions` makes better-auth mint the replacement session without its `dontRememberMe` argument. Consequently, a caller carrying the signed `dont_remember` cookie gets a full-TTL replacement session after changing their password, silently promoting a deliberately non-remembered session. Preserve the original lifetime by clamping the rotated row and its cookie when that cookie is present.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
```diff
@@ -944,6 +949,146 @@ describe('IdentityService.resetPassword', () => {
...
+    getSessionMock.mockResolvedValue({
+      user: { ...betterAuthUser, id: account.id },
+      session: { id: current.id },
+    });
+
+    // createAuth is mocked in this suite, so better-auth's own revokeOtherSessions
+    // never runs - stand in for it: delete EVERY session (including the caller's,
+    // as better-auth does) and mint a fresh one, returning its token in the body.
+    const rotatedToken = randomUUID();
+    changePasswordMock.mockImplementation(async () => {
```
damianrzepka, line 979: The row assertions below still only re-read rows written by `changePasswordMock`, so they do not verify better-auth’s revocation or the SSE suppression branch. The changed routes need a `bootTestApp` E2E that opens streams on sessions A and B, changes the password on A, and proves only B receives `{ type: `revoked` }` (`docs/standards/testing.md`).

damianrzepka (changes requested):
NO-GO — password rotation still permits a concurrent old-password sign-in to survive, and the session-lifetime regression plus end-to-end coverage remain open.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Update on this one: the deferred sweep is staying as the mitigation for now — extending it to a durable session-generation invariant means an advisory lock (or equivalent) around better-auth's sign-in hot path, which is a change we want on its own ticket with its own review rather than folded in here. Tracked as a follow-up. Continued on the newer thread at :1963.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Also now covered end to end: `packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts` opens two real `streamSession` connections, changes the password on one, and asserts the other receives `{ type: 'revoked' }` while the acting connection does not — i.e. the pre-rotation id the SSE connection holds is exactly what `exceptSessionId` now carries.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 5250d025. `changePassword` now clamps the rotated session row's `expiresAt` back to the 24h non-remembered TTL when the request carries the `dont_remember` cookie (`hasDontRememberCookie`). better-auth's own `setSessionCookie` already honours that cookie for the session-cookie maxAge — it's only the DB row `createSession(userId)` mints at full length. Covered by the rememberMe:false / rememberMe:true TTL cases in the new e2e.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed in 5250d025 — `packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts` is a `bootTestApp` e2e that opens two live `streamSession` connections on two real sessions, changes the password on one via `POST /identity/password/change`, and asserts the other connection gets `{ type: 'revoked' }` and the acting one stays silent. The push filter itself is now exercised.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
jakubfilinger-b: The real revoke + SSE-suppression path is now covered by the new `bootTestApp` e2e (`change-password-session-revoke.e2e.test.ts`). Left the int-tier row assertions in place as a stub sanity check, with a comment at :1007 pointing at the e2e for the production-regression coverage.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Acknowledged. Decision on our side is to keep the 2s deferred sweep as the mitigation for this PR and take the durable fix (a session-generation / transactional invariant across the change-password + sign-in path) as its own ticket — it's an auth-hot-path change and we don't want it riding in on this one. The sweep + better-auth's own delete-all narrows the window from "indefinite" to ~2s; the follow-up closes it. Happy to link the ticket here once it's filed.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 5250d025 — the rotated row's `expiresAt` is clamped back to the 24h non-remembered TTL when the request carries the `dont_remember` cookie. The cookie maxAge was already correct (better-auth's `setSessionCookie` reads that cookie); this was only the DB row. New e2e asserts the rotated TTL for both a remembered and a non-remembered caller.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
jakubfilinger-b: Done in 5250d025: `packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts`. Opens `streamSession` on sessions A and B, changes the password on A, asserts only B receives `{ type: 'revoked' }` (A, the acting tab, does not). Also covers the rotated-session TTL for remembered vs non-remembered callers.

jakubfilinger-b:
Review round 3 pushed (`5250d025`, rebased on dev).

- **dont_remember TTL** (mp-blurify + damianrzepka): fixed — the rotated session row's `expiresAt` is clamped back to the 24h non-remembered TTL when the request carries the `dont_remember` cookie. The cookie maxAge was already correct via better-auth's `setSessionCookie`; only the `createSession` DB row was full-length.
- **bootTestApp E2E for the changed routes** (mp-blurify + damianrzepka): added `packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts` — two live `streamSession` connections, change on A, assert B is kicked and A is spared; plus rotated-session TTL for remembered vs non-remembered callers. 3/3 green.
- **Rotation race**: staying with the 2s deferred sweep for this PR. The durable session-generation invariant is an auth-hot-path change and is deferred to its own ticket — I'll link it here once filed.

Gate: check:types / lint / format / shape / boundaries / drift green; identity int (64) + the new e2e (3) green.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -87,6 +87,31 @@ function nodeHeadersToHeaders(nodeHeaders: NodeHeaders) {
...
+// `deleteUserSessions()` ran. That row predates `deadline` (captured once the change
+// is known committed), so this catches it while sparing anything created afterwards.
+// Exported for direct testing; the service schedules it via setTimeout.
+export async function sweepRacedSessionsAfterPasswordChange(
+  db: DrizzleService['db'],
+  userId: User['id'],
+  sparedToken: Session['token'],
+  deadline: Date,
+): Promise<void> {
+  await db
```
mp-blurify: **Blocker — a silent, unauditable revoke.** The sweep deletes session rows — forcibly revoking a player's sessions — and emits nothing: no `identity.sessions.revoked_all`, no audit record. A sign-in that lands at t+0.3s on the old password opens `/identity/session/stream`, is deleted at t+2s, and that client receives no `{ type: 'revoked' }` push (its UI stays live until its next request 401s) while the audit log holds no record that the session was killed.

This is separate from the open thread about `:1963`, which is about the size of the window, not about the sweep's own deletions being silent.

Rule: `docs/standards/audit.md` — every mutation that changes player state must produce an audit record with actor, resource, outcome.
Fix: have the sweep return the deleted session ids and emit `identity.sessions.revoked_all` (with `exceptSessionId` still set to the caller) when it deleted anything.

packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    await new Promise((resolve) => setTimeout(resolve, 200));
+
+    const res = await acting.post('/identity/password/change', {
+      currentPassword: PASSWORD,
+      newPassword: NEW_PASSWORD,
+    });
+    expect(res.status).toBe(200);
+
+    await expect(withTimeout(otherStream.result, 4000, null)).resolves.toBe('revoked');
+    await expect(withTimeout(actingStream.result, 1500, null)).resolves.toBeNull();
```
mp-blurify, line 155: **Blocker — the surviving half is not asserted.** This line proves the acting tab got no `revoked` push, which is not the same as proving the acting session still works. `revokeOtherSessions` deletes the caller's row too, so "the acting session stays alive" rests entirely on `forwardCookies` shipping the rotated `Set-Cookie` (`identity.service.ts:1906`) — and an already-open SSE connection stays silent whether or not the acting session was destroyed, because the generator never re-validates. `asPlayer` (`packages/testing/src/request.ts:81`) pins one fixed cookie and never re-reads `set-cookie`, so `acting` is never exercised after the change.

This differs from the resolved `router:144` coverage threads, which are about the `exceptSessionId` suppression branch.
Fix: assert the change-password response carries a rotated session cookie, and that a follow-up request with it authenticates while the pre-change cookie 401s.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1815,24 +1879,99 @@ export class IdentityService {
 
   async changePassword(input: ChangePasswordInput, reqHeaders: NodeHeaders, resHeaders: Headers) {
     const headers = nodeHeadersToHeaders(reqHeaders);
-    const userId = await this.currentUserId(headers);
+    // Resolve the caller's session BEFORE the change: better-auth's revokeOtherSessions
+    // deletes every session row (the caller's included) and mints a fresh one, so this
+    // id only exists now. It is also the id the caller's already-open streamSession
+    // connection captured at stream-open, which is what the SSE handler compares
+    // `exceptSessionId` against - so this is the value that spares the initiating tab.
+    const caller = await this.currentSession(headers);
```
mp-blurify: `changePassword` pays a second full `getSession` round-trip to re-derive a value the request already carries: `create-app.ts:447` publishes `{ userId, sessionId }` onto `context.auth`, and the router exposes it as `requireSessionId(context)` (`router/index.ts:32`, used at `:116` and `:126`) — the same id `streamSession` compares against at `router/index.ts:135`. The new private `currentSession` (`:581-587`) is also a line-for-line copy of `SessionResolver.resolveSession` (`server/auth/session-resolver.ts:30-36`).

Rule: `docs/standards/module-structure.md` — reuse the shared helpers, do not re-roll.
Fix: pass `requireSessionId(context)` into `changePassword` from the router and drop `currentSession`.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -87,6 +87,31 @@ function nodeHeadersToHeaders(nodeHeaders: NodeHeaders) {
...
 
+// Deletes every session row for `userId` created at or before `deadline`, except the
+// caller's freshly minted one (matched by `sparedToken`). Run a short delay after a
+// password change: better-auth rotates the password hash and deletes sessions in
+// separate, untransacted statements, so a sign-in that verified the OLD password just
+// before the rotation can still insert its session row immediately after
+// `deleteUserSessions()` ran. That row predates `deadline` (captured once the change
+// is known committed), so this catches it while sparing anything created afterwards.
+// Exported for direct testing; the service schedules it via setTimeout.
+export async function sweepRacedSessionsAfterPasswordChange(
```
mp-blurify, line 99: Four positionals, two of them adjacent interchangeable strings — swapping `userId` and `sparedToken` typechecks and silently deletes nothing (or everything) at the call sites in `identity.service.ts:1964-1968` and `identity.service.int.test.ts:1076-1081`.

Rule: `docs/standards/functions.md` — more than 3 parameters → a single named-object param; a leading handle like `tx` may stay positional.
Fix: `sweepRacedSessionsAfterPasswordChange(db, { userId, sparedToken, deadline })`.

packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  bootTestApp,
+  registerPlayer,
+  asPlayer,
+  seedMinimal,
+  type TestDb,
+  type TestApp,
+} from '../index.js';
+
+/**
+ * The two halves of BF-521's "invalidate all other sessions on password change" that
```
mp-blurify: This comment names a downstream client's ticket id in a public repo. `tools/lint/verify-oss-hygiene.ts:32` matches that pattern against every git-tracked `.ts` file, so `pnpm check:hygiene` — a CI gate `pnpm verify` does not cover — fails on this line.

Rule: `docs/standards/testing.md` — a downstream client's tracker key does not belong in a file name, comment, or `describe()`/`it()` title.
Fix: drop the id and describe the behaviour instead ("the two halves of the password-change session revoke that only show up end to end").

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 75923583. `sweepRacedSessionsAfterPasswordChange` now returns the ids it deleted (`.returning({ id: session.id })`); `changePassword's setTimeout callback emits `identity.sessions.revoked_all` (same audit + SSE path as the primary revoke) when that list is non-empty, sparing the still-legitimate rotated session by id. Covered by a new int test ("emits a second revoked_all when the deferred sweep force-deletes a raced session") that seeds a session landing right after the mocked revoke and waits out the real delay.

packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts
jakubfilinger-b: Fixed in 75923583. The first e2e case now pulls the rotated `Set-Cookie` off the change-password response and asserts it authenticates against `/identity/security/me` (200) while the acting tab's pre-change cookie no longer does (401) - so "the acting session survives" is actually exercised, not just "no push arrived".

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 75923583. The router now resolves `getUserId(context)` / `requireSessionId(context)` (already populated by the request-level session resolve) and passes both into `changePassword` as a `caller` param; the handler no longer calls `currentSession(headers)` itself. `currentSession`/`currentUserId` stay as-is for the dozen other handlers that still only have raw headers to work with.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 75923583 - `sweepRacedSessionsAfterPasswordChange(db, { userId, sparedToken, deadline })`. `db` stays positional as the leading handle.

packages/testing/src/__tests__/change-password-session-revoke.e2e.test.ts
jakubfilinger-b: Fixed in 75923583 - dropped the ticket id from the docblock, described the behaviour instead. Confirmed `pnpm check:hygiene` passes clean on the branch now.

jakubfilinger-b:
Pushed 75923583, addressing today's round of review:

- **Blocker — silent sweep**: fixed. The deferred sweep's deletions now emit `identity.sessions.revoked_all` (with audit) when it actually deletes something, sparing the legitimate rotated session by id. New int test covers the full wiring, not just the pure sweep function.
- **Blocker — surviving session unasserted**: fixed. The e2e now proves the acting tab's rotated cookie authenticates and the pre-change one 401s, not just that no push arrived.
- Dropped the redundant `getSession` round trip in `changePassword` (router now passes the already-resolved `caller`).
- `sweepRacedSessionsAfterPasswordChange` takes a named object param instead of 4 positionals.
- Dropped the client ticket id from the e2e docblock (`check:hygiene` was failing on it).
- Rebased onto latest `dev` — picked up `docs/catalog.json` moving to generated/gitignored, resolved cleanly.

Full local gate green: types, lint, format, boundaries, shape, hygiene, drift, and both test suites (identity int 78/78, e2e 3/3).

Requesting re-review.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
+            if (deletedIds.length === 0) {
+              return;
+            }
+            this.events.emit('identity.sessions.revoked_all', {
+              userId,
+              playerId,
+              actorId: userId,
+              ip: ip ?? null,
+              userAgent: userAgent ?? null,
+              ...(rotated ? { exceptSessionId: rotated.id } : {}),
```
damianrzepka, line 1992: The second identity.sessions.revoked_all reintroduces the self-revoke that the first event correctly avoids. An already-open stream for the initiating tab captured the pre-rotation sessionId; here exceptSessionId is the freshly rotated row id, so it cannot match. When the sweep actually deletes a raced row, the initiating tab receives revoked two seconds later and logs out despite holding a valid rotated cookie. Keep the original connection session id available to this callback (or support an exclusion set covering both identities), and cover the sweep interleaving with a live initiating stream.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
+      exceptSessionId: sessionId,
+    });
+    // better-auth rotates the password hash and deletes sessions in separate,
+    // untransacted statements; a sign-in that verified the OLD password just before
+    // the rotation can still insert its session row right after. Re-sweep shortly
+    // after to delete it, sparing the caller's fresh session by token. Skipped when
+    // the rotated token is unknown - without it the sweep cannot tell the caller's
+    // own new session apart. Residual exposure: RACED_SESSION_SWEEP_DELAY_MS.
+    if (rotatedToken) {
+      const timer = setTimeout(() => {
```
damianrzepka, line 1973: The security cleanup is an unref process-local timer. A restart, deploy, or scale-down during this two-second delay drops it permanently, leaving a raced old-password session live indefinitely; the catch only logs and cannot retry. This is scheduled work that needs durable, idempotent execution through JOB_QUEUE (or a durable session-generation invariant), per the messaging-and-microservices rule.

packages/core/src/pam/identity/router/index.ts
```diff
@@ -132,12 +132,19 @@ export function createIdentityRouter(
 
     streamSession: os.streamSession.handler(({ signal, context }) => {
       const userId = getUserId(context);
+      const connectionSessionId = getSessionId(context);
       return createEventStreamGenerator(
         (push) => {
           const unsubscribeRevoked = eventBus.on('identity.sessions.revoked_all', (event) => {
```
damianrzepka, line 138: This uses the domain-event broker as the client-push transport. With the shipped Redis Streams broker, replicas share the service consumer group, so one identity.sessions.revoked_all event is consumed by only one replica. Streams connected to sibling replicas therefore never receive this revoke and the claimed immediate cross-tab logout fails in a normal multi-replica deployment. Send the user-scoped signal through REALTIME_TRANSPORT, the required broadcast seam for server-to-client push, and add a multi-replica transport-level test.

damianrzepka (changes requested):
NO-GO — the latest revision still has a self-revocation regression when the deferred sweep runs, and its race cleanup is neither durable nor broadcast correctly across replicas.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
     this.forwardCookies(res, resHeaders);
-    if (userId) {
+    // Captured once the rotation is known committed - the deferred sweep below deletes
+    // any session row created at or before this instant (i.e. one a concurrent
+    // old-password sign-in raced in), sparing anything created afterwards.
+    const sweepDeadline = new Date();
+    // The token of the session better-auth just minted for the caller. `ensureOk` only
+    // reads the body on failure, so this read is safe here.
+    const rotatedToken =
+      ((await res.json().catch(() => ({}))) as { token?: string | null }).token ?? null;
```
damianrzepka, line 1979: This `as { token?: string | null }` cast on `res.json()` is a legitimate third-party type-inference boundary (fetch's `Response.json()` is untyped), but per `docs/standards/types.md` that sanctioned exception needs a one-line `// Library boundary:` note to read as intentional rather than an unmarked compiler-silencing cast.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
+      ip: ip ?? null,
+      userAgent: userAgent ?? null,
+    });
+    // Same event the admin "revoke all sessions" path emits, so the streamSession
+    // SSE handler pushes { type: 'revoked' } to this user's OTHER tabs.
+    // exceptSessionId is the caller's PRE-rotation session id - it's the id the
+    // caller's already-open streamSession connection captured at stream-open, and the
+    // only value its SSE compare can ever match, so this is what spares the
+    // initiating tab.
+    this.events.emit('identity.sessions.revoked_all', {
```
damianrzepka, line 2028: This hand-rolls the same `identity.sessions.revoked_all` shape that `SessionService.revokeAllSessions` (session.service.ts:204-218) already builds and emits, just without an except-session id. Might be worth adding an optional `exceptSessionId` param to the shared method instead of a one-off copy here - `disableTwoFactor` (line 1811) still calls `revokeAllSessions` directly and can't spare the caller's own session, the same gap this PR closes for changePassword only.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
       await this.drizzle.db
-        .update(user)
-        .set({ passwordMeetsPolicy: true })
-        .where(eq(user.id, userId));
+        .update(session)
+        .set({ expiresAt: new Date(Date.now() + DONT_REMEMBER_SESSION_TTL_MS) })
+        .where(eq(session.id, rotated.id));
+    }
+    await this.drizzle.db
+      .update(user)
```
damianrzepka: `update(user).set({ passwordMeetsPolicy: true })` and `getPlayerIdByUserIdSafe` below don't depend on each other or on `rotated` - could run via `Promise.all` to shave a round trip off this rate-limited, user-facing endpoint. (The session TTL update above this does depend on `rotated`, so that one has to stay sequential.)

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1813,26 +1879,124 @@ export class IdentityService {
...
+          deadline: sweepDeadline,
+        })
+          .then((deletedIds) => {
+            // A session the sweep just force-deleted got no push and no audit record
+            // otherwise: it raced in after the primary revoked_all above already
+            // fired, so this is the only notice it gets of either.
+            if (deletedIds.length === 0) {
+              return;
+            }
+            this.events.emit('identity.sessions.revoked_all', {
```
damianrzepka, line 2056: If `rotated` is `undefined` (the session-by-token lookup missed) at the point this sweep actually deletes raced sessions, this emit goes out with no `exceptSessionId`, revoking the caller's own already-rotated legitimate session too. Narrow (race-on-a-race), but maybe worth a log line when `rotated` is missing here so it's visible if it ever fires.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
```diff
@@ -944,6 +949,203 @@ describe('IdentityService.resetPassword', () => {
...
+      .returning({ id: session.id, token: session.token })
+      .then(([row]) => row!);
+
+  it('delegates the session revoke to better-auth and emits identity.sessions.revoked_all', async () => {
+    const account = await seedUser();
+    const current = await seedChangePwSession(account.id);
+    const otherA = await seedChangePwSession(account.id);
+    const otherB = await seedChangePwSession(account.id);
+
+    // createAuth is mocked in this suite, so better-auth's own revokeOtherSessions
```
damianrzepka: `docs/standards/comments.md`: "Never in tests - a test name states the behaviour and the assertions state the evidence." This narrates why the mock stands in for better-auth; worth folding into the `it(...)` description or a helper name instead.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
```diff
@@ -944,6 +949,203 @@ describe('IdentityService.resetPassword', () => {
...
+      { userId: account.id, sessionId: current.id },
+    );
+
+    expect(result).toEqual({ success: true });
+    expect(changePasswordMock).toHaveBeenCalledWith(
+      expect.objectContaining({
+        body: expect.objectContaining({ revokeOtherSessions: true }),
+      }),
+    );
+    // Sanity check on the stand-in only: better-auth is mocked here, so this re-reads
```
damianrzepka: Same "never in tests" rule - this comment explains what's covered elsewhere (the e2e test). Worth trimming per `docs/standards/comments.md`.

damianrzepka (commented):
A few notes from review. The main mechanism (sparing the acting session via `caller.sessionId` captured pre-rotation) checks out correctly against the `streamSession` SSE handler's `connectionSessionId` - nice fix. Left inline comments on smaller items below.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 406f6d49. `exceptSessionId` on the sweep's second emit is now the caller's PRE-rotation `sessionId` (same value the primary emit above uses), not `rotated.id`. New int test assertion updated to match - it previously asserted the buggy `rotated.id` value, which is exactly why this slipped through.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Agreed this is a real gap, and distinct from the domain-event/EventBus point below. Grouping it with the already-acknowledged durable-invariant follow-up rather than folding it into this PR: both are about hardening the sweep's execution guarantees (this PR keeps the 2s deferred-`setTimeout` mitigation rather than a fully durable/transactional one), and it's the same auth-hot-path surface we already agreed deserves its own change with its own review. Will link the ticket here once filed alongside the session-generation invariant.

packages/core/src/pam/identity/router/index.ts
jakubfilinger-b: Fixed in 406f6d49 - good catch, this one's grounded directly in docs/standards/messaging-and-microservices.md's own channel table ("server-to-client push -> REALTIME_TRANSPORT", not the domain EventBus). `streamSession` was subscribing to EVENT_BUS directly, which under the shipped Redis Streams broker is exactly the one-consumer-group-per-SERVICE_NAME behavior you describe.

Mirrored the existing compliance/plugin.ts KYC-status pattern: the plugin now subscribes to `identity.sessions.revoked_all` / `identity.user.unlocked` once (durable, cross-replica-delivered-once via the consumer group) and republishes onto `REALTIME_TRANSPORT` on a per-user channel (`sessionEventsChannel`, Redis Pub/Sub - fans out to every replica). The router now subscribes to that channel instead of the EventBus directly. Covered by the existing change-password-session-revoke e2e (bootTestApp binds the same RedisPubSubRealtimeTransport-shaped transport production uses) plus the full identity int suite, both green.

jakubfilinger-b:
Pushed 406f6d49, addressing damianrzepka's round of review:

- **Self-revoke regression**: fixed. The deferred sweep's second `identity.sessions.revoked_all` emit (added in 75923583) used the rotated (new) session id as `exceptSessionId` - but the acting tab's SSE stream captured the pre-rotation id at connect time, so the two could never match. A raced-session sweep would have force-logged-out the legitimate acting tab, the exact failure mode this PR exists to prevent. Now uses the caller's pre-rotation `sessionId`, same as the primary emit. The int test that encoded the bug is updated to assert the correct id.
- **Cross-replica push**: fixed. `streamSession` subscribed to the domain `EVENT_BUS` directly; under the shipped Redis Streams broker that's a per-`SERVICE_NAME` consumer group, so only one replica ever received `identity.sessions.revoked_all` - a tab connected to any other replica never got the push. Per `docs/standards/messaging-and-microservices.md`'s own channel table, server-to-client push belongs on `REALTIME_TRANSPORT`. Mirrored the existing `compliance/plugin.ts` KYC-status pattern: the plugin subscribes to the domain events once and republishes onto a per-user `REALTIME_TRANSPORT` channel (Pub/Sub, fans out to every replica); the router subscribes to that channel instead.
- **Durable sweep execution (JOB_QUEUE)**: acknowledged as a real gap, grouped with the already-deferred session-generation invariant rather than folded into this PR - same auth-hot-path surface, own ticket.

Gate green: types, lint, format, boundaries, shape, hygiene, drift; identity int suite (83) and the e2e suite (182 across the package) all passing.

Requesting re-review.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Already covered by 406f6d49 (pushed just after this review landed) - the emit no longer depends on `rotated` at all. It now passes `exceptSessionId: sessionId` unconditionally, i.e. the caller's PRE-rotation id from `caller.sessionId`, which is a required (non-optional) param. So there's no longer a path where this emit goes out without an except-id, and the id it carries is the one the acting tab's open SSE connection actually holds.

(`rotated` is still used above for the `dont_remember` TTL clamp, where a miss is correctly a no-op.)

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Looked at sharing it and it doesn't fit cleanly - `revokeAllSessions` isn't just the emit, it also does the DB work:

```
update(session).set({ expiresAt: now() })
  .where(and(eq(session.userId, userId), gt(session.expiresAt, now())))
```

By the time `changePassword` emits, better-auth has already deleted the old rows and minted the caller's replacement, so calling the shared method here would expire that freshly-minted rotated session and log the caller out - the exact failure this PR exists to prevent.

An `exceptSessionId` param wouldn't resolve it either, because the two ids differ: the row the DB update must spare is the NEW rotated id, while the value the emit must carry is the PRE-rotation id (the only one an already-open `streamSession` connection holds). One param can't be both, and a "skip the DB work" flag reads worse than the explicit emit does. What's left as actual duplication is a 6-field payload whose shape `domainEventSchemas` already enforces, so a helper would add indirection without removing a failure mode.

The `disableTwoFactor` observation is a real and separate finding though - it revokes every session including the caller's own, so a player disabling 2FA logs themselves out. Out of scope here (this PR's contract is changePassword), but worth its own ticket; happy to file it.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 77d5d699 - added the `// Library boundary:` note naming why the cast exists (`Response.json()` resolves to `any` and better-auth publishes no response type for this route), so it reads as the sanctioned exception in `docs/standards/types.md` rather than an unmarked cast.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed in 77d5d699 - the `passwordMeetsPolicy` update and `getPlayerIdByUserIdSafe` now go out together under `Promise.all`, with a note that the TTL clamp above stays sequential because it depends on `rotated`. Good call on a rate-limited, user-facing path.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
jakubfilinger-b: Fixed in 77d5d699 - the narration is gone, folded into a named helper instead: `stubBetterAuthDeletingEverySessionAndMintingOne(userId, rotatedToken, duringRotation?)`. The name carries what the comment used to say, and the two tests that stubbed this inline now share it. The raced-session insert in the sweep test became `seedOldPasswordSignInRacingIn(userId)` passed as `duringRotation`, which retired that test's narrating comment too.

packages/core/src/pam/identity/__tests__/identity.service.int.test.ts
jakubfilinger-b: Fixed in 77d5d699. Rather than just deleting it (mp-blurify had asked for that pointer in an earlier round, so the information does need to survive), folded it into the name: `remaining` is now `rowsLeftByTheBetterAuthStub`, so the assertion states on its own that it re-reads the stub's writes. The `it(...)` title already says "delegates the session revoke to better-auth", which carries the rest.

jakubfilinger-b:
Pushed 8fb9ffb3 (rebased on latest `dev`), covering @damianrzepka's latest six notes.

**Fixed (77d5d699):**
- **Library-boundary cast** — the `res.json()` assertion now carries the `// Library boundary:` note naming why it exists, per `docs/standards/types.md`.
- **Two round trips** — `passwordMeetsPolicy` and the `playerId` lookup now run under one `Promise.all`; the TTL clamp stays sequential since it depends on `rotated`.
- **Narrating test comments** — both folded into names rather than deleted: a shared `stubBetterAuthDeletingEverySessionAndMintingOne(userId, rotatedToken, duringRotation?)` helper (also used by the sweep test, via `seedOldPasswordSignInRacingIn`), and `remaining` → `rowsLeftByTheBetterAuthStub`.

**Already addressed:**
- **`rotated` undefined on the sweep emit** — resolved by 406f6d49, which landed just after your review. That emit no longer reads `rotated` at all: it passes `exceptSessionId: sessionId` unconditionally (the caller's pre-rotation id, a required param), so there's no path where it goes out without an except-id.

**Discussed, no change:**
- **Sharing `revokeAllSessions`** — doesn't fit: that method also expires every live session for the user, which here would kill the caller's freshly-minted rotated row. An `exceptSessionId` param wouldn't bridge it either, since the row the DB must spare is the *new* id while the emit must carry the *pre-rotation* id. Details in the thread. Your `disableTwoFactor` observation is a real separate bug though (it logs the caller out of their own session) — worth its own ticket, happy to file it.

Gate green serially (`TURBO_CONCURRENCY=1`): types, lint, format, boundaries, shape, hygiene, deprecations, drift, unit + both integration tiers. Note for anyone reproducing locally — running the core and testing integration tiers concurrently saturates a single local Postgres/Redis and produces moving, unrelated failures (notifications stream timeouts, bonus-rollover, tag concurrency); serial is clean.

Requesting re-review.

jakubfilinger-b:
Rebased onto latest `dev`, which merged #158 (per-player auto-logout / 2FA on every login) into the same handler. Flagging one thing that grew the diff, since it wasn't in the original scope:

**`identity.session.revoked` now rides the same realtime transport.** #158 added a third push to `streamSession` — the single-session revoke that `SessionIdleService`'s idle expiry fires — subscribed directly on the domain `EventBus`. That's the same channel-choice bug this PR just fixed for `revoked_all`: under the shipped Redis Streams broker only one replica consumes each event, so a tab on any other replica never gets told it was logged out. Leaving it on the EventBus would also have meant keeping `eventBus` as a router param purely for that one subscription, right next to the transport that replaced it.

So all three pushes now go through `sessionEventsChannel`. The internal payload gained a discriminant so each connection can still decide for itself:

```ts
export type SessionEventsPush =
  | { type: 'revoked-all'; exceptSessionId?: Session['id'] }
  | { type: 'session-revoked'; sessionId: Session['id'] }
  | { type: 'unlocked' };
```

The wire contract the client sees is unchanged (`{ type: 'revoked' | 'unlocked' }`).

That push had no test — #158 added it without one — so 5583fbef covers it end to end in the existing bootTestApp e2e: two live streams, revoke one session by id via `POST /identity/sessions/me/revoke`, assert the named tab gets `revoked` and the other stays silent. 4/4 green there, 191 identity int tests green, full serial gate otherwise clean.

Happy to split the `session-revoked` routing into its own PR if you'd rather review it separately — it only landed here because the merge left the two halves inconsistent.

zaxovaiko (approved):

PR #151, opened 2026-09-08
test: cut the suite's wall clock and tighten what it asserts

zaxovaiko:
## Summary

Cuts the test suite's wall clock and tightens what it asserts, without adding a tool or a framework. The core integration tier goes from 90s to 29s in CI, the e2e tier from 100s to 54s. Alongside that, one worthless test is removed, eight assertions that could not observe the behaviour they were named for are fixed, and five untested branches are covered.

## Why

Two questions were asked of the suite: how fast is it, and is any of it worthless.

On speed, the measurements pointed at one thing. Both integration tiers spend most of their time waiting on Postgres commits, against a database holding nothing but throwaway data. A control run on a fresh default `postgres:16-alpine` took 80.1s for the core tier; the same container with `fsync`, `full_page_writes` and `synchronous_commit` off took 28.8s. All three are reloadable settings, so CI needs a step rather than a rewrite of the services block.

That left the e2e tier as the longest leg. It was serialised because all 19 suites shared one database - and nothing actually kept them out of each other's rows, since `truncateAll` existed for that job but no suite ever called it. Each suite now clones a template migrated once per run (a ~50ms `CREATE DATABASE ... TEMPLATE`, against ~250ms of migration checks it was already paying), and two run at a time.

On content, 177 test files and 2268 tests were read. The suite is in good shape; there was no slop to harvest. What turned up instead was a handful of tests whose names promised more than their assertions could see - `kyc-status-writer`'s "silent no-op" case had no assertion at all, while the null it returns is what compliance reads to decide whether to emit a KYC transition event. Each strengthened assertion was verified by breaking the line it covers and confirming the test fails.

Two flakes surfaced on the gate while this was being built, both real and both pre-existing:

- `bootTestApp` handed back an app before its Redis client had connected. The login limiter is `onUnavailable: 'deny'`, so a first request that beat the handshake was answered with a fail-closed 429 rather than reaching the route. Only reproducible under load, which is exactly when the handshake is slowest.
- The RG deposit-limit cases stamped `createdAt` from the app clock while the spend window's upper bound is `now()` on the database. Measured on one machine, the host runs ~43ms ahead of the Postgres container, which is enough to drop a just-seeded deposit out of the window.

## Alternatives considered

**Splitting `chat.service.int.test.ts`.** At 3042 lines and 139 tests it was the critical path of the core tier - 90s of a 90s wall. The durability change removed that: it is now 13.1s inside a 29s tier, so splitting it would buy roughly 3s for a 3000-line diff. Not done.

**Widening Redis past 16 logical databases.** Would allow more than two e2e workers, but `databases` is a startup-only flag and a GitHub Actions service container cannot be passed a command, so it would mean replacing the service with a `docker run` step. Runners have 4 cores, so it buys nothing there.

**Coverage tooling.** A percentage decides nothing here. `pnpm report:route-coverage` names the actual gap instead: 39 of 259 routes are referenced by no test, 21 of them chat moderation and room-rule routes, 12 identity session and admin-security routes. It is a report, not a gate - it stays out of `pnpm verify` and always exits 0.

**Turning that report into a gate.** Would mean either writing 39 tests first or carrying a baseline file. Neither is worth it to state something the report already states plainly.

## Risks

The Postgres durability change applies to CI only; the local dev cluster is untouched, because there the dev and test databases share one cluster and `ALTER SYSTEM` is cluster-wide. The data it affects is recreated per run, so there is nothing a lost flush could cost.

`maxWorkers: 2` on the e2e tier is bounded by two numbers that must move together: this tier owns 8 of Redis's 16 logical databases, and one suite keeps three booted apps alive at once, so a worker needs a slice of 4. Both are commented where they are set. Raising the worker count without widening Redis makes two suites flush each other's keys.

`LOG_LEVEL=silent` discards the output rather than using pino's own silent level. That is deliberate: a disabled level never reaches `logMethod`, where the error-reporter hook lives, so the obvious implementation would have made a request for quiet logs silently stop reporting errors to the bound tracker. The logger's own test covers that hook and is what caught it.

Test databases are dropped by prefix in the global teardown rather than by each suite, since only 2 of the 19 dispose theirs. A killed run leaves clones behind until the next run sweeps them.

### DIFF
--- a/packages/core/src/compliance/__tests__/rg-monitoring.service.int.test.ts
+++ b/packages/core/src/compliance/__tests__/rg-monitoring.service.int.test.ts
@@ -64,8 +64,12 @@ async function seedPlayer(userId: string, currency: string) {
   await db.drizzle.db.insert(player).values({ userId, currency, kycStatus: 'verified' });
 }
 
-const seedDeposit = (userId: string, amount: string, createdAt = new Date()) =>
-  seedCompletedDeposit(db, userId, amount, { createdAt });
+// No createdAt unless a case wants a backdated one: the column defaults to the database
+// clock, and the spend window's upper bound is `now()` on that same clock. An app-side
+// `new Date()` a millisecond ahead of the database - the host clock running ahead of the
+// container's is enough - puts the deposit past the window and out of the player's spend.
+const seedDeposit = (userId: string, amount: string, createdAt?: Date) =>
+  seedCompletedDeposit(db, userId, amount, createdAt ? { createdAt } : {});
 
 async function seedBet(userId: string, betAmount: string, winAmount = '0') {
   const [g] = await db.drizzle.db
--- a/packages/core/src/compliance/__tests__/rg-monitoring.service.int.test.ts
+++ b/packages/core/src/compliance/__tests__/rg-monitoring.service.int.test.ts
@@ -64,12 +64,8 @@ async function seedPlayer(userId: string, currency: string) {
   await db.drizzle.db.insert(player).values({ userId, currency, kycStatus: 'verified' });
 }
 
-// No createdAt unless a case wants a backdated one: the column defaults to the database
-// clock, and the spend window's upper bound is `now()` on that same clock. An app-side
-// `new Date()` a millisecond ahead of the database - the host clock running ahead of the
-// container's is enough - puts the deposit past the window and out of the player's spend.
-const seedDeposit = (userId: string, amount: string, createdAt?: Date) =>
-  seedCompletedDeposit(db, userId, amount, createdAt ? { createdAt } : {});
+const seedDeposit = (userId: string, amount: string, backdatedTo?: Date) =>
+  seedCompletedDeposit(db, userId, amount, backdatedTo ? { createdAt: backdatedTo } : {});
 
 async function seedBet(userId: string, betAmount: string, winAmount = '0') {
   const [g] = await db.drizzle.db
--- a/packages/testing/src/app.ts
+++ b/packages/testing/src/app.ts
@@ -131,8 +131,7 @@ export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
     // A half-booted app still owns a Postgres pool, Redis connections and the claimed
     // logical database. Left behind, the next suite to claim that database inherits its
     // keys, and the run leaks a connection per failed boot.
-    await created?.close();
-    await redisDatabase.release();
+    await Promise.allSettled([created?.close(), redisDatabase.release()]);
     throw err;
   }
 
--- a/packages/testing/src/app.ts
+++ b/packages/testing/src/app.ts
@@ -1,4 +1,5 @@
 import { randomUUID } from 'node:crypto';
+import { once } from 'node:events';
 import { fileURLToPath } from 'node:url';
 import {
   createApp,
@@ -24,6 +25,8 @@ import { user, session, account, verification, twoFactor } from '@openora/core/p
 import { SseClientAuthorizer } from '@openora/core/testing';
 import { acquireTestRedisDatabase } from './redis.js';
 import type { Hono } from 'hono';
+
+const REDIS_READY_TIMEOUT_MS = 10_000;
 
 export type TestApp = {
   /** The Hono app - drive it directly with `app.request(path, init)`. */
@@ -64,6 +67,7 @@ export type BootTestAppConfig = Pick<CreateAppConfig, 'plugins' | 'igaming'> & {
 export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
   const redisDatabase = await acquireTestRedisDatabase();
   const serviceName = `test-${randomUUID()}`;
+  let client: ReturnType<typeof createRedisClient> | undefined;
 
   const created = await createApp(
     {
@@ -84,6 +88,7 @@ export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
     },
     (container: Container<CoreTokenCatalog>) => {
       const redis = createRedisClient(redisDatabase.url);
+      client = redis;
       container.onDispose(() => redis.close());
 
       container.register(MESSAGE_BROKER, () => {
@@ -111,6 +116,15 @@ export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
       }
     },
   );
+
+  // `createRedisClient` fires connect() without awaiting it - right in production, where
+  // boot must not block on Redis, wrong for a test that issues its first request straight
+  // after this returns. The login limiter is `onUnavailable: 'deny'`, so a request that
+  // beats the handshake is answered with a fail-closed 429 rather than reaching the route.
+  // Only reproducible under load, which is exactly when the handshake is slowest.
+  if (client && !client.isReady) {
+    await once(client, 'ready', { signal: AbortSignal.timeout(REDIS_READY_TIMEOUT_MS) });
+  }
 
   return {
     app: created.app,
--- a/packages/testing/src/app.ts
+++ b/packages/testing/src/app.ts
@@ -69,61 +69,71 @@ export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
   const serviceName = `test-${randomUUID()}`;
   let client: ReturnType<typeof createRedisClient> | undefined;
 
-  const created = await createApp(
-    {
-      plugins: [
-        {
-          id: 'testing-registration-config',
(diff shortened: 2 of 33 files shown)


packages/testing/src/global-setup.ts
```diff
@@ -0,0 +1,38 @@
...
+  return async () => {
+    await dropClones(admin);
+    await admin.end();
+  };
+}
+
+async function dropClones(admin: Pool): Promise<void> {
+  const { rows } = await admin.query<{ datname: string }>(
+    `SELECT datname FROM pg_database WHERE datname LIKE $1`,
+    [`${TEMPLATE_DATABASE}_%`],
```
klaudia-blazyczek-blurify: This cleanup is shared by every run that uses the same Postgres server. A second test run creates databases with the same `oss_igaming_test_tpl_` prefix, and this query can drop them before its own run or while they are still in use. Use a run-specific prefix/template and clean only that run's databases. This breaks the deterministic and isolated test requirement in `docs/standards/testing.md`.

packages/testing/src/app.ts
```diff
@@ -112,6 +117,15 @@ export async function bootTestApp(config: BootTestAppConfig): Promise<TestApp> {
     },
   );
 
+  // `createRedisClient` fires connect() without awaiting it - right in production, where
+  // boot must not block on Redis, wrong for a test that issues its first request straight
+  // after this returns. The login limiter is `onUnavailable: 'deny'`, so a request that
+  // beats the handshake is answered with a fail-closed 429 rather than reaching the route.
+  // Only reproducible under load, which is exactly when the handshake is slowest.
+  if (client && !client.isReady) {
+    await once(client, 'ready', { signal: AbortSignal.timeout(REDIS_READY_TIMEOUT_MS) });
```
klaudia-blazyczek-blurify: If Redis readiness times out or emits an error, this function rejects without calling `created.close()` or `redisDatabase.release()`. The Postgres/Redis clients and the claimed Redis database are then leaked, which can make later suites collide or hang. Wrap this boot path in cleanup before rethrowing. See `docs/standards/testing.md`.

packages/testing/src/global-setup.ts
```diff
@@ -0,0 +1,38 @@
...
+export default async function setup(): Promise<() => Promise<void>> {
+  const admin = new Pool({ connectionString: adminUrl(), connectionTimeoutMillis: 5000 });
+
+  await dropClones(admin);
+  await admin.query(`DROP DATABASE IF EXISTS "${TEMPLATE_DATABASE}" WITH (FORCE)`);
+  await admin.query(`CREATE DATABASE "${TEMPLATE_DATABASE}"`);
+  await applyMigrations(urlForDatabase(TEMPLATE_DATABASE));
+
+  return async () => {
+    await dropClones(admin);
```
klaudia-blazyczek-blurify: When `dropClones`, `CREATE DATABASE`, or `applyMigrations` fails, this `admin` pool is never closed because only the returned teardown calls `admin.end()`. A failed global setup can leave a live database connection behind. Close the pool in a `catch`/`finally` path before rethrowing.

packages/core/src/compliance/__tests__/rg-monitoring.service.int.test.ts
```diff
@@ -64,8 +64,12 @@ async function seedPlayer(userId: string, currency: string) {
   await db.drizzle.db.insert(player).values({ userId, currency, kycStatus: 'verified' });
 }
 
-const seedDeposit = (userId: string, amount: string, createdAt = new Date()) =>
-  seedCompletedDeposit(db, userId, amount, { createdAt });
+// No createdAt unless a case wants a backdated one: the column defaults to the database
```
klaudia-blazyczek-blurify: Please remove the explanatory comments added inside tests. The repository convention says comments should not be added in tests; keep the behavior in the test name and assertions instead. The same pattern appears in several changed test files. See `.rulesync/rules/conventions.md`.

mp-blurify:
## Code review — findings

All three tiers were run locally against real Postgres/Redis — core unit (72 files / 674 tests), core integration (87 files / 1565 tests), testing e2e (19 files / 166 tests). All green, no flakes observed.

### Blocking
None. Every removed or reworded test was traced back to the production code it covered; no coverage silently disappeared and no assertion was net-loosened. The one deletion — `user-commands.int.test.ts` "lets other unique violations through untouched" — is confirmed worthless as claimed: it asserted a raw Postgres unique-constraint throw and never called the port's `setUsername` catch/narrow logic it was named for.

### Worth fixing
- **`docs/standards/testing.md:20`** — still states that integration Vitest configs run with `poolOptions.threads.singleThread = true` and that "the harness shares one test database". After this PR that is untrue (per-suite cloned databases, `maxWorkers: 2`, no `fileParallelism: false`), so a reader following this doc misunderstands the current isolation model. `pnpm check:drift` will not catch it — that gate only diffs `docs/catalog.json` and `docs/platform/system-design.md`.

### Note
`packages/core/src/server/kernel/logger.ts` is production code changed so `LOG_LEVEL` is actually honored (it was previously never read by `pino()`), letting tests request silence without breaking the error-reporter hook. Backward compatible when `LOG_LEVEL` is unset, and `logger.test.ts` exercises the discard branch implicitly now that the core unit tier runs under ambient `LOG_LEVEL=silent`.

klaudia-blazyczek-blurify (changes requested):

packages/testing/src/global-setup.ts
```diff
@@ -0,0 +1,48 @@
...
+export default async function setup(): Promise<() => Promise<void>> {
+  process.env['OPENORA_TEST_RUN_ID'] = randomUUID().replaceAll('-', '').slice(0, 8);
+  const template = templateDatabase();
+  const admin = new Pool({ connectionString: adminUrl(), connectionTimeoutMillis: 5000 });
+
+  try {
+    await admin.query(`CREATE DATABASE "${template}"`);
+    await applyMigrations(urlForDatabase(template));
+  } catch (err) {
+    await admin.end();
```
klaudia-blazyczek-blurify, line 30: If `CREATE DATABASE` succeeds but `applyMigrations` fails, this catch only closes `admin`. Because global setup rejects, Vitest never receives the teardown function, so the run-specific `template` database is left behind. Drop `template` in this failure path before rethrowing. This violates the cleanup and isolation requirement in `docs/standards/testing.md`.

packages/testing/src/app.ts
```diff
@@ -64,53 +67,74 @@ export type BootTestAppConfig = Pick<CreateAppConfig, 'plugins' | 'igaming'> & {
...
+    // beats the handshake is answered with a fail-closed 429 rather than reaching the route.
+    // Only reproducible under load, which is exactly when the handshake is slowest.
+    if (client && !client.isReady) {
+      await once(client, 'ready', { signal: AbortSignal.timeout(REDIS_READY_TIMEOUT_MS) });
+    }
+  } catch (err) {
+    // A half-booted app still owns a Postgres pool, Redis connections and the claimed
+    // logical database. Left behind, the next suite to claim that database inherits its
+    // keys, and the run leaks a connection per failed boot.
+    await created?.close();
```
klaudia-blazyczek-blurify: If `created?.close()` rejects while handling a boot failure, `redisDatabase.release()` is skipped and the original boot error is replaced by the cleanup error. Use `try/finally` or `Promise.allSettled` so both the app and the claimed Redis database are always cleaned up. This is required for deterministic test isolation in `docs/standards/testing.md`.

klaudia-blazyczek-blurify (changes requested):

packages/testing/src/global-setup.ts
```diff
@@ -0,0 +1,49 @@
...
+export default async function setup(): Promise<() => Promise<void>> {
+  process.env['OPENORA_TEST_RUN_ID'] = randomUUID().replaceAll('-', '').slice(0, 8);
+  const template = templateDatabase();
+  const admin = new Pool({ connectionString: adminUrl(), connectionTimeoutMillis: 5000 });
+
+  try {
+    await admin.query(`CREATE DATABASE "${template}"`);
+    await applyMigrations(urlForDatabase(template));
+  } catch (err) {
+    await dropRunDatabases(admin, template).catch(() => {});
```
klaudia-blazyczek-blurify: This empty catch hides a failed cleanup. If `dropRunDatabases` cannot connect or a `DROP DATABASE` fails, the template can still leak and the test run reports only the migration error. Keep the original error, but log the cleanup error with context before rethrowing it. Silent catches are forbidden by `docs/standards/errors.md`.

klaudia-blazyczek-blurify (approved):

PR #152, opened 2026-09-09
feat(compliance): country rules and global KYC toggle

damianrzepka:
## Summary

Rewrites the old binary allow/block geo rule into three independent per-country flags (blacklisted, redirectIp, kycRequired) backing the Regulatory Overview Backoffice screen, and adds a new admin-editable, audited global KYC toggle. Config surface only - BF-534.

## Why

The Regulatory Overview screen (Confluence: Regulatory Overview) replaces the platform's single blocked-country list with three independent settings plus a platform-wide KYC toggle. The acceptance criteria require country-rule access to be limited to Super Admin and Compliance Manager only, and every rule change to be audited by actor, country, setting, previous value, new value, and timestamp.

## Alternatives considered

- Gating the new routes on the existing `compliance` RBAC resource - rejected: `compliance:read_write` is already held by `risk-fraud-manager` and `kyc-aml-officer` (for unrelated override-limit/manage-rg actions), and a `read_write` grant unlocks every action on a resource, so this would leak country-rule access to those roles. A new `regulatory-overview` resource, granted only to `super-admin` (automatic) and `compliance-manager` (explicit), is the only way to meet the "Super Admin and Compliance Manager only" requirement.
- One audit row per mutation with a JSON before/after blob (matches the existing `addGeoRule` pattern) - rejected in favor of one audit row per changed boolean setting, since the acceptance criteria explicitly ask for actor/country/**setting**/previous/new per change.

## Risks

- Migration renames `geo_rule` to `country_rule` and drops the `action` column after backfilling `blacklisted` from it; hand-authored (drizzle-kit cannot generate a table rename non-interactively) and verified by confirming `drizzle-kit generate` reports zero diff against the new snapshot, and by applying it against a real Postgres instance.
- The IAM role seeder (`pnpm db:seed`) is idempotent but not invoked automatically at app boot - an operator with an existing `compliance-manager` row needs to re-run it after this change lands for the role to pick up the new `regulatory-overview` grant.
- `checkRegistration`'s external shape (`Promise<{ allowed: boolean }>`) is unchanged - other code depends on it live at registration time.
- No new runtime enforcement (redirect delivery, KYC-by-country auto-approval) - explicitly out of scope per the ticket, deferred to separate stories.


mp-blurify:
## Code review — findings

### Blocking

**1. `packages/core/src/compliance/drizzle/migrations/0008_country_rule_and_global_kyc.sql:1` — ticket id in file content in a public repo**

The file starts with `-- BF-534: rewrites the old binary allow/block geo rule…`. Root `AGENTS.md` states a client's ticket id must never appear in a file name or in file content. It is the only such occurrence in the tree, and CI misses it: `tools/lint/verify-oss-hygiene.ts` filters on `scannedExtensions = new Set(['.ts', '.tsx', '.md', '.json'])`, so `.sql` is never content-scanned. Left as-is, the ticket key ships to the public npm package and to GitHub.

Fix: drop the `BF-534:` prefix (the rest of the comment is fine), and separately consider adding `.sql` to `scannedExtensions`.

**2. `packages/core/src/compliance/service/compliance.service.ts:44-52` — unresolvable-IP sentinel now fires on unrelated rows**

The sentinel still asks "does *any* rule row exist?" (`select({ id: countryRule.id }).from(countryRule).limit(1)`), but under the new model a row no longer implies that blocking is configured.

Failure scenario: an admin opens Regulatory Overview and sets `kycRequired: false` for Germany — nothing blacklisted anywhere. That writes the first `country_rule` row. From that moment `geoCheck` returns `allowed: false` for every request whose IP the GeoIP adapter cannot resolve, and `IdentityService:554` (`if (this.geoCheck && !(await this.geoCheck.checkRegistration(ip)).allowed)`) rejects those registrations platform-wide.

It is fail-closed, so not a compliance breach, but it is a silent production registration outage triggered by an unrelated toggle. No test covers this branch before or after the change (`compliance.service.int.test.ts` has no "rules exist + unresolvable IP" case).

Fix: scope the sentinel to `where(eq(countryRule.blacklisted, true))` and add the missing test.

### Worth fixing

- **`packages/core/src/compliance/service/compliance.service.ts:72-100`** — `upsertCountryRule` is a full-object last-write-wins PUT with no optimistic concurrency, and the `before` row is read with a plain `SELECT` (no `FOR UPDATE`) inside the transaction. Two consequences: (a) admin A loads the screen, admin B blacklists RU, admin A then toggles `kycRequired` for RU from their now-stale page and silently sets `blacklisted` back to `false` — a restricted country is re-opened by someone who never intended to touch it; (b) under concurrent writes to the same country, the second transaction blocks on the `ON CONFLICT` row lock but keeps its pre-block `before` snapshot, so the per-setting audit row the acceptance criteria demand can record a wrong previous value. Suggested fix: take the `before` read `FOR UPDATE` (or `FOR NO KEY UPDATE`) and accept an `updatedAt`/version from the client to reject stale writes.
- **`packages/core/src/compliance/contract/index.ts:492-503`** — the confirmation guard is inverted relative to risk: `.refine((input) => !input.blacklisted || input.confirmBlacklist === true)` demands confirmation only for the *protective* direction. Un-blacklisting a country, and setting `kycRequired: false` for one, both go through with no confirmation at all — while `SetGlobalKycConfigInputSchema` correctly requires `confirm` in both directions. Require confirmation on any transition that *weakens* a country's rules.
- **`packages/core/src/compliance/schema/index.ts:87-107`** — `countryRule.kycRequired`, `countryRule.redirectIp` and `globalKycConfig.enabled` have no readers anywhere outside the compliance config surface itself (`git grep` for each returns only the schema, the service's own CRUD, the contract and the tests; `geoCheck` reads only `blacklisted`). The PR body says runtime enforcement is deliberately out of scope and the direction is fail-safe, but this ships a Backoffice screen with three controls that do nothing — an operator can set "KYC not required for country X" and reasonably believe it took effect. Worth confirming the follow-up stories exist and are linked, and labelling these as not-yet-enforced in the UI.

### Notes
- `docs/platform/system-design.md:91` still reads `COMP["compliance ⟨geo_rule·user_limit⟩"]`. It sits outside the `gen:catalog-reference` block (lines 290-309), so `pnpm check:drift` will not catch it; the generated table at line 301 is correct.
- `packages/core/src/compliance/schema/index.ts:99` points at `docs/modules/compliance.md`, which does not exist (`docs/modules/` holds only `cms.md` and `wallet.md`).
- Breaking removals for a published `@openora/core@0.4.0`: routes `POST`/`GET /compliance/geo-rules`, exports `GeoRuleSchema` / `GeoRuleActionSchema` / `geoRuleActions`, and the `compliance.geo-rule.added` topic from `domainEventSchemas` — a consumer overlay subscribing to that topic breaks at boot.
- Existing deployments need `pnpm db:seed` re-run before compliance-manager sees the new grant (fail-closed until then), as the PR body notes; `seedRoles` is a convergent upsert, so re-running is safe.
- No SealedToken is rebound or bypassed; every new admin route calls `await adminGuard.assert(context, 'regulatory-overview', <action>)` as its first line; both audit writes are inside the mutation's transaction via the sealed `AUDIT_WRITER`; `CountryCodeSchema` is uppercase-enforced so there is no case-mismatch blacklist bypass; the country used by `geoCheck` comes from the server-side `GEO_IP_ADAPTER`, never from client input.

jakubfilinger-b:
## Second review pass

I went through the diff independently (service, contract, RBAC wiring, migration, tests) before reading the review already on this PR. Most of it I reproduce, so treat this as corroboration plus five findings that pass did not cover.

### Confirmed blocking

**`geoCheck` unresolvable-IP sentinel — `packages/core/src/compliance/service/compliance.service.ts:44-52`**

Same conclusion as the existing review, reached separately, and I traced the blast radius: the sentinel still asks "does any rule row exist?", but under the new model a row no longer implies blocking is configured. Setting `kycRequired: false` for one country writes the first `country_rule` row, and from that point `checkRegistration` returns `allowed: false` for every IP the GeoIP adapter cannot resolve — rejected at `packages/core/src/pam/identity/service/identity.service.ts:554`. Fail-closed, so not a compliance breach, but a silent registration outage triggered by an unrelated toggle.

I also confirmed the branch is untested in both directions: `compliance.service.int.test.ts` covers "no geo-ip port", "lookup resolves no country **with an empty table**", "resolved country blacklisted" and "resolved country not blacklisted" — there is no "rules exist + unresolvable IP" case before or after this change.

Fix: scope the probe to `where(eq(countryRule.blacklisted, true))`, and add the missing test.

### Not covered by the earlier pass

**1. `igaming.blockedCountries` becomes a second, unenforced blocking source — and downstream already relies on it**

`packages/core/src/contracts/schemas/igaming-config.ts:60` documents `blockedCountries` as "Countries blocked regardless of licensing (geo-block list)". `git grep blockedCountries` across `packages/` and `apps/` returns only the schema and its own unit test — nothing reads it at runtime. Meanwhile a downstream consumer sets `blockedCountries: ['US']` in its `apps/api/src/main.ts` and reasonably believes the US is geo-blocked. It is not.

The gap predates this PR, but this PR is what makes `country_rule.blacklisted` the canonical blocking store, so it is the right place to resolve it: either seed `country_rule` from `igaming.blockedCountries` at boot, or drop the field from the config schema. Leaving two declared geo-block mechanisms where only one is wired is the worst of the three options on a real-money path.

**2. The "a jurisdiction cannot be blocked" invariant is enforced in config only**

`IgamingConfigSchema.superRefine` (`igaming-config.ts:67-74`) rejects a config where a country appears in both `jurisdictions` and `blockedCountries`. `upsertCountryRule` has no equivalent check, so an admin can blacklist the operator's own licensed jurisdiction through the new screen. Same rule, two enforcement levels. If the invariant is worth a hard config failure it is worth a guard on the mutation, otherwise it should come out of the schema.

**3. `COUNTRY_RULE_FIELD_DEFAULTS` duplicates the column defaults, tied together only by a comment**

`compliance.service.ts:22-31` restates `blacklisted: false, redirectIp: false, kycRequired: true`, and the same pattern repeats for `GLOBAL_KYC_ENABLED_DEFAULT`. Changing a `.default()` in `schema/index.ts` silently makes the audit `before.value` wrong for first-time rows — which is precisely the field the acceptance criteria demand. No test pins the constants to the actual column defaults. Either derive them from the drizzle table definition or add a test that asserts they match a freshly inserted row.

**4. `getGlobalKycConfig` returns two different shapes**

The row branch carries `id`, `singletonKey` and `createdAt`, none of which `GlobalKycConfigSchema` declares; the fallback branch does not. The service's return type is a union that does not match its own contract, and it works only because zod strips unknown keys on output. Worth normalizing to the declared shape inside the service rather than relying on that.

**5. The `serializeRow` inference comments**

Two six-line comments justify assigning to a local instead of returning inline. The stated cause (`serializeRow`'s date-field generic widening to `keyof Row` under a nested contextual type) is not something a reader can verify from the comment, and it is duplicated verbatim. Either pin it with a minimal repro or drop both comments and keep the locals unexplained.

### Verified clean

- **Migration.** Snapshot chain is intact (`0008_snapshot.prevId` matches `0007_snapshot.id`); `country_rule` carries all six expected columns plus `country_rule_country_code_unique`; the backfill runs before `DROP COLUMN "action"`; `geo_rule_country_code_unique` really was a named table constraint in `0000_talented_mattie_franklin.sql`, so `RENAME CONSTRAINT` resolves; `action` was plain `text`, so the absent `DROP TYPE` is correct.
- **RBAC.** `regulatory-overview: RW` expands through `levelToActions` to all three actions, `read` to `view` only; super-admin picks the resource up automatically via `allReadWrite` over `statement`; the `admin` DB role correctly does not get it. The static `adminRole` in `permissions.ts` does grant it, but that table is only consulted when a user has no DB assignment (the bootstrap path), and it is consistent with every other resource there.
- `seedRoles` is a convergent upsert on `(roleId, resource)`, so re-running `db:seed` does pick up the new grant, as the PR body says; the grants cache TTL is 10s, so rollout is effectively immediate.
- No dangling `geoRule` / `addGeoRule` references in code. The remaining hits are a historical ADR and `packages/mcp/docs/catalog.json`, which is already stale on `dev` (last regenerated 2026-08-05, missing e.g. `chat_command_config`) and is not covered by `check:drift` — pre-existing, not this PR's problem.

One correction to the earlier pass: `docs/modules/` contains only `wallet.md`, not `cms.md` and `wallet.md`. The point about the dangling `docs/modules/compliance.md` reference stands.

damianrzepka:
@mp-blurify Thank you — addressed in 12f4fd15.

- Removed the public ticket identifier from the migration and extended the hygiene check to SQL files.
- Scoped the unresolved-IP fail-close probe to actual blacklists and added the missing regression coverage.
- Added locked snapshots and expectedUpdatedAt optimistic concurrency for country and global settings; stale full-object writes now conflict. Confirmation is evaluated against the stored country rule, so it is required for every weakening transition.
- Enforced configured blockedCountries at runtime and reject blacklisting a configured licensed jurisdiction.
- Normalized the global KYC response to its contract shape, removed duplicated defaults and inference comments, and corrected the system diagram and dangling documentation reference.
- Restored the legacy geo-rule routes, contract exports and event topic for existing overlays.

The per-country KYC and redirect values remain persisted, audited policy configuration; their runtime enforcement needs a separately defined enforcement path and is not claimed as part of this change. The seed re-run note remains a rollout requirement, as documented.

damianrzepka:
@jakubfilinger-b Thank you for the independent pass — addressed in 12f4fd15.

- blockedCountries is now an enforced geo-check baseline, including fail-closed behavior for an unresolved IP when a block is configured.
- Country-rule updates lock the existing row and require expectedUpdatedAt; stale writes conflict, audit before values come from the locked row, and a configured licensed jurisdiction cannot be blacklisted.
- Country-rule weakening requires confirmation based on the previous persisted state.
- Audit defaults now come from a newly inserted default row; the global default has one schema-owned source. The global KYC service output is normalized to the declared contract shape, and the serializeRow inference comments are gone.
- Restored the published legacy geo-rule routes, exports and event topic, with compatibility handlers backed by country_rule.

The migration chain and RBAC findings required no change. Per-country KYC and redirect enforcement still need a dedicated follow-up with explicit runtime enforcement points; this headless core does not ship the Backoffice UI.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -56,30 +146,185 @@ export class ComplianceService {
...
-        .insert(geoRule)
-        .values({ ...input })
-        .onConflictDoUpdate({
-          target: geoRule.countryCode,
-          set: { action: input.action },
-        })
-        .returning(),
-      new GeoRuleNotFoundError(input.countryCode),
-    );
+    if (input.action === 'block' && this.igaming?.jurisdictions.includes(input.countryCode)) {
```
jakubfilinger-b: The restored legacy writer mutates the same `country_rule.blacklisted` state through a route guarded only by `compliance:override-limit`, so risk-fraud-manager and kyc-aml-officer can bypass BF-534's restricted roles, confirmation/version checks, and canonical per-setting before/after audit; route this compatibility path through the `regulatory-overview` guard and the same transactional mutation invariants, or retire the legacy write.

packages/core/src/compliance/drizzle/migrations/0008_country_rule_and_global_kyc.sql
```diff
@@ -0,0 +1,25 @@
+-- Rewrites the old binary allow/block geo rule into three independent per-country
+-- flags (blacklisted, redirectIp, kycRequired), and adds a new global_kyc_config singleton
+-- for the platform-wide KYC toggle. Hand-authored (not drizzle-kit generated) because the
+-- geo_rule -> country_rule rename needs a data backfill from the dropped `action` column
+-- before it is dropped - see 0006_backfill_session_limit_currency.sql for the same pattern
+-- in this module. `action` was a plain `text` column (geoRuleActions was a TS-only enum,
+-- never a Postgres enum type), so there is no DROP TYPE step.
+ALTER TABLE "geo_rule" RENAME TO "country_rule";--> statement-breakpoint
```
jakubfilinger-b: Renaming `geo_rule` and later dropping `action` in this migration breaks every previous-version instance still querying that shape during a rolling deployment; ship an additive table/column plus backfill phase first, keep the old reader shape until all old instances are retired, and generate the migration artifacts through the repository workflow instead of the hand-authored destructive migration.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -56,30 +146,185 @@ export class ComplianceService {
...
+            .from(countryRule)
+            .where(eq(countryRule.countryCode, input.countryCode))
+            .for('update'),
+          new CountryRuleNotFoundError(input.countryCode),
+        );
+
+      if (!hasExpectedVersion(before.updatedAt, input.expectedUpdatedAt)) {
+        throw new CountryRuleVersionConflictError();
+      }
+      if (weakensCountryRule(before, input) && !input.confirm) {
```
jakubfilinger-b: This guard requires confirmation only for weakening transitions, so a new/default rule can change `blacklisted` from `false` to `true` with no `confirm`; BF-534 explicitly requires confirmation before blacklisting, so require `confirm === true` for the `false -> true` blacklist transition and add the missing rejection test.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -1,50 +1,140 @@
...
-      const [anyRule] = await this.drizzle.db
-        .select({ action: geoRule.action })
-        .from(geoRule)
+      const [blacklistedRule] = await this.drizzle.db
+        .select({ countryCode: countryRule.countryCode })
+        .from(countryRule)
+        .where(eq(countryRule.blacklisted, true))
         .limit(1);
-      return anyRule
+      return blacklistedRule || this.igaming?.blockedCountries.length
```
jakubfilinger-b, line 140: When no GeoIP adapter is bound, `countryCode` is null and any configured blacklist makes this branch deny every request, while the Regulatory Overview spec says country lookup is permissive and no rule takes effect until a provider is bound; return the permissive result before querying rules when `this.geoIp` is absent, and cover `no adapter + existing blacklist`.

packages/core/src/compliance/__tests__/regulatory-overview.router.int.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    assert: vi.fn(async (_ctx: unknown, resource?: string, action?: string) => {
+      if (resource && action && !allowed.includes(`${resource}:${action}`)) {
+        throw new ORPCError('FORBIDDEN', { message: `Missing permission: ${resource}:${action}` });
+      }
+      return { userId: ADMIN, role: 'admin' };
+    }),
+  });
+}
+
+function build(guard: AdminGuard) {
```
jakubfilinger-b, line 57: This test mocks both `AdminGuard` and `ComplianceService`, so it cannot prove the seeded-role authorization, real persistence, or transactional audit required by BF-534; add a `bootTestApp` E2E with real Postgres covering super-admin/compliance-manager success, excluded-role denial, and persisted audit rows.

jakubfilinger-b (changes requested):
Changes requested: the legacy geo-rule path bypasses BF-534's authorization, confirmation, and audit invariants, the migration is unsafe for rolling deployment, GeoIP fallback contradicts the spec, and the PR must also be rebased to resolve its current `dev` conflict.

jakubfilinger-b (commented):
## Review — BF-534 core side

Solid work overall: the per-setting audit rows match the AC exactly, `SELECT … FOR UPDATE` + `expectedUpdatedAt` gives real optimistic locking, the new RBAC resource carries its reasoning in a code comment, and the hand-written migration documents *why* it is hand-written. Findings below, most severe first.

### 1. [High] The legacy geo-rule routes bypass the new `regulatory-overview` gate

`addGeoRule` and `listGeoRules` now read and write the **same** `country_rule` table, but are still guarded on the old resource — `packages/core/src/compliance/router/index.ts:112-124`:

```ts
adminGuard.assert(context, 'compliance', 'override-limit')   // addGeoRule
adminGuard.assert(context, 'compliance', 'view')             // listGeoRules
```

`compliance: RW` is held by **Risk & Fraud Manager** and **KYC/AML Officer** — precisely the two roles the "Alternatives considered" section says must not reach country rules. A KYC/AML Officer can call `POST /compliance/geo-rules {"countryCode":"US","action":"allow"}` and clear a blacklist, bypassing every guarantee the new path adds: no confirmation on weakening, no `expectedUpdatedAt` check, no licensed-jurisdiction guard, and a different audit shape than the AC requires. `listGeoRules` on `compliance:view` similarly exposes the blacklist to Admin, Payments, Finance and Support.

The new resource cannot deliver the AC while a second door into the same table stays open. Re-gate both to `regulatory-overview`, or delete them.

### 2. [Medium] `geoCheck` fail-closed behaviour silently weakens on migration

Old: *any* `geo_rule` row → unresolvable IP fails closed. New: only a **blacklisted** row (or non-empty `igaming.blockedCountries`). Since the migration backfills old `action='allow'` rows to `blacklisted=false`, a deployment whose list was purely allow-based stops failing closed on unresolvable IPs immediately after deploy. Defensible as intent, but it is a licence-relevant loosening that deserves an explicit note in the PR body and the migration header.

### 3. [Medium] `kycRequired`, `redirectIp` and `globalKycConfig.enabled` are written, audited, and never read

Out of scope per the ticket — but the paired backoffice screen will ship an admin toggle labelled *KYC* that has zero runtime effect, fully audited. That is a compliance-theatre risk in an audit review. Either the UI marks these as not-yet-enforced, or enforcement lands in the same release.

### 4. [Low] `listCountryRules` has no `ORDER BY`

Postgres returns heap order, and an updated row physically moves — so the backoffice grid reshuffles after every save. Add `.orderBy(countryRule.countryCode)`.

### 5. [Low] A no-op upsert still creates a row

The `insert … onConflictDoNothing()` runs before the change check, so saving pure defaults materialises a `country_rule` row. The schema comment states "row presence = country has a rule", so this muddies that invariant.

### 6. [Low] Inconsistent `confirm` modelling

`UpsertCountryRuleInputSchema` uses `confirm: z.boolean().optional()` plus a service-side weakening check; `SetGlobalKycConfigInputSchema` uses `confirm: z.boolean()` + `.refine(=== true)` — `z.literal(true)` expresses that directly. Also `.strict()` on the first but not the second.

### 7. [Low] `addGeoRule` audits no-ops

It returns early when `before.blacklisted === blacklisted` but still emits `compliance.geo-rule.added`, writing an audit row for a no-op — the opposite of the new path's deliberate "zero audit rows when nothing changed".

### 8. [Nit] `packages/mcp/docs/catalog.json` still lists `geo_rule`

Pre-existing drift (`check:drift` only covers `docs/catalog.json`), but it now names a table that no longer exists.

---

**Downstream note:** betfeel `!378` consumes these routes but sends `confirmBlacklist` instead of `confirm` and never sends `expectedUpdatedAt`, and it ships no global-KYC toggle. The two halves need to agree before either merges.

packages/core/src/compliance/router/index.ts
```diff
@@ -119,6 +123,45 @@ export function createComplianceRouter({
       return compliance.listGeoRules();
     }),
 
+    upsertCountryRule: os.upsertCountryRule.handler(async ({ input, context }) => {
+      const { userId, ip, userAgent } = await adminGuard.assert(
+        context,
+        'regulatory-overview',
+        'manage-country-rules',
+      );
+      return mapErrors(
```
mp-blurify, line 141: `upsertCountryRule` maps `LicensedJurisdictionBlacklistError` to `CONFLICT` here, but the sibling `addGeoRule` handler at :112 now throws the same error (`compliance.service.ts:282`) with no `mapErrors`, so it escapes unmapped and surfaces as a 500 for the identical condition. `docs/standards/errors.md:24`.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -1,50 +1,140 @@
...
 export const LimitNotFoundError = makeNotFoundError('Limit');
 
 export const LimitOwnershipError = makeOwnershipError('Limit');
 
-export const GeoRuleNotFoundError = makeNotFoundError('GeoRule');
+export const CountryRuleNotFoundError = makeNotFoundError('CountryRule');
+
+export const GlobalKycConfigNotFoundError = makeNotFoundError('GlobalKycConfig');
+
+export const CountryRuleVersionConflictError = makeConflictError(
```
mp-blurify, line 33: These four new typed errors (`CountryRuleVersionConflictError`, `GlobalKycConfigVersionConflictError`, `LicensedJurisdictionBlacklistError`, `CountryRuleConfirmationRequiredError`) are not added to the export block in `compliance/server.ts:1`, unlike `LimitNotFoundError` / `LimitOwnershipError`. A consumer calling `ComplianceService.upsertCountryRule` directly cannot catch or re-map them. `docs/standards/errors.md:25`.

packages/core/src/compliance/contract/index.ts
```diff
@@ -163,6 +175,37 @@ const DeleteLimitInputSchema = LimitSchema.pick({ id: true });
 export const AddGeoRuleInputSchema = GeoRuleSchema.pick({ countryCode: true, action: true });
 export type AddGeoRuleInput = z.infer<typeof AddGeoRuleInputSchema>;
 
+export const UpsertCountryRuleInputSchema = z
```
mp-blurify: `UpsertCountryRuleInputSchema` re-declares `countryCode`, `blacklisted`, `redirectIp` and `kycRequired` instead of deriving them from `CountryRuleSchema`, so the input and output shapes can drift independently. `AddGeoRuleInputSchema` three lines up already does it the right way: `CountryRuleSchema.pick({ ... }).extend({ expectedUpdatedAt, confirm }).strict()`. `docs/standards/types.md:39`.

packages/core/src/compliance/contract/index.ts
```diff
@@ -163,6 +175,37 @@ const DeleteLimitInputSchema = LimitSchema.pick({ id: true });
...
+export type UpsertCountryRuleInput = z.infer<typeof UpsertCountryRuleInputSchema>;
+
+export const GlobalKycConfigSchema = z.object({
+  enabled: z.boolean(),
+  updatedAt: TimestampSchema.nullable(),
+  updatedBy: UuidSchema.nullable(),
+});
+export type GlobalKycConfig = z.infer<typeof GlobalKycConfigSchema>;
+
+export const SetGlobalKycConfigInputSchema = z
```
mp-blurify, line 200: `confirm: z.boolean()` plus a `.refine()` that rejects `false` encodes at runtime what the type can encode directly, and the `.refine()` turns this into a `ZodEffects` consumers can no longer `.pick()` / `.extend()` - unlike every other input schema in this file. It also drops the `.strict()` its sibling at :187 carries, so unknown keys pass silently. `confirm: z.literal(true)` with `.strict()`. `docs/standards/types.md:42`.

packages/core/src/compliance/contract/index.ts
```diff
@@ -195,6 +238,24 @@ export const complianceContract = {
     .route({ method: 'GET', path: '/compliance/geo-rules' })
     .output(z.array(GeoRuleSchema)),
 
+  upsertCountryRule: oc
```
mp-blurify, line 241: `upsertCountryRule`, `listCountryRules`, `getGlobalKycConfig` and `setGlobalKycConfig` ship with no E2E: `packages/testing/src/__tests__/` is untouched by this diff, so nothing exercises the real Hono + oRPC app, the real `AdminGuard` against seeded `regulatory-overview` grants, or the audit rows landing in `audit_log`. `docs/standards/testing.md:9` calls that the acceptance artifact for a route.

packages/core/src/contracts/schemas/iam.ts
```diff
@@ -25,6 +25,12 @@ export const adminStatement = {
   game: ['view', 'enable', 'disable'] as const,
   content: ['create', 'update', 'delete', 'publish'] as const,
   compliance: ['view', 'override-limit', 'manage-rg'] as const,
+  // Separate from `compliance` on purpose: risk-fraud-manager and kyc-aml-officer already
```
mp-blurify: Five-line comment explaining why this resource is separate from `compliance` - a rationale, not a fact the code cannot hold, and already stated verbatim in the PR description. `docs/standards/comments.md:7`.

packages/core/src/compliance/__tests__/compliance.service.int.test.ts
```diff
@@ -50,61 +62,325 @@ describe('ComplianceService.geoCheck (real PG)', () => {
...
+      actorId,
+    );
+
+    expect(rule).toMatchObject({
+      countryCode: 'FR',
+      blacklisted: true,
+      redirectIp: false,
+      kycRequired: true,
+    });
+    // Only `blacklisted` differs from the column defaults (false, false, true) - one audit row.
```
mp-blurify: Narration comment restating what the assertions below already state (same at :173). `docs/standards/comments.md:9`: never in tests.

mp-blurify:
The two blockers already raised on this PR stand (legacy write path against the same column, and the expand/contract migration). On top of those: the four new typed errors are not exported from the module's server subpath, and the four new routes ship without an E2E through the real app. Details inline.

packages/core/src/compliance/router/index.ts
damianrzepka: Fixed in 97971850. The compatibility handler now maps the shared confirmation, stale-version, and licensed-jurisdiction country-rule errors to CONFLICT as well.

packages/core/src/compliance/service/compliance.service.ts
damianrzepka: Fixed in 97971850. All four new typed country-rule/global-KYC errors are exported from the compliance server subpath.

packages/core/src/compliance/contract/index.ts
damianrzepka: Fixed in 97971850. UpsertCountryRuleInputSchema now derives its shared fields from CountryRuleSchema and only extends it with the mutation metadata.

packages/core/src/compliance/contract/index.ts
damianrzepka: Fixed in 97971850. The confirmation field is now z.literal(true) and the schema is strict, preserving an object schema that consumers can extend or pick.

packages/core/src/compliance/contract/index.ts
damianrzepka: Fixed in 97971850. Added real-app E2E coverage for all four routes with the seeded compliance-manager role, player denial, and persisted audit rows for both mutations.

packages/core/src/contracts/schemas/iam.ts
damianrzepka: Removed in 97971850.

packages/core/src/compliance/__tests__/compliance.service.int.test.ts
damianrzepka: Removed in 97971850.

packages/core/src/compliance/service/compliance.service.ts
damianrzepka: Fixed in 97971850. The legacy writer now requires regulatory-overview:manage-country-rules and delegates to the canonical transactional mutation, including confirmation, version checks, and per-setting audit records.

packages/core/src/compliance/drizzle/migrations/0008_country_rule_and_global_kyc.sql
damianrzepka: Fixed in 97971850. Replaced the destructive migration with repository-generated additive migration artifacts: geo_rule and action stay in place for existing instances, while the new API derives blacklisted from action and adds only the new columns.

packages/core/src/compliance/service/compliance.service.ts
damianrzepka: Fixed in 97971850. A false-to-true blacklist transition now also requires confirm: true, with a rejection test.

packages/core/src/compliance/service/compliance.service.ts
damianrzepka: Fixed in 97971850. geoCheck is now permissive when no GeoIP adapter is bound, even if blacklist rows exist; coverage was added.

packages/core/src/compliance/__tests__/regulatory-overview.router.int.test.ts
damianrzepka: Fixed in 97971850. Added bootTestApp E2E coverage with the seeded compliance-manager role, player denial, real persistence, and audit assertions.

damianrzepka:
Addressed in 97971850. The legacy writer now shares the regulatory-overview authorization and mutation invariants; the generated migration is additive and retains the geo_rule/action reader shape for rolling deployments. The typed errors are exported, and real-app E2E now covers all four routes with seeded authorization and audit persistence.

jakubfilinger-b (dismissed):

packages/core/src/compliance/contract/index.ts
```diff
@@ -160,9 +172,40 @@ export type GeoRule = z.infer<typeof GeoRuleSchema>;
 
 const DeleteLimitInputSchema = LimitSchema.pick({ id: true });
 
-export const AddGeoRuleInputSchema = GeoRuleSchema.pick({ countryCode: true, action: true });
+export const AddGeoRuleInputSchema = GeoRuleSchema.pick({ countryCode: true, action: true })
+  .extend({ confirm: z.literal(true).optional() })
```
mp-blurify, line 176: **Blocker — breaking change on the legacy route.** `confirm` is optional here, but `addGeoRule` forwards it into `upsertCountryRule` (`compliance.service.ts:313`), where `before.action !== 'block' && input.blacklisted` (`:192`) always demands it for a block. An existing client calling `POST /compliance/geo-rules {"countryCode":"US","action":"block"}` — the route's only real use — now gets 409 `confirmation_required` instead of 200, with no deprecation path and no test covering the omitted-`confirm` call.

Rule: `docs/adr/0037-admin-rg-limit-override-is-reduce-only.md:88` — an input gaining a required field is a recorded breaking change.
Fix: default `confirm` to `true` inside `addGeoRule` so the legacy shape keeps working, and require it only on `upsertCountryRule`.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -56,30 +163,178 @@ export class ComplianceService {
...
+      if (!hasExpectedVersion(before.updatedAt, input.expectedUpdatedAt)) {
+        throw new CountryRuleVersionConflictError();
+      }
+      if (
+        ((before.action !== 'block' && input.blacklisted) || weakensCountryRule(before, input)) &&
+        !input.confirm
+      ) {
+        throw new CountryRuleConfirmationRequiredError();
+      }
+      if (!hasCountryRuleChanges(before, input)) {
```
mp-blurify, line 197: When a new country is saved with values equal to the row defaults (`blacklisted:false, redirectIp:false, kycRequired:true`), the `INSERT` at `:172` has already committed the row but `hasCountryRuleChanges` returns false, so the method returns before `:216` — the country appears in `listCountryRules` with `updatedBy: null` and zero audit rows naming who created it. `compliance.service.int.test.ts:180` asserts this as correct.

Rule: `docs/standards/audit.md:5` — every mutation of configuration state produces a record with actor and before/after.
Fix: treat a freshly inserted row as a creation — stamp `updatedBy`/`updatedAt` and write a creation audit row even when the values match the defaults.

packages/testing/src/__tests__/regulatory-overview.e2e.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      actorId: complianceManagerId,
+      actorType: 'admin',
+      resourceType: 'global-kyc-config',
+      resourceId: 'global',
+      before: { enabled: true },
+      after: { enabled: false },
+    });
+  });
+
+  it('rejects a player from every regulatory overview route', async () => {
```
mp-blurify, line 166: The only negative is a player. The design this PR exists for — a new `regulatory-overview` resource so that `compliance: read_write` holders (`risk-fraud-manager`, `kyc-aml-officer` at `default-admin-roles.ts:110,141`) cannot touch country rules — has no test; seeding a `kyc-aml-officer` and granting it `regulatory-overview` by mistake would go unnoticed.

Rule: `docs/standards/testing.md:48` — always include authz negatives.
Fix: add an assignment of the seeded `kyc-aml-officer` role and assert 403 on all four routes.

docs/platform/system-design.md
```diff
@@ -88,7 +88,7 @@ flowchart TB
     PM["player-management (PAM)"]:::gated
   end
   subgraph DCOMP["@openora/core/compliance · RG/KYC"]
-    COMP["compliance ⟨geo_rule·user_limit⟩"]:::core
+    COMP["compliance ⟨country_rule·user_limit⟩"]:::core
```
mp-blurify: The diagram now names the compliance table `country_rule`, but the migration is additive and the table is still `geo_rule` — the generated catalog row at `:301` of this same file says `geo_rule`, so the doc contradicts itself.

Rule: `docs/standards/documentation.md:38-41`.
Fix: revert the node label to `geo_rule`.

packages/core/src/compliance/schema/index.ts
```diff
@@ -82,11 +82,28 @@ export const userLimit = pgTable(
...
+  redirectIp: boolean().notNull().default(false),
+  kycRequired: boolean().notNull().default(true),
+  createdAt: timestamp({ withTimezone: true }).notNull().defaultNow(),
+  updatedAt: timestamp({ withTimezone: true }),
+  updatedBy: uuid(),
+});
+
+export const GLOBAL_KYC_ENABLED_DEFAULT = true;
+
+// Platform-wide KYC toggle (admin-editable, audited). Does not overwrite a country's own
```
mp-blurify: Three lines that restate the table name and the `singletonKey` column, plus a behavioural claim ("does not overwrite a country's own kycRequired exemption") that no code implements — nothing reads `globalKycConfig` yet.

Rule: `docs/standards/comments.md:7,11` — a reason is not a fact; never restate a name.
Fix: delete the comment.

packages/core/src/compliance/__tests__/regulatory-overview.router.int.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    expect(result).toEqual(GLOBAL_KYC_CONFIG);
+    expect(compliance.setGlobalKycConfig).toHaveBeenCalledWith(
+      { enabled: false, confirm: true, expectedUpdatedAt: GLOBAL_KYC_CONFIG.updatedAt },
+      ADMIN,
+      expect.anything(),
+    );
+  });
+
+  it('a regulatory-overview:view grant alone does not unlock manage-country-rules', async () => {
```
mp-blurify: `'a regulatory-overview:view grant alone does not unlock manage-country-rules'` has a body byte-identical to `'upsertCountryRule requires regulatory-overview:manage-country-rules'` at `:107` — same guard, same call, same assertion.

Rule: `docs/standards/testing.md:7` — a test earns its keep by running real code.
Fix: delete one of the two.

packages/core/src/compliance/schema/index.ts
damianrzepka: Addressed in 31f694e3. Removed the redundant and unenforced behavioural comment from the global KYC schema.

packages/core/src/compliance/__tests__/regulatory-overview.router.int.test.ts
damianrzepka: Addressed in 31f694e3. Removed the duplicate router authorization test; the existing test already covers the same guard, call, and assertion.

packages/core/src/compliance/contract/index.ts
damianrzepka: Addressed in 31f694e3. The compatibility handler now defaults its internal confirmation to true, so existing POST /compliance/geo-rules block calls retain their prior input shape. The regulatory-overview endpoint still requires explicit confirmation.

packages/core/src/compliance/service/compliance.service.ts
damianrzepka: Addressed in 31f694e3. A newly inserted all-default country rule is now stamped with updatedBy and updatedAt and records a compliance.country_rule.created audit entry with the full initial state. A subsequent unchanged save remains audit-free.

packages/testing/src/__tests__/regulatory-overview.e2e.test.ts
damianrzepka: Addressed in 31f694e3. The real-app E2E now assigns the seeded kyc-aml-officer role and verifies 403 responses from all four regulatory-overview routes.

docs/platform/system-design.md
damianrzepka: Addressed in 31f694e3. The diagram now names the physical geo_rule table, matching the additive migration and generated catalog.

PR #153, opened 2026-09-09
feat(cms): carry dropped image URLs on banner mutations

jakubfilinger-b:
## The problem

An uploaded banner image is an object in someone's bucket. Core does not own that bucket, but it
owns the only signal that a file stopped being referenced - and that signal was unusable:

- `cms.banner.image.deleted` and `cms.banner.configuration.deleted` carried ids only, no URLs.
- Both fire *after* the rows are gone (`deleteConfiguration` cascades to `banner_image` via the
  FK's `onDelete: 'cascade'`), so a consumer cannot read the URLs back.
- `cms.banner.image.set` upserts. Overwriting a slot orphans the previous objects exactly like a
  delete does, and nothing said so.

The net effect downstream: every image ever uploaded outlives the banner that used it, with no way
to find it again. A consumer cannot even keep its own ledger, because `set` never carried the URLs
either.

## The change

All three events now carry `droppedImageUrls: string[]`:

| path | what it reports |
| --- | --- |
| `deleteBannerImage` | the row's two URLs |
| `deleteConfiguration` | every URL the cascade takes, read **inside the transaction** while the rows still exist |
| `setBannerImage` | what the upsert overwrote, minus any URL the new row reuses |

That last subtraction matters: editing only the desktop half of a slot, or only the link, must not
report the retained mobile URL as dropped - a consumer acting on it would blank a live banner.
There is a test for exactly that.

One field name across all three so a subscriber can handle them identically. Empty when nothing was
dropped.

## Scope

Core owns no object storage and makes no claim the URL is unused elsewhere - it reports only what
*this row* stopped pointing at. Deciding whether an object is safe to delete stays with whoever owns
the bucket.

## Compatibility

Additive. Existing consumers are unaffected, and the audit mapper builds a fixed shape from
`actorId`/`bannerImageId`/`bannerConfigurationId`, so it ignores the new field.

## Tests

Four integration cases in `cms.service.int.test.ts` (deleted row, cascade, first-time set, partial
overwrite). Full CMS suite 44/44.


mp-blurify:
## Code review — findings

### Blocking

**`packages/core/src/cms/service/cms.service.ts:517,660-662,740` + `packages/core/src/contracts/schemas/events.ts:40-47` — a dropped URL may still be live**

`droppedImageUrls` is computed per-row only; nothing queries whether the same URL string is still used by another `banner_image` row (different `sortOrder`, `locale`, or `bannerConfigurationId`). `bannerImage` has no uniqueness constraint on `desktopImageUrl` / `mobileImageUrl` (`packages/core/src/cms/schema/index.ts:53-78`), and `docs/modules/cms.md` explicitly documents that operators keep multiple configurations per placement for "a seasonal swap, an A/B draft" — i.e. URL reuse across configurations is a normal, supported pattern.

The changeset's own stated purpose is "a consumer that owns the object storage… can delete the orphans."

Failure scenario: configuration A (live) and configuration B (draft copy) both point `sortOrder 0` at the same CDN URL. Deleting or overwriting B's image reports that URL as dropped even though A still serves it live — a consumer following this contract's stated intent deletes a live asset.

The comment in `events.ts` already concedes this ("core does not check whether some other row still uses the same URL"), which makes it a documented sharp edge rather than a safe contract. Either add the reference check, or make the event name/shape say clearly that these are candidates requiring consumer-side verification.

### Worth fixing

- **`packages/core/src/cms/service/cms.service.ts:646-663` (`setBannerImage`)** — the `replaced` read and the `insert().onConflictDoUpdate()` are two separate non-transactional calls, unlike `deleteConfiguration`'s tx-wrapped read/delete. Two concurrent `setBannerImage` calls on the same `(bannerConfigurationId, sortOrder, locale)` slot can interleave so the `replaced` read is stale by the time the upsert commits, emitting a `droppedImageUrls` value that does not match what was actually overwritten. Narrow window, not covered by a test.
- **`packages/core/src/cms/__tests__/cms.service.int.test.ts`** — no test exercises the "still referenced elsewhere" case, so the gap above is unguarded by tests as well as by code. Suggested case: two configurations sharing an image URL, delete/overwrite one, assert the shared URL is *not* in `droppedImageUrls`.
- **`docs/modules/cms.md`** — the "Expected consumer integration" section (the doc a downstream operator reads to build exactly the storage-deletion consumer this feature is for) is not updated to mention `droppedImageUrls` or its per-row-only caveat. Worth adding before an operator wires blind deletion off these events.

### Note
`docs/catalog.json` does not encode event payload shapes, so `pnpm check:drift` / CI verify (already green here) cannot catch the new field — no action needed on that front.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -657,7 +665,9 @@ export const domainEventSchemas = {
   'cms.page.updated': cmsPageEventBase,
   'cms.page.deleted': cmsPageEventBase,
   'cms.banner.configuration.created': cmsBannerConfigurationEventBase,
-  'cms.banner.configuration.deleted': cmsBannerConfigurationEventBase,
+  'cms.banner.configuration.deleted': cmsBannerConfigurationEventBase.extend({
+    droppedImageUrls: droppedImageUrlsSchema,
```
damianrzepka, line 665: The default is only applied when a consumer calls parse itself. A legacy envelope already in the stream is forwarded unchanged by EventBus.on(); the publisher-side safeParse result is discarded. A newly deployed typed handler can therefore receive droppedImageUrls === undefined rather than []. If it iterates this field, EventBus catches the error and Redis ACKs the message, losing the cleanup signal. Please normalize known inbound event payloads before invoking handlers (or provide an equivalent upcaster) and cover a legacy envelope through the broker. This is required for the rolling-deployment guarantee in ADR-0016.

damianrzepka (changes requested):
NO-GO: older queued CMS banner events are not normalized when delivered to subscribers, so the promised empty droppedImageUrls default is not available during a rolling deployment. Please address the inline comment and add broker-to-subscriber coverage for a legacy envelope.

.changeset/banner-dropped-image-urls.md
```diff
@@ -0,0 +1,5 @@
+---
```
marek-chmielowski-blurify, line 1: Whats the point of this file?

packages/core/src/contracts/schemas/events.ts
```diff
@@ -40,6 +40,14 @@ const cmsBannerConfigurationEventBase = z
 const cmsBannerImageEventBase = z
   .object({ bannerImageId: UuidSchema, bannerConfigurationId: UuidSchema, actorId: UuidSchema })
   .extend(authContextBase.shape);
+// Image URLs this mutation stopped pointing at, so a consumer that owns the object
```
marek-chmielowski-blurify: There is no need of such a long comment. I think name of the const is clear and describes the purpose of it.

packages/core/src/contracts/schemas/events.ts
jakubfilinger-b: Addressed in 0c54fee6. EventBus now parses known inbound payloads before invoking subscribers and passes the normalized payload through both handler arguments. The Redis Streams integration test covers a legacy envelope without droppedImageUrls and verifies that the handler receives an empty array.

.changeset/banner-dropped-image-urls.md
jakubfilinger-b: This file is release metadata for the additive @openora/core contract change. Changesets uses it to generate the package version bump and changelog entry when the release is published, so I kept it and updated the description to match the final reference-safe behavior in eac3489b.

packages/core/src/contracts/schemas/events.ts
jakubfilinger-b: Agreed. I removed the explanatory comment and kept the schema name as the source of intent in eac3489b.

jakubfilinger-b:
Addressed in eac3489b. Each banner mutation now collects candidate URLs, applies the write, and checks both desktop and mobile URL columns across all remaining banner image rows inside the same transaction. URLs that are still referenced are excluded from droppedImageUrls. I added integration coverage for overwrite, image deletion, and configuration cascade cases with shared URLs, and documented the downstream storage-cleanup contract. The setBannerImage concurrency concern was addressed earlier in b163aa42 with transaction-scoped serialization, and legacy broker payload normalization is covered by 0c54fee6.

packages/core/src/cms/service/cms.service.ts
```diff
@@ -72,6 +72,31 @@ const pageCacheKey = (slug: string) => `cms:page:${slug}`;
 const publicBannerCacheKey = (placement: string, locale: string) =>
   `cms:banner-placement:${placement}:${locale}`;
 const bannerScheduleLockKey = (placement: string) => `cms:banner-schedule:${placement}`;
+const bannerConfigurationMutationLockKey = (bannerConfigurationId: Uuid) =>
+  `cms:banner-configuration:${bannerConfigurationId}`;
+
+async function unreferencedBannerImageUrls(tx: DrizzleTx, candidates: string[]) {
```
damianrzepka, line 92: This reference check is still racy across configurations. The advisory lock is keyed only by bannerConfigurationId, so concurrent mutations of A and B can interleave: A removes URL U and this query sees no reference; B writes U and commits; A then commits and emits U as dropped even though it is live. That breaks the documented guarantee and can make a cleanup subscriber delete a live object. Please serialize mutations by the affected/incoming URLs (or weaken the contract), and add a cross-configuration concurrency test.

packages/core/src/cms/service/cms.service.ts
```diff
@@ -72,6 +72,31 @@ const pageCacheKey = (slug: string) => `cms:page:${slug}`;
...
+  if (uniqueCandidates.length === 0) {
+    return [];
+  }
+  const references = await tx
+    .select({
+      desktopImageUrl: bannerImageTable.desktopImageUrl,
+      mobileImageUrl: bannerImageTable.mobileImageUrl,
+    })
+    .from(bannerImageTable)
+    .where(
```
damianrzepka, line 103: This new lookup filters banner_image by both desktopImageUrl and mobileImageUrl, but neither column is indexed. Please add indexes for the two lookup columns and generate the migration; the database standard requires every filtered column to be indexed.

packages/core/src/cms/service/cms.service.ts
jakubfilinger-b: Fixed in 2d341d5a. Banner image mutations now acquire deterministic, deduplicated advisory locks for every incoming and removed URL while retaining the per-configuration lock. This serializes reference checks across configurations and avoids deadlocks through sorted lock acquisition. Added a real-Postgres cross-configuration concurrency test that verifies both mutations wait on the shared URL lock.

packages/core/src/cms/service/cms.service.ts
jakubfilinger-b: Fixed in 2d341d5a. Added btree indexes for banner_image.desktop_image_url and banner_image.mobile_image_url and generated the CMS migration 0003_worthless_titanium_man.

damianrzepka (approved):

zaxovaiko (approved):

PR #154, opened 2026-09-09
feat(compliance): add per-game geo-blocking

okapitula:
## Summary

Add operator-managed per-game country restrictions enforced before new gaming rounds start. Existing rounds remain eligible to settle, while rule mutations are permissioned, audited, and concurrency-safe.

Country-code resolution is unchanged: the existing GeoIP adapter lookup and request metadata are reused. This change adds per-game policy evaluation and validation, not a new provider or resolution path.

## Why

Global geo rules cannot represent game-specific licensing restrictions. Operators need to block new starts for a specific game without disrupting already-admitted rounds.

## Alternatives considered

- Client-side blocking — rejected because jurisdiction enforcement must happen server-side.
- Global rules only — insufficient for game-specific licensing policies.
- Unsynchronized rule mutations — rejected because concurrent updates could produce inaccurate audit snapshots.

## Risks

- Adds a compliance migration for per-game geo rules.
- Unresolved GeoIP requests fail closed when the specific game has a geo rule.
- Requires the new `compliance:manage-geo` permission.
- Global rules continue to apply when a country is resolved.

## Validation

- `pnpm verify` passed.
- 20 test files and 167 tests passed.
- Typecheck, lint, formatting, migration, catalog, and boundary checks passed.


packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -90,7 +108,18 @@ export class GamingService {
...
+  async startRound(
+    userId: User['id'],
+    gameId: Game['id'],
+    currency: string,
+    betAmount: string,
+    ipAddress: string | null = null,
+  ) {
+    const selectedGame = await this.getGame(gameId);
+    if (!selectedGame.isActive) {
+      throw new GameNotFoundError(gameId);
```
jakubfilinger-b, line 120: `startRound` now throws `GameNotFoundError` for an inactive game. Before this PR it would open a round on an inactive game — `getGame` doesn't filter on `isActive`. Reasonable tightening, but it's a behaviour change on a money path that the PR description doesn't mention (scope is stated as geo policy evaluation only). Can you confirm it's intended, and that nothing relies on the old behaviour (e.g. a game disabled mid-session with a launch URL already handed out)? Otherwise worth splitting out.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -51,6 +76,44 @@ export class ComplianceService {
     return { allowed: true, countryCode, reason: null };
   }
 
+  async checkGame(input: GameGeoCheckInput) {
+    const rawCountryCode =
+      this.geoIp && input.ipAddress ? (await this.geoIp.lookup(input.ipAddress)).countryCode : null;
+    const countryCode = normalizeCountryCode(rawCountryCode);
+
+    if (!countryCode) {
```
jakubfilinger-b, line 93: `checkGame` re-implements the global geo decision (GeoIP lookup + `normalizeCountryCode` + global `block` query) instead of delegating to `geoCheck`, so the two gates can drift. Concrete gap today: `geoCheck` fails closed on unresolved geo whenever any global rule exists, but here an unresolved lookup lets the round start when only a global rule exists and the game has no rule of its own. It's covered by a test and matches the PR description, so it looks deliberate — but since this is a money/compliance gate please get an explicit sign-off, and consider sharing the country-resolution + global-block logic with `geoCheck` so a future change to one can't silently bypass the other.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -25,8 +49,9 @@ export class ComplianceService {
   ) {}
 
   async geoCheck(ipAddress: string | null) {
-    const countryCode =
+    const rawCountryCode =
       this.geoIp && ipAddress ? (await this.geoIp.lookup(ipAddress)).countryCode : null;
+    const countryCode = normalizeCountryCode(rawCountryCode);
```
jakubfilinger-b, line 54: `geoCheck` now runs the adapter result through `normalizeCountryCode`. Canonical ISO codes are unaffected, but non-canonical / invalid vendor values now normalise or fall through to fail-closed. That's a semantics change to the existing registration/login geo gate, which the PR description says is untouched. Fine if intended — just flag it so it's a conscious change and not a side effect.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -28,6 +30,21 @@ export const RgRestrictedError = makeConflictError(
   'RgRestrictedError',
   'play is restricted by an active responsible-gambling exclusion',
 );
+
+export type GameGeoRestrictedData = Pick<
+  Extract<GameGeoDecision, { allowed: false }>,
+  'reason' | 'countryCode'
+>;
+
+export class GameGeoRestrictedError extends Error {
```
jakubfilinger-b, line 39: `GameGeoRestrictedError` is a hand-rolled `Error` subclass while its siblings in this file use `makeConflictError` / `createDomainError`. It works (the E2E asserts 409 + `data`), but it's inconsistent with the module's error conventions — `createDomainError` with a data payload, like `RgLimitExceededError`, would fit better.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -61,6 +61,15 @@ const permissionLevelEntries = z.array(
   z.object({ resource: z.string(), level: PermissionLevelSchema }),
 );
 
+const gameGeoRuleEventState = z.object({
+  id: UuidSchema,
+  gameId: UuidSchema,
+  countryCode: CountryCodeSchema,
+  reason: z.string(),
```
jakubfilinger-b: `reason` here is `z.string()` but the contract's `GameGeoRuleSchema.reason` is `z.string().trim().min(1)`. Minor, but worth aligning the event payload with the API shape.

jakubfilinger-b (commented):
Reviewed the branch with a request trace on `POST /gaming/rounds/start` and the `/compliance/game-geo-rules` routes, plus the migration. Ran the compliance + gaming integration tests locally (41/41) and `check:drift` (clean). No blockers. Two things need a conscious decision before merge: the new rejection of inactive games in `startRound`, and the unresolved-geo + global-rule gap between `checkGame` and `geoCheck`. The rest are consistency nits. GO.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Confirmed intentional. A game marked inactive is not playable, so startRound must reject it before any responsible-gambling or geo checks, debit, or provider launch. This behavior is covered by the inactive-game guard in [32edc08d](https://github.com/blurifycom/openora/commit/32edc08d60d0aa910fa22b2bd8b805824cbebcbe).

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: I kept GameGeoRestrictedError as a typed Error subclass, following the existing RgLimitExceededError pattern from [32edc08d](https://github.com/blurifycom/openora/commit/32edc08d60d0aa910fa22b2bd8b805824cbebcbe). It carries per-decision reason and countryCode data and is mapped to CONFLICT correctly. The existing shape already works for this dynamic-data case, so I did not broaden the shared error factory.

packages/core/src/contracts/schemas/events.ts
okapitula: Addressed in [ff9b0191](https://github.com/blurifycom/openora/commit/ff9b01918e139c443eabddd124f1cef206d939b2). NonEmptyReasonSchema, using trim and min length 1, is now shared by the game-geo API and event payload/state schemas. Tests cover whitespace and blank reasons, and the catalog was regenerated.

packages/core/src/compliance/service/compliance.service.ts
okapitula: Addressed in [ff9b0191](https://github.com/blurifycom/openora/commit/ff9b01918e139c443eabddd124f1cef206d939b2). checkGame now delegates country resolution and global policy evaluation to geoCheck, so the gates cannot drift. Unresolved geo now fails closed when any global rule exists, with regression coverage updated accordingly.

packages/core/src/compliance/service/compliance.service.ts
okapitula: This is intentional and was introduced in [32edc08d](https://github.com/blurifycom/openora/commit/32edc08d60d0aa910fa22b2bd8b805824cbebcbe). Normalizing adapter output is a safeguard for valid values returned in non-canonical casing, such as us, so a configured US rule cannot be missed. Uppercase ISO codes are unchanged. Since checkGame now reuses geoCheck, the same normalization applies consistently.

packages/core/src/casino/gaming/service/gaming.service.ts
zaxovaiko: Still intentional after merging `dev`. The manual `isActive` guard is gone: `startRound` now opens with `await this.getGame(gameId, { activeOnly: true })`, the catalog-aware check `dev` introduced, so a deactivated provider is rejected too, not just an inactive game. It stays the first statement in the method, ahead of the RG and geo gates, the wallet debit and the provider launch.

packages/core/src/compliance/service/compliance.service.ts
zaxovaiko: Verified again after merging `dev`, including the country-rule rename in #152. `checkGame` still delegates the whole global decision to `geoCheck` and only then reads `game_geo_rule`, so it now inherits the `igaming.blockedCountries` pass and the no-adapter short-circuit for free. The gap you named is closed in both directions: unresolved geo with any blocking country rule fails closed in `geoCheck`, and `checkGame` returns that as `geo_unresolved` before it ever looks at the game.

packages/core/src/compliance/service/compliance.service.ts
zaxovaiko: Still a conscious change, and it survived the `dev` merge. `geoCheck` keeps dev's early return when no GeoIP adapter is bound, and the resolved lookup is wrapped in `normalizeCountryCode` before any rule query. Canonical uppercase ISO codes are untouched; a vendor value such as `us` now matches a configured `US` rule instead of silently missing it, and an unparseable value falls through to the existing fail-closed branch.

PR #155, opened 2026-09-10
feat(agents): let consumer agents change core through an oss worktree

zaxovaiko:
## Summary

- Consumer agents can now change core themselves: `pnpm oss:worktree <branch>` opens a git worktree of the linked checkout under `.worktrees/`, and the consumer `guard-core` hook allows writes there and nowhere else in the checkout.
- That path is gated so core stays generic: in place only for small fixes and contract alignment (features go through a session rooted here), a written genericity test before the first edit (why it cannot live in the consumer, why another operator would want it), and this repo's own review workflow plus the `contract-reviewer` checklist on the worktree diff before any push.
- The consumer `review` skill finds the paired core change by branch name (or by a canary pin bump), reviews it by this repo's own rules, and cross-checks the contract between the two diffs. Operator-specific behavior in core is a BLOCK there and in this repo's review. `create-pr` lands the core PR first.
- The guard now resolves paths from the repo root, so absolute paths are caught too. The old literal `../<oss>` match let them through.

## Why

The consumer guard denied every write into the linked checkout, so any change that needed a core fix ended in a handoff, and no review ever looked at both halves of a paired change together. The risk of lowering that bar is operator logic creeping into core, since a consumer session carries one operator's requirements - hence the genericity and scope gates. Everything ships through the consumer template, so every agent tool rulesync renders for gets the same flow on the next install, with no per-developer setup beyond the access flags listed in the new `docs/agents/cross-repo.md`.

## Alternatives considered

- Loading this repo's instructions into a consumer session (Claude's additional-directory loading, symlinked skills). Rejected because it only works in one tool; the rules instead tell every agent to read `AGENTS.md`, `.rulesync/rules/`, and `docs/standards/` from the worktree.
- Linking the two requests to each other. Rejected because this repo is public and a consumer request may be read outside the team, so the shared branch name is the only pairing.

## Risks

- A consumer only gets the new flow once its pin points at a canary that carries this template; until then its installs render the old one.
- The gates are instructions, not enforcement: this repo's edit hooks and reviewer agents do not run from a consumer session, and `guard-core` stays a best-effort string guard for shell commands (a write after a `cd` into the checkout is not caught). The human review on this repo's PR remains the real gate.


tools/templates/consumer/__dot__rulesync/hooks/guard-core.mjs.tpl
```diff
@@ -1,44 +1,62 @@
...
 if (command) {
   const writeToCore = [
-    [new RegExp(String.raw`\b(?:sed|perl)\b[^|;&]*\s-\w*i\w*\b[^|;&]*(?:${PROTECTED})`), 'in-place edit (sed/perl -i)'],
-    [new RegExp(String.raw`(?:>>?|>\|)\s*['"]?[^'"\s|;&]*(?:${PROTECTED})`), 'shell redirection'],
-    [new RegExp(String.raw`\btee\b\s+(?:-a\s+)?['"]?[^'"\s|;&]*(?:${PROTECTED})`), 'tee'],
-    [new RegExp(String.raw`\b(?:rm|truncate|dd|chmod|chown|unlink|shred|mv)\b[^|;&]*(?:${PROTECTED})`), 'destructive file op'],
+    [new RegExp(String.raw`\b(?:sed|perl)\b[^|;&]*\s-\w*i\w*\b[^|;&]*${PROTECTED}`), 'in-place edit (sed/perl -i)'],
+    [new RegExp(String.raw`(?:>>?|>\|)\s*['"]?[^'"\s|;&]*${PROTECTED}`), 'shell redirection'],
+    [new RegExp(String.raw`\btee\b\s+(?:-a\s+)?['"]?[^'"\s|;&]*${PROTECTED}`), 'tee'],
+    [new RegExp(String.raw`\b(?:rm|truncate|dd|chmod|chown|unlink|shred|mv)\b[^|;&]*${PROTECTED}`), 'destructive file op'],
```
jakubfilinger-b: `mv` is in this list and `cp` is not, which reads as an oversight rather than a decision — copying a file over a core source file is at least as ordinary an agent move as moving one, and it is the first thing that works when a direct `Write` gets denied. `cp overlay/thing.ts ../oss/packages/core/src/thing.ts` matches none of the four patterns: it is not `sed -i`, not a redirect, not `tee`, and not on this list. It overwrites core silently.

Worth adding in the same pass, for the same reason: `ln` (a symlink into the checkout, which then also defeats the check at :54 — see that thread), `install`, `patch`, and `git apply`. `git -C <oss> apply` / `git -C <oss> checkout .` deserve particular thought, because `git -C <oss>` is the idiom this PR's own `oss-worktree.mjs` uses, so it is the spelling an agent will have seen most recently.

The test at :72 is the right place to pin whichever ones you take — it already drives real payloads through the rendered hook, so each is one more `assert.equal(exitCode({ command: ... }), DENIED)`.

tools/templates/consumer/__dot__rulesync/hooks/guard-core.mjs.tpl
```diff
@@ -1,44 +1,62 @@
...
-// 2) Direct file edit/write whose target is inside a protected path.
+// 2) Direct file edit/write whose target resolves inside a protected path.
 const filePath = extractFilePath(payload);
-if (filePath && new RegExp(PROTECTED).test(filePath)) {
-  deny(
-    `Blocked: ${filePath} is inside OSS core / node_modules, which is read-only here. ` +
-      'Extend from the OUTSIDE (overlay, adapter, UI plugin, config), or report an upstream OSS issue.',
-  );
+if (filePath) {
+  const target = resolve(payload.cwd ?? process.cwd(), filePath);
```
jakubfilinger-b: `resolve()` normalises `.` and `..` but does not follow symlinks, so this check is on the path as written, not on the file that gets opened. A symlink anywhere inside the consumer repo pointing into the checkout — `consumer/core-src -> ../oss/packages/core/src` — makes `target` resolve to a consumer path, `isInside(OSS)` false, and the write allowed; the filesystem then follows the link into core.

That is not hypothetical here: agents create convenience symlinks, and the PR's own alternatives section mentions symlinked skills as something that was considered. It also compounds with the missing `ln` in the shell list above — an agent denied a direct write can create the link and then write through it, and neither step is caught.

The fix is small: `realpathSync` the deepest existing ancestor of `target` before the `isInside` comparisons (the target file itself may not exist yet, so walk up until one does). The test file already imports `realpathSync` for the macOS `/var` case, so the mechanism is familiar — it just needs applying to the target as well as to the sandbox.

tools/templates/consumer/tools/oss-worktree.mjs.tpl
```diff
@@ -0,0 +1,83 @@
...
+  git('fetch', 'origin', '--quiet');
+  const base = gitOrEmpty('rev-parse', '--abbrev-ref', 'origin/HEAD') || 'origin/dev';
+  if (gitOrEmpty('rev-parse', '--verify', '--quiet', `refs/heads/${branch}`)) {
+    git('worktree', 'add', worktree, branch);
+  } else if (gitOrEmpty('rev-parse', '--verify', '--quiet', `refs/remotes/origin/${branch}`)) {
+    git('worktree', 'add', '--track', '-b', branch, worktree, `origin/${branch}`);
+  } else {
+    git('worktree', 'add', '--no-track', '-b', branch, worktree, base);
+  }
+  run('pnpm', ['-C', worktree, 'install', '--frozen-lockfile']);
```
jakubfilinger-b, line 77: This installs but never builds, and `@openora/core` is consumed from `dist` (`"files": ["bin", "dist", ...]`, and the package's exports point there). So `pnpm oss:worktree <branch> --link` completes successfully and hands the consumer a linked package with an empty `dist/` — every `@openora/core` import fails to resolve until somebody works out that they also need to build inside the worktree. The first person through this hits it, and the error they see (module not found) points nowhere near the cause.

The same applies after every subsequent core edit, which is the thing that most surprises people working across the two repos: a change to core `src` does not reach the consumer, or a consumer e2e, until the worktree is rebuilt.

Please run the build here after the install — and say the rebuild-after-edit rule out loud in `docs/agents/cross-repo.md`. "Same command for every agent and every human" is the promise this script makes; a worktree that does not work until you know one more undocumented step does not keep it.

tools/templates/consumer/tools/oss-worktree.mjs.tpl
```diff
@@ -0,0 +1,83 @@
...
+const isLinkedTo = (dir) =>
+  existsSync('pnpm-workspace.yaml') &&
+  readFileSync('pnpm-workspace.yaml', 'utf8').includes(`link:${relative(process.cwd(), dir)}/`);
+
+if (flag === '--remove') {
+  const wasLinked = isLinkedTo(worktree);
+  if (existsSync(worktree) && gitOrEmpty('worktree', 'remove', worktree) === '' && existsSync(worktree)) {
+    die(`${worktreeFromHere} has uncommitted changes - commit or discard them, then rerun.`);
+  }
+  if (wasLinked && hasLinkScript()) {
```
jakubfilinger-b, line 60: Minor, but this one took me a re-read: the removal check works only because `git worktree remove` prints nothing on success, so `gitOrEmpty(...) === ''` is true in both the success and the failure case and the `existsSync` that follows is what actually distinguishes them. Correct, but it reads as if an empty return means failure. A `try { git('worktree', 'remove', worktree) } catch { die(...) }` would say the same thing in a way the next reader does not have to reconstruct.

jakubfilinger-b (changes requested):
This solves a real problem. The handoff-on-every-core-fix loop is genuinely expensive, and the observation that no review ever saw both halves of a paired change is the important one — that is where operator logic slips into core, not at the moment of writing it. Gating on a written genericity test and running this repo's own reviewer checklist on the worktree diff is the right shape, and landing it through the template so every rulesync target gets it identically beats per-tool wiring.

I'm blocking on two things in the guard itself, since it is the one mechanical control under all of the instruction-level gates and the PR is explicit that the instructions are not enforcement. Both are narrow and both are testable — details on the threads. Short version: `cp` is missing from the destructive-op list while `mv` is in it, and the direct-write check resolves paths without resolving symlinks.

The honesty in the Risks section is appreciated and I'm not asking you to close the `cd`-then-write hole — that one really is out of reach for a string guard, and the human review on the PR is the backstop. I am asking that the holes which *are* in reach get closed, because "best effort" is a much easier sell when the obvious cases are covered.

Third item, unrelated to security: `pnpm oss:worktree --link` currently hands back a worktree the consumer cannot build against. Thread on that too.

tools/templates/consumer/__dot__rulesync/hooks/guard-core.mjs.tpl
zaxovaiko: Good catch, thanks. Done in `86a52594` - `cp`, `install` and `rsync` are now denied when the checkout is the destination (copying out of core stays allowed, it's a read), `ln` and `patch` on any checkout path, and `git -C <oss>` with a writing subcommand (`apply`, `checkout`, `reset`, `restore`, `stash`, ...). Each one is pinned in `guard-core-template.test.mjs`.

tools/templates/consumer/__dot__rulesync/hooks/guard-core.mjs.tpl
zaxovaiko: True, `resolve()` alone was judging the path as written. Done in `86a52594` - the target now goes through `realpathSync` on its nearest existing ancestor, and so does the checkout path itself, otherwise a checkout that is itself a symlink would open the same hole from the other side. Added a test with a `core-src` link into the checkout.

tools/templates/consumer/tools/oss-worktree.mjs.tpl
zaxovaiko: Fair point. The `link:oss` script I tested against already builds the dir it's given, but the template can't rely on every consumer's script doing that. Done in `86a52594` - `--link` now runs `pnpm -C <worktree> build` before linking, and `cross-repo.md` plus `oss-boundaries.md` now say that a core edit needs a rebuild before the consumer sees it. If a consumer's `link:oss` builds too, the second build should just be a turbo cache hit.

tools/templates/consumer/tools/oss-worktree.mjs.tpl
zaxovaiko: Agree, it read backwards. Done in `86a52594` - switched to `try { git('worktree', 'remove', worktree) } catch { die(...) }`.

jakubfilinger-b (approved):
Rechecked the follow-up. The requested guard coverage, symlink handling, worktree build step, and removal flow are all addressed and covered by tests.

PR #157, opened 2026-09-10
docs(agents): make module AGENTS.md optional in the consumer structure rules

zaxovaiko:
## Summary

The consumer structure rules in `oss-boundaries` no longer say every frontend module has an `AGENTS.md`. Every module still has `index.ts`; an `AGENTS.md` is optional and only for knowledge local to the module that the rules and the code do not show.

## Why

A required `AGENTS.md` per module ends up as generated boilerplate that restates the same frontend rules in every module. An agent working across several modules then loads that text once per module, which costs context and buries the few notes that matter. The rules already live in the shared rule files and are enforced by the consumer's shape check and oxlint, so the per-module copy adds nothing.

## Risks

None. This is instruction text only; consumers pick it up with their next template sync.

### DIFF
--- a/tools/templates/consumer/__dot__rulesync/rules/oss-boundaries.md.tpl
+++ b/tools/templates/consumer/__dot__rulesync/rules/oss-boundaries.md.tpl
@@ -66,7 +66,7 @@ Enforced by `pnpm check:lint` (oxlint, per-edit), the pre-commit hook, CI, and t
 
 Runs the checks a dependency graph cannot make, because they are about files that do not exist or edges that do not exist:
 
-- Every module has `index.ts`. An `AGENTS.md` is optional - only for knowledge local to the module that the rules and the code do not show.
+- Every module has `index.ts`. An `AGENTS.md` is optional - only for knowledge local to the module that the rules and the code do not show. A `check:shape` that still requires one drops that assertion with this rule, or the gate contradicts it.
 - The barrel re-exports only; the single side effect it may carry is `import './locales'`.
 - Every file under a module's `pages|components|hooks|utils` is reachable from the barrel, and every `packages/ui/src` file from the `{{scope}}/ui` barrel - otherwise it is dead code.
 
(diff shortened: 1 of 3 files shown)


tools/templates/consumer/__dot__rulesync/rules/oss-boundaries.md.tpl
```diff
@@ -51,7 +51,7 @@ Enforced by `pnpm check:lint` (oxlint, per-edit), the pre-commit hook, CI, and t
 
 Runs the checks a dependency graph cannot make, because they are about files that do not exist or edges that do not exist:
 
-- Every module has `index.ts` and `AGENTS.md`.
+- Every module has `index.ts`. An `AGENTS.md` is optional - only for knowledge local to the module that the rules and the code do not show.
```
jakubfilinger-b: The wording is right, and the shape check this describes is consumer-owned so nothing in this repo needs to change with it. Two neighbours to keep in mind, neither blocking:

- `commands/scaffold-module.md` still advertises `AGENTS.md` in what it generates. That is the backend add-on *package* generator, not a frontend module, so it isn't strictly contradicted by this line — but a reader who has just been told the file is optional will hit that description next and wonder. A word there saying the package-level one is still generated would settle it.
- Consumers only pick this up on their next template sync, as the description says. Since the sentence a consumer is dropping ("every module has an `AGENTS.md`") is enforced by their own shape check, it's worth being explicit in the rule that removing the requirement means removing the check too — otherwise the first consumer to sync gets a rule saying optional and a gate saying mandatory, and trusts the gate.

jakubfilinger-b (dismissed):
Agreed with the reasoning, and it matches what the file actually became downstream: a per-module `AGENTS.md` that has to exist gets filled with a copy of the shared rules, and an agent then pays for that copy once per module it opens. Worse than the context cost is that the copies go stale independently — I've seen one still telling agents to reach for a component library the shared rule had already declared dead, so the required file was actively steering work the wrong way.

Approving. One consistency note on the thread.

tools/templates/consumer/__dot__rulesync/rules/oss-boundaries.md.tpl
zaxovaiko: Thanks! Both done in `4baecd71`. On the first one, the `module` generator actually never writes an `AGENTS.md`, so I dropped it from the `scaffold-module` description rather than saying it's still generated. On the second, the rule now says a `check:shape` that still requires the file drops that assertion with it. Side note - `gen module` is OSS-only (`ossOnly('module')`), so that consumer command throws in a consumer repo anyway. But that's for later.

jakubfilinger-b (dismissed):
Rechecked the follow-up. Both consistency points are addressed. The current verify failure is in unrelated FX timing tests, so it does not block this documentation/template change.

PR #158, opened 2026-09-10
feat(identity): per-player auto-logout window and 2FA on every login

jakubfilinger-b:
Backs BF-522 (Betfeel "Remember Device" settings). Adds the two per-player security preferences that page needs, plus the enforcement behind them.

## `autoLogoutDuration`

How long a session may sit idle before it is cut: `15m` / `1h` / `24h` / `7d` / `30d`. There is deliberately no "off" - the longest option already coincides with better-auth's absolute 30-day expiry, which this PR does not touch, so a "never" would have been indistinguishable from `30d` while reading as a real escape hatch. Defaults to `7d`.

Enforcement goes through a new `SESSION_IDLE_POLICY` port, bound by identity and resolved once per authenticated request in the `create-app` middleware, right beside the existing `PLAYER_ACTIVITY_TRACKER` call. It stamps `session.lastSeenAt` and expires a session that has outrun the window, using the same shape an explicit revoke does (`expiresAt = now()`, `updatedAt` left alone so the device list still shows real last-use). An expired session returns `expired`, and the middleware then serves the request on the unauthenticated path rather than publishing a session it has just killed.

Two details worth a reviewer's eye:

- The `lastSeenAt` write is throttled to 60s, deliberately tighter than the admin guard's 5 minutes. There the column is a display value; here it is the input to the idle comparison, and a 5-minute lag would stretch the shortest window by a third before it bit.
- A session whose `lastSeenAt` is null (it predates the column) is stamped and left alive rather than read as "idle since the beginning of time". That is what keeps existing rows from being cut on deploy day as they adopt the 7-day default.

## `requireTwoFactorOnLogin`

When set, a trusted device buys nothing. Honoured in three places: `login` strips the trust cookie without consulting the trusted-device rows, `verifyTwoFactor` refuses to grant trust, and `trustCurrentDevice` rejects outright rather than handing back a device that would still be challenged on the next login. Turning the preference on revokes the devices already trusted, so the player is not shown grants that no longer mean anything.

Turning it on is refused when the account has no second factor enrolled, mirroring how `setLoginWithdrawalAlerts` refuses on an unverified email - otherwise the toggle is a no-op that reads as protection.

## Shared

Both setters mirror `setLoginWithdrawalAlerts`, including its player-only guard - extracted into `assertPlayerPreferenceCaller` now that three setters share it, with the audit denial signal preserved verbatim - and each carries its own audit event so the trail says which preference moved rather than just "a security preference did".

## Tests

New `session-idle.service.int.test.ts` (8 cases: the 7-day default applied and respected, over-window expiry, under-window survival, last-used preservation, first-activity stamping, the system-attributed revoke event, throttle) and a login case asserting a live trusted device is still challenged while the preference holds.

The repo's integration suite is flaky under full-suite parallel load, independently of this change: across five runs a different set of files failed each time (fx, tag, wallet, kyc, chat, social, outbox), every one green in isolation, none touching identity. `pnpm db:clean:test` cleared 184 orphaned per-run databases and brought the count down but did not eliminate it. Both test files above pass inside a full run.


packages/core/src/pam/identity/service/session-idle.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  private async expire(userId: User['id'], sessionId: Session['id']): Promise<void> {
+    await this.drizzle.db
+      .update(session)
+      // Same shape as an explicit revoke: expire in place and leave `updatedAt` alone, so
+      // the device list still shows when the session was last actually used.
+      .set({ expiresAt: sql`now()`, updatedAt: session.updatedAt })
+      .where(eq(session.id, sessionId));
+
+    // No `actorId`: nobody revoked this, the player's own inactivity window did.
+    this.events.emit('identity.session.revoked', {
```
jakubfilinger-b, line 81: This is the gap that matters most. `streamSession` (router/index.ts:145) subscribes to `identity.sessions.revoked_all` only, so nothing pushes on `identity.session.revoked` — the tab that has been sitting idle gets no signal at all. The session is expired in the database, the audit row is written, and the browser carries on showing an authenticated UI until the next request happens to fall through to the unauthenticated path.

For an inactivity timeout that inverts the feature. The scenario it exists for is a player who walked away from a shared machine; "the next request kicks them" is exactly the request that never arrives, so the screen stays logged in for as long as the tab is open.

The gap is pre-existing — a single-session revoke from the device list has the same silence — but this is the feature that makes it load-bearing, so it should be closed here. `streamSession` should also subscribe to `identity.session.revoked` and push `{ type: 'revoked' }` when the event's `sessionId` matches the connection's own. The change-password work already established the pattern of comparing against the id the SSE connection captured at open, so the filter has a precedent to copy.

Worth an e2e alongside it: open a stream, expire that session through `touch`, assert the connection receives `revoked`.

packages/core/src/pam/identity/drizzle/migrations/0014_parallel_moira_mactaggert.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TYPE "public"."auto_logout_duration" AS ENUM('15m', '1h', '24h', '7d', '30d');--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "auto_logout_duration" "auto_logout_duration" DEFAULT '7d' NOT NULL;--> statement-breakpoint
```
jakubfilinger-b, line 2: `auto_logout_duration` lands on `user` with `DEFAULT '7d' NOT NULL`, which means every row gets it — admin accounts included — and the middleware at create-app.ts:472 calls `touch` for every authenticated session without asking whether it is a player's.

So this PR, titled "per-player auto-logout", also ships a seven-day idle logout for backoffice admins. The window is mild and arguably a good thing, but it is a behaviour change on a surface the description says is untouched, and it is asymmetric in a way that will confuse someone later: the setter is behind `assertPlayerPreferenceCaller`, so an admin is pinned to `7d` for ever with no route and no UI to change it. Enforcement everywhere, control for players only.

Two honest options. Either scope the enforcement to players (skip `touch` when the resolved session is not a player's, and say so in the port doc), or own it deliberately — state in the description and the changeset that admin sessions now idle out at seven days, and file the follow-up that gives operators a knob. I lean towards the first for this PR, since admins already have their own `lastSeenAt` tracking through `AdminGuard` and this adds a second mechanism doing a similar job.

packages/core/src/server/runtime/create-app.ts
```diff
@@ -463,7 +464,18 @@ export async function createApp(
...
     }
 
-    const { userId } = resolved;
+    const { userId, sessionId } = resolved;
+
+    // The player's "auto-logout when inactive" window. Checked before the session is
+    // published onto the context, so an idled-out request falls through to the
+    // unauthenticated path above rather than being served and 401ing somewhere deeper.
+    if (sessionId && container.has(SESSION_IDLE_POLICY)) {
+      const state = await container.get(SESSION_IDLE_POLICY).touch(userId, sessionId);
```
jakubfilinger-b, line 473: This `await` is unguarded, and the `PLAYER_ACTIVITY_TRACKER` call twelve lines down is deliberately not: it is fire-and-forget with a `.catch` that logs. So a transient database error inside `touch` now turns every authenticated request into a 500, where the same error in its neighbour is a log line.

Failing closed on a security control is the right instinct and I do not want to swap it for the tracker's swallow-and-continue. But a 500 is the wrong shape of failing closed: it tells the client the server is broken rather than that they are not signed in, and it takes down every authenticated route on a blip in one query. Falling through to the unauthenticated path — the same branch an `expired` result already takes — fails closed *and* degrades honestly.

Whichever we pick, it needs a comment saying it was a choice, because the contrast with the line below reads as an oversight.

packages/core/src/pam/identity/service/session-idle.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  private readonly identityReader: IdentityReader;
+
+  constructor({ drizzle, events, identityReader }: SessionIdleServiceDeps) {
+    this.drizzle = drizzle;
+    this.events = events;
+    this.identityReader = identityReader;
+  }
+
+  async touch(userId: User['id'], sessionId: Session['id']): Promise<'active' | 'expired'> {
+    const [row] = await this.drizzle.db
```
jakubfilinger-b, line 39: Two smaller things on this read.

The joined `select` runs on every authenticated request — the throttle only spares the write. Right below it in the middleware, `PLAYER_ACTIVITY_TRACKER.touchLastSeen` goes to the same `user`/`session` rows for a closely related purpose. That is two round trips per request where the data overlaps almost entirely. Worth checking whether the idle read can return what the tracker needs (or the reverse) before this ships, because it is much harder to merge them once both are established on the hot path.

And `AUTO_LOGOUT_MINUTES[row.autoLogoutDuration]` at :52 has no fallback. Today the column enum and the map come from the same contract constant so they cannot disagree — but if they ever do, `windowMs` is `NaN`, `idleForMs > NaN` is `false`, and the session simply never expires. A security control whose failure mode is silent permanence deserves a default and a log rather than trusting the two definitions to stay in step.

packages/core/src/pam/identity/service/session-idle.service.ts
```diff
@@ -0,0 +1,89 @@
...
+      .select({
+        lastSeenAt: session.lastSeenAt,
+        autoLogoutDuration: user.autoLogoutDuration,
+      })
+      .from(session)
+      .innerJoin(user, eq(user.id, session.userId))
+      .where(and(eq(session.id, sessionId), eq(session.userId, userId)))
+      .limit(1);
+    if (!row) {
```
jakubfilinger-b, line 48: `return 'active'` when the row is missing is failing open, and it is the one branch here with no comment explaining itself. It is reachable: the session can be revoked between better-auth resolving it and this query running, and the caller then publishes the session onto the context and serves the request as authenticated.

Narrow window, and better-auth has already validated the session, so I do not think it is a real hole — but every other decision in this file is annotated with why, and the branch that hands back `active` without evidence is the one that most needs it. Either return `'expired'` here (the caller already handles it, and "no such session" is not a worse answer than "idle") or write down why active is safe.

jakubfilinger-b (commented):
Self-review pass before asking anyone else to read this — GitHub won't take an approval or a changes-request from the author, so these are notes on my own branch rather than a verdict.

Four things I want to fix or justify before this goes out for review. In rough order of how much they change the design:

1. An idled-out tab is never told. `expire` emits `identity.session.revoked`, but `streamSession` only subscribes to `identity.sessions.revoked_all` — so the session dies in the database and the open tab keeps rendering a logged-in UI until someone clicks. For an *inactivity* timeout that is the defining case: the player walked away, and "idle" means the next request is not coming. Thread below.
2. The enforcement is universal but the control is player-only. Thread below.
3. `touch` is awaited unguarded on the request path, where its immediate neighbour is fire-and-forget. Thread below.
4. An extra read per authenticated request that could probably be folded into the tracker's.

On the AC deviation already recorded downstream (no "Never" option): the reasoning still holds — the longest window coincides with better-auth's absolute 30-day expiry, so "Never" would have read as an escape hatch that does not exist. Leaving it as documented rather than reopening it.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -1571,6 +1601,13 @@ export class IdentityService {
       throw new ORPCError('UNAUTHORIZED', { message: 'Not signed in' });
     }
     await assertRateLimit(this.limiter, `trustDevice:${userId}`, TWO_FACTOR_PASSWORD_RATE_LIMIT);
+    // Granting trust the login path is contractually going to ignore would leave the
+    // player with a device listed as trusted that still gets challenged every time.
+    if (await this.requiresTwoFactorEveryLogin(userId)) {
```
mp-blurify, line 1606: **Blocker — an unescapable state.** This guard is correct, but `disableTwoFactor` (`:1777-1782`) tears down `twoFactorMethod`, trusted devices and sessions while leaving `requireTwoFactorOnLogin` set. A player who enables 2FA, turns on "require on every login", then disables 2FA is left with `twoFactorEnabled: false, requireTwoFactorOnLogin: true` — `securityControlsFor` reports enforcement that no longer exists, and this guard then rejects every trust attempt with "Turn off two-factor on every login before trusting this device", so the trust-a-device flow is dead until the player finds and flips a toggle the service itself refuses to let them set in that state (`:1996`).

Fix: clear `requireTwoFactorOnLogin` in the same teardown, next to `setTwoFactorMethod(userId, null)`, and in `AdminSecurityService.resetTwoFactor` (`admin-security.service.ts:361`).

packages/core/src/pam/identity/service/session-idle.service.ts
```diff
@@ -0,0 +1,89 @@
...
+      await this.drizzle.db
+        .update(session)
+        .set({ lastSeenAt: sql`now()` })
+        .where(eq(session.id, sessionId));
+    }
+    return 'active';
+  }
+
+  private async expire(userId: User['id'], sessionId: Session['id']): Promise<void> {
+    await this.drizzle.db
```
mp-blurify, line 73: The expiring `update` is unconditional — no `expiresAt > now()` predicate, no `.returning()` check — so concurrent requests each pass the idle test against the same stale `lastSeenAt`, each write, and each emits `identity.session.revoked`. A returning player whose SPA fires five parallel calls after the window lapsed writes five append-only audit rows into the hash chain for one logout. `revokeSession` (`session.service.ts:186-191`) already guards exactly this with a filtered update plus a `returning` length check.

Rule: `docs/standards/database.md:68` (critical mutations transactional and idempotent — a DB guard, not just a key); `docs/standards/audit.md` (append-only hash chain).
Fix: add `gt(session.expiresAt, sql`now()`)` to the where, `.returning({ id })`, and emit only when a row came back.

packages/core/src/pam/identity/service/session-idle.service.ts
```diff
@@ -0,0 +1,89 @@
...
+      })
+      .from(session)
+      .innerJoin(user, eq(user.id, session.userId))
+      .where(and(eq(session.id, sessionId), eq(session.userId, userId)))
+      .limit(1);
+    if (!row) {
+      return 'active';
+    }
+
+    const windowMs = AUTO_LOGOUT_MINUTES[row.autoLogoutDuration] * 60_000;
```
mp-blurify, line 52: (Anchored here because the affected file is not in this diff — the blast radius is `session.service.ts:116`.)

Idle expiry is lazy, computed from this window only when a request arrives. But `listAllActiveSessions` (`session.service.ts:116`) — the Super Admin "who is logged in right now" read — still derives liveness from `expiresAt` alone, and `listSessions`'s `activeOnly` filter (`:68`) does the same. So a player on a `15m` window who closes their laptop keeps showing as an active session for up to the full 30-day `expiresAt`, and the same row is reported as current in the player's own device list.

Rule: blast radius — a changed column's other queries must still hold.
Fix: add `lastSeenAt IS NULL OR lastSeenAt > now() - user.autoLogoutDuration` to both active-session filters via the joined `user` row.

packages/core/src/pam/identity/service/session-idle.service.ts
jakubfilinger-b: Fixed. `streamSession` (router/index.ts) now also subscribes to `identity.session.revoked` and pushes `{ type: 'revoked' }` when the event's `sessionId` matches the connection's own — same id-comparison pattern the change-password work established for the SSE stream.

packages/core/src/pam/identity/drizzle/migrations/0015_parallel_moira_mactaggert.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TYPE "public"."auto_logout_duration" AS ENUM('15m', '1h', '24h', '7d', '30d');--> statement-breakpoint
+ALTER TABLE "user" ADD COLUMN "auto_logout_duration" "auto_logout_duration" DEFAULT '7d' NOT NULL;--> statement-breakpoint
```
jakubfilinger-b: Went with the first option. `SessionIdleService.touch` now reads `user.role` off the joined row and returns `'active'` immediately for anything that is not `'player'`, so the column landing on every row no longer means enforcement on every row. Admins keep their existing `AdminGuard`/`AdminSecurityService.touchLastSeen` mechanism, unchanged.

packages/core/src/server/runtime/create-app.ts
jakubfilinger-b: Went with falling through to the unauthenticated path, matching the expired-session branch: `touch` is now wrapped in try/catch, logs the error, and treats the failure as `'expired'` rather than throwing. Commented at the call site so the contrast with the tracker's fire-and-forget read isn't mistaken for an oversight.

packages/core/src/pam/identity/service/session-idle.service.ts
jakubfilinger-b: Fixed the fallback: an unmapped `autoLogoutDuration` now logs an error and expires the session rather than producing `NaN`/never-expiring behaviour. Left the two-round-trip merge with `PLAYER_ACTIVITY_TRACKER.touchLastSeen` as a follow-up rather than folding it into this PR — it's a real optimization but touches the tracker's own contract, and I'd rather do that as its own reviewable change than bundle it here.

packages/core/src/pam/identity/service/session-idle.service.ts
jakubfilinger-b: Went with returning `'expired'` — the caller already handles it and, as you said, "no such session" is not a worse answer than "idle". Added a comment explaining why: reporting active here would publish a session onto the context that this same read could not find.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Fixed exactly as suggested: `disableTwoFactor` now also clears `requireTwoFactorOnLogin` in the same teardown as `setTwoFactorMethod(userId, null)`, and `AdminSecurityService.resetTwoFactor`'s transactional update does the same.

packages/core/src/pam/identity/service/session-idle.service.ts
jakubfilinger-b: Fixed by mirroring `revokeSession` exactly: the expiring update now carries `gt(session.expiresAt, sql`now()`)` in the where clause plus a `.returning({ id })` check, and the `identity.session.revoked` event only fires when a row actually came back.

packages/core/src/pam/identity/service/session-idle.service.ts
jakubfilinger-b: Fixed with a shared `notIdledOut()` SQL helper, built dynamically off `AUTO_LOGOUT_MINUTES` (via `sql.join`) so the duration-to-minutes mapping can't drift from the one place it's actually defined. Wired into both `listSessions`'s `activeOnly` filter and `listAllActiveSessions`, each now joined against `user` for the check.

packages/core/src/pam/identity/router/index.ts
```diff
@@ -136,20 +144,31 @@ export function createIdentityRouter(
...
           const unsubscribeRevoked = eventBus.on('identity.sessions.revoked_all', (event) => {
             if (event.userId === userId) {
               push({ type: 'revoked' });
             }
           });
+          // A single-session revoke - including the idle-timeout expiry `SessionIdleService`
+          // performs - only ever names one session. Without this, the tab it just killed
+          // gets no signal at all: the DB write and the audit row already happened, but the
+          // open tab keeps rendering as signed in until its next request happens to 401.
+          const unsubscribeSessionRevoked = eventBus.on('identity.session.revoked', (event) => {
```
damianrzepka, line 159: `EventBus` korzysta tu z Redis Streams, gdzie repliki tej samej usługi współdzielą consumer group. W efekcie `identity.session.revoked` trafi tylko do jednej repliki; jeśli SSE tego gracza jest obsługiwane przez inną, jego zakładka nie dostanie `revoked` po idle-timeoucie. Ten sygnał powinien iść przez `REALTIME_TRANSPORT` (albo inny broadcast do wszystkich replik), nie przez broker zdarzeń.

damianrzepka (approved):

PR #159, opened 2026-09-10
feat(iam): add a swap-config admin permission resource

zaxovaiko:
## Summary

Adds a `swap-config` admin permission resource (`view`, `update`), granted to the built-in admin role and the seeded Finance / Accounting role.

## Why

A screen that sets swap pricing had to gate on `wallet-asset`, so anyone allowed to edit the asset catalog could also change what players pay to swap. Pricing is a finance decision, the asset catalog is an ops one.

## Alternatives considered

- Keep gating swap pricing on `wallet-asset`. It couples two unrelated duties in one grant.

## Risks

- Custom roles do not get the resource automatically. A consumer that moves its swap settings screen onto `swap-config` has to grant it to those roles first.

### DIFF
--- a/packages/core/src/iam/seed/data/default-admin-roles.ts
+++ b/packages/core/src/iam/seed/data/default-admin-roles.ts
@@ -198,6 +198,7 @@ export const DEFAULT_ADMIN_ROLES: readonly DefaultAdminRole[] = [
       audit: R,
       analytics: R,
       'wallet-asset': RW,
+      'swap-config': RW,
     },
   },
   {
--- a/.changeset/swap-config-permission.md
+++ b/.changeset/swap-config-permission.md
@@ -1,0 +1,5 @@
+---
+'@openora/core': minor
+---
+
+New admin permission resource `swap-config` (`view`, `update`) for the screen that sets a swap desk's pricing. Until now an overlay that owned swap pricing had to gate it on `wallet-asset`, so anyone allowed to edit the asset catalog could also change what players pay to swap. The built-in `admin` role and the seeded Finance / Accounting role are granted it; every other seeded role starts with no access.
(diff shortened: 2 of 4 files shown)


packages/core/src/iam/seed/data/default-admin-roles.ts
```diff
@@ -198,6 +198,7 @@ export const DEFAULT_ADMIN_ROLES: readonly DefaultAdminRole[] = [
       audit: R,
       analytics: R,
       'wallet-asset': RW,
+      'swap-config': RW,
```
jakubfilinger-b, line 201: This only reaches an existing operator when `seedRoles` runs again — there's no migration behind it. The changeset's Risks section covers custom roles but not this: on an already-seeded database the Finance / Accounting role keeps the permission set it has, so the operator upgrades, moves their swap screen onto `swap-config`, and their finance team is locked out of it until someone re-seeds.

And re-seeding isn't free: `seedRoles` upserts with `set: { level: excluded.level }` across every seeded role, so it also reverts any permission an operator has deliberately tightened or loosened on the other system roles since install. Worth saying both things in the changeset — "re-run the seed to pick this up, and note it re-asserts the shipped matrix for every system role" — so nobody discovers the second half after the fact.

jakubfilinger-b (dismissed):
Separating swap pricing from the asset catalog is the right call — those really are two different duties, and gating a pricing screen on `wallet-asset` meant ops could change what players pay to swap. The change is minimal and lands in all three places it has to (`adminStatement`, the built-in `adminRole`, the seed matrix), so `permissions.test.ts` will keep the catalog and the fallback role from drifting apart.

Approving. One rollout question on the thread that I'd like answered in the changeset rather than in code.

PR #160, opened 2026-09-10
fix(wallet): check address-credited deposits against the player's RG deposit limit

zaxovaiko:
## Summary

A deposit credited by address (the custody webhook and poller path) is now checked against the player's responsible-gambling deposit limit. The credit still happens, and an over-limit deposit files a `rg_limit_breach` reconciliation finding with the limit, the period and what was already used, plus the usual `wallet.reconciliation_finding.recorded` audit row.

## Why

The PSP path refuses an over-limit deposit before charging. The address path credited any amount with no check at all, so a player with a daily limit could deposit any amount on chain and nothing recorded that the limit had been passed. An operator had no list of deposits to return the excess on.

## Alternatives considered

- Refusing the credit. The funds are already on chain when the webhook lands, so refusing would leave the player's money outside the ledger.
- Holding the deposit in a pending state until an operator releases it. Needs a new state and a release flow; the finding gives operators the same list without either.

## Risks

- Adds `rg_limit_breach` to `WALLET_RECONCILIATION_FINDING_KINDS`; the wallet migration `0020` adds it to the finding kind enum.
- The limit check runs before the credit transaction and never blocks it. If the check itself fails, the error is logged and no finding is filed, so an outage of the limits read means a missed finding rather than a stuck deposit.
- A replayed webhook for an already-credited deposit does not file a second finding.


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2686,6 +2687,15 @@ export class WalletService {
       return;
     }
 
+    // The funds are already on chain, so the player's deposit limit cannot refuse this credit
+    // the way it refuses a PSP charge. It is still asked, with this deposit as the attempted
+    // move, and a refusal becomes a finding so a human can act on the excess.
+    const rgDecision = await this.rgDecisionForLandedDeposit(
```
jakubfilinger-b, line 2693: The comment explains why this sits outside the credit transaction and I agree with the reasoning — a gate failure must not roll back on-chain funds. But it also reads `used` on `this.drizzle.db` *before* the transaction opens, and the sibling path does the opposite: `assertWithinRgDepositLimit` is handed `txn` and called from inside the deposit transaction at :961.

That asymmetry has a concrete consequence. Two deposits to the same player landing close together both compute `used` from the same pre-credit snapshot, so each is judged on its own against the limit. A player with a daily limit of 1 who lands 0.6 and 0.6 within the window has both credited, both judged allowed, and no finding filed — even though the day closes at 1.2 against a limit of 1. That is precisely the row an operator needs to return the excess, and it is the case a determined player would produce deliberately.

It also makes the numbers in the finding itself unreliable: `used` is the value from before this credit and before any concurrent one, so the `(X already used)` in `detail` can already be wrong by the time the row is written.

You don't need the check inside the transaction to fix it — you need it *after* the credit commits, reading post-credit state. Move the call below the transaction (still guarded by `!replayed`) and ask the gate about the state as it now stands. A gate failure there still cannot roll anything back, which is the property the comment is protecting, and the answer stops depending on how many deposits raced.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2745,6 +2755,26 @@ export class WalletService {
...
+            kind: 'rg_limit_breach',
+            currency: event.currency,
+            network: event.network ?? depositAddress.network,
+            amount: event.amount,
+            address: event.address,
+            tag: event.tag ?? null,
+            txHash: event.txHash,
+            externalId: event.externalId,
+            transactionId,
+            detail: `credited over the player's ${rgDecision.period} ${rgDecision.limitType} limit of ${rgDecision.limit} (${rgDecision.used} already used) - the funds were already on chain`,
```
jakubfilinger-b, line 2773: The changeset promises the finding carries "the limit, the period and what was already used", but all three exist only inside this English sentence. `wallet_reconciliation_finding` has no metadata column, so `limit`, `period`, `limitType` and `used` are unreachable to anything but a human reading `detail` — an operator UI cannot sort or filter by them, and the excess-return workflow this row exists to trigger has to regex the string to learn how much the excess actually was. The test pins the prose (`toContain('daily deposit limit of 1')`), which will also make the wording load-bearing the first time someone rephrases it.

The table already carries `amount` (the deposit) and `currency`; what is missing is the limit side. Either add a `metadata jsonb` column in the same migration that adds the enum value and put the decision in it, or if that is too much for this PR, say so in the changeset rather than implying the fields are queryable.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2755,6 +2785,27 @@ export class WalletService {
...
   }
 
+  // Outside the credit transaction on purpose: a gate failure must never roll back a credit
+  // for funds that already arrived, and a finding is a report, so it needs no lock.
+  private async rgDecisionForLandedDeposit(
+    userId: User['id'],
+    amount: string,
+    currency: string,
+  ): Promise<RgLimitDecision | null> {
+    if (!this.rgLimits) {
```
jakubfilinger-b, line 2795: An unbound `RG_LIMITS` returns `null` here and the deposit is credited with no check and no trace — same as before this PR, which is fair, but it is now the one path where a licence-facing gate can be absent and nothing anywhere says so. `assertWithinRgDepositLimit` at :799 has the identical early return, so an operator who never binds the port has both deposit routes silently ungated.

Not this PR's job to make the port required, but a one-time warning at plugin bind time ("wallet loaded without RG_LIMITS - deposit limits are not enforced") would turn a silent condition into a visible one, and it is the kind of thing that gets asked about in an audit. Worth a follow-up issue if not here.

Separately: the `try/catch` logs `userId` and `currency` but not the `externalId` or `txHash`, so the log line cannot be tied back to the deposit it failed on. Those are already in scope at the call site — worth threading through.

packages/core/src/wallet/__tests__/wallet.service.int.test.ts
```diff
@@ -1808,6 +1808,90 @@ describe('WalletService.creditDepositByAddress (real PG)', () => {
...
+      currency: 'BTC',
+      externalId,
+      txHash: '0xoverlimit',
+    });
+
+    expect(await balanceOf(w.userId)).toBe(2);
+    expect(emittedTopics(events)).toEqual(['wallet.deposit.completed']);
+    expect(rgLimits.checkDeposit).toHaveBeenCalledWith(expect.anything(), w.userId, '2', 'BTC');
+    const findings = await findingsFor(externalId);
+    expect(findings).toHaveLength(1);
```
jakubfilinger-b, line 1848: Good set — over-limit, under-limit and gate-down are the three that matter, and asserting the balance in all three is what makes the "credit always happens" claim real rather than stated.

Worth one more, given the `(kind, externalId)` partial unique index and `onConflictDoNothing` in `recordReconciliationFinding`: a second over-limit deposit for the same player with a *different* `externalId` should file a second finding. The dedup key makes replay safe, which is the behaviour the changeset promises, but nothing currently pins that it does not also swallow a genuinely new breach.

jakubfilinger-b (changes requested):
The gap is real and the shape of the answer is right: the funds are on chain, so refusing is not on the table, and a finding plus an audit row is the honest thing to leave behind. Crediting first and reporting second, with the check outside the credit transaction so a gate failure can never roll back money that already arrived, is the correct trade — and the three tests cover exactly the three outcomes that matter.

One blocking item and two smaller ones. The blocking one is the read placement: the sibling PSP path runs the same check *inside* its transaction and this one does not, and the difference is load-bearing under concurrency. Detail on that thread.

A correction for the description and changeset while you're in there: they both say "the custody webhook and poller path", but `creditDepositByAddress` has exactly one caller — the webhook route in `wallet/router/index.ts` — and `reconciliation-finding.service.ts` says as much in its own doc comment ("hits an unattributable deposit in real time, not on a poll"). I checked the other `creditWalletBalance` call sites: the PSP deposit is already gated, and the rest are `manualAdjust`, swap legs and withdrawal refunds, none of which are player deposits. So the coverage is right — it's just the sentence that claims more than it does, and this one lands in an operator-facing changeset.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2745,6 +2755,26 @@ export class WalletService {
     });
 
     if (!replayed) {
+      if (rgDecision && !rgDecision.allowed) {
+        await recordReconciliationFinding(
```
klaudia-blazyczek-blurify, line 2759: If `recordReconciliationFinding` fails after the credit commits, the webhook fails but the money is already credited. A retry sets `replayed` and skips this block, so the over-limit deposit can permanently miss its finding and audit trail. Please make the report retryable on replays, or persist it with a retry-safe outbox.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2745,6 +2755,26 @@ export class WalletService {
...
+            runId: LIVE_WEBHOOK_RUN_ID,
+            providerName: depositAddress.providerName,
+            kind: 'rg_limit_breach',
+            currency: event.currency,
+            network: event.network ?? depositAddress.network,
+            amount: event.amount,
+            address: event.address,
+            tag: event.tag ?? null,
+            txHash: event.txHash,
+            externalId: event.externalId,
```
klaudia-blazyczek-blurify, line 2771: `externalId` is only unique within a provider, but the finding index deduplicates only by `kind` and `externalId`. Two providers can reuse an ID: both deposits are credited, but the second RG breach finding is silently dropped. Please include `providerName` in the finding deduplication key.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Fixed. The dedup index is now `(kind, providerName, externalId)`, migration `0021_mean_chameleon`, and the schema comment says why. It applies to every finding kind, not just this one - the old key would have swallowed a second vendor's `missing_deposit` or `stuck_sweep` the same way. The reconciliation cycle's two already-reported checks now match the provider as well, so the ordering hint keeps mirroring the real key (the withdrawal one coalesces to `DEFAULT_PAYMENT_PROVIDER`, matching what the loop files). New test credits the same `externalId` through two providers and expects two findings.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `7b506850`. The gate now runs after the transaction commits, with `attempted: 0` - the committed deposit row is already inside the window the gate counts, so asking with `event.amount` would count it twice. Two deposits landing together are now judged against the same post-credit window instead of each against its own pre-credit snapshot, and the `used` in `detail` is the value as of the write. A failure there still cannot roll back on-chain funds, which was the property the original comment was protecting.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in the same change. The block is deliberately not guarded by `replayed`: a failed finding write is exactly what makes the vendor retry the webhook, and the retry sees the credit as a replay. Since the dedup index is what stops a duplicate, the retry fills the gap instead of filing twice. Test `files the finding on a replayed webhook whose first attempt never recorded one` pins it.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Took the second option rather than adding the column - the changeset now states plainly that the limit side is prose in `detail`, readable and not queryable, instead of implying the fields are structured. A `metadata jsonb` column is a table-wide change worth its own PR, since every finding kind would want to use it.

The prose-pinning point stands and I have left it: the two `toContain` assertions are deliberately on the numbers (`daily deposit limit of 1`), not the sentence, so a rewording that keeps the limit and period visible still passes.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Both halves done. The `try/catch` now logs `externalId` and `txHash` through a `context` argument, so a failed check ties back to the deposit it failed on.

Took the bind-time warning here rather than deferring it: the wallet plugin now resolves `RG_LIMITS` once and logs `wallet loaded without RG_LIMITS - deposit limits are not enforced on any route` when it is absent. The port stays optional, so behaviour is unchanged - the silent condition is just no longer silent.

packages/core/src/wallet/__tests__/wallet.service.int.test.ts
zaxovaiko: Added, it is `files a second finding for a second over-limit deposit under its own externalId`: two over-limit deposits for the same player under different vendor ids, expecting a finding for each. Your point about the dedup key turned out to have a second half, which the later review caught - the key also needed `providerName`, so there is now a companion test for two vendors reusing one id.

PR #162, opened 2026-09-10
fix(wallet): audit swaps, compare them exactly, and hand the desk the player

zaxovaiko:
## Summary

- A settled swap records `wallet.swap.completed` and a returned hold records `wallet.swap.refunded`, each in the same transaction as its ledger rows.
- The idempotent-replay, insufficient-balance and fill-amount checks compare with `moneyEquals`/`moneyCompare` instead of `Number()`.
- `SwapAdapter.getQuote` and `SwapAdapter.execute` receive `userId`, so a desk can bind a quote to the player who asked for it and refuse it from anyone else.

## Why

A swap moved two balances with no audit row at all, so there was nothing to reconstruct who converted what on a real-money path. The replay check compared amounts as doubles, so two 18-decimal amounts that collapse to the same double passed as the same swap. And `SwapService.quote` had the player but dropped it at the port, so a desk had no way to tie a signed quote to the player it was issued to.

## Alternatives considered

- Writing the audit row from each swap adapter. Every desk would reimplement it, and the row would not share the ledger transaction.

## Risks

- `userId` is a new field on the adapter inputs. Adapters that ignore it keep compiling.
- A fill amount that is not a money string now reads as "no fill" and leaves the swap `processing` instead of reaching the ledger. #161 reports those from the reconciliation cycle.

### DIFF
--- a/packages/core/src/wallet/service/swap.service.ts
+++ b/packages/core/src/wallet/service/swap.service.ts
@@ -4,14 +4,17 @@ import {
   createLogger,
   findOneOrThrow,
   makeConflictError,
-  moneyToNumber,
+  moneyCompare,
+  moneyEquals,
   type DrizzleService,
   type EventBus,
   type DrizzleTx,
 } from '@openora/core/server';
 import {
   makeRateLimitKey,
+  MoneyAmountSchema,
   RATE_LIMIT_KEYS,
+  type AuditWritePort,
   type PlatformConfig,
   type RateLimitKey,
   type RateLimiterAdapter,
@@ -70,6 +73,12 @@ const SwapLegMetadataSchema = z.object({
   quoteId: z.string().optional(),
 });
 
+// A vendor amount is untrusted input: a malformed string must read as "no fill", never
+// reach the ledger, and never be compared as a float.
+function isPositiveMoney(value: string): boolean {
+  return MoneyAmountSchema.safeParse(value).success && moneyCompare(value, '0') > 0;
+}
+
 export type SwapResult = {
   transactionId: WalletTransaction['id'];
   status: WalletTransaction['status'];
@@ -93,6 +102,7 @@ export class SwapService {
   private readonly drizzle: DrizzleService;
   private readonly events: EventBus;
   private readonly adapter: SwapAdapter;
+  private readonly audit: AuditWritePort;
   private readonly platformConfig?: PlatformConfig;
   private readonly limiter?: RateLimiterAdapter<RateLimitKey>;
 
@@ -100,12 +110,14 @@ export class SwapService {
     drizzle: DrizzleService;
     events: EventBus;
     adapter: SwapAdapter;
+    audit: AuditWritePort;
     platformConfig?: PlatformConfig;
     limiter?: RateLimiterAdapter<RateLimitKey>;
   }) {
     this.drizzle = deps.drizzle;
     this.events = deps.events;
     this.adapter = deps.adapter;
+    this.audit = deps.audit;
     this.platformConfig = deps.platformConfig;
     this.limiter = deps.limiter;
   }
@@ -126,7 +138,7 @@ export class SwapService {
   }): Promise<SwapQuote> {
     await this.rateLimit(userId);
     this.assertPair(input.fromCurrency, input.toCurrency);
-    const quote = await this.adapter.getQuote(input);
+    const quote = await this.adapter.getQuote({ ...input, userId });
     if (!quote) {
       throw new SwapPairUnsupportedError(input.fromCurrency, input.toCurrency);
     }
@@ -203,7 +215,7 @@ export class SwapService {
           readWalletBalance(txn, current.id, fromCurrency),
           readLockedBonusAmount(txn, current.id, fromCurrency),
         ]);
-        if (moneyToNumber(available) < moneyToNumber(fromAmount)) {
+        if (moneyCompare(available, fromAmount) < 0) {
           throw new InsufficientBalanceError(available, fromAmount);
         }
         throw new BonusRolloverLockedError(locked);
@@ -221,6 +233,7 @@ export class SwapService {
     let execution;
     try {
       execution = await this.adapter.execute({
+        userId,
         quoteId,
         fromCurrency,
         toCurrency,
@@ -282,10 +295,11 @@ export class SwapService {
     externalId: SwapExecution['externalId'],
     toAmount: string | undefined,
   ): Promise<SwapResult> {
-    if (toAmount === undefined || Number(toAmount) <= 0) {
+    if (toAmount === undefined || !isPositiveMoney(toAmount)) {
       throw new SwapFillAmountMissingError(externalId);
     }
     const toCurrency = this.toCurrencyOf(out);
+    const userId = await this.userIdForWallet(out.walletId);
 
     const settled = await this.drizzle.db.transaction(async (txn) => {
       const flipped = await txn
@@ -310,12 +324,26 @@ export class SwapService {
         // belongs to the out-leg. The in-leg points back at its own pair instead.
         metadata: JSON.stringify({ swapTransactionId: out.id, externalId }),
       });
+      await this.audit.recordInTransaction(txn, {
+        actorType: 'system',
+        action: 'wallet.swap.completed',
+        resourceType: 'wallet_transaction',
+        resourceId: out.id,
+        after: {
+          userId,
+          fromCurrency: out.currency,
+          fromAmount: out.amount,
+          toCurrency: balanceKey(toCurrency),
+          toAmount,
+          externalId,
+        },
+      });
       return true;
     });
 
     if (settled) {
       this.events.emit('wallet.swap.completed', {
-        userId: await this.userIdForWallet(out.walletId),
+        userId,
         transactionId: out.id,
         fromCurrency: out.currency,
         fromAmount: out.amount,
@@ -333,6 +361,7 @@ export class SwapService {
     out: WalletTransaction,
     externalId?: SwapExecution['externalId'],
   ): Promise<void> {
+    const userId = await this.userIdForWallet(out.walletId);
     await this.drizzle.db.transaction(async (txn) => {
       const flipped = await txn
         .update(walletTransaction)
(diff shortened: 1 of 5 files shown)


packages/core/src/wallet/service/swap.service.ts
```diff
@@ -333,6 +361,7 @@ export class SwapService {
     out: WalletTransaction,
     externalId?: SwapExecution['externalId'],
   ): Promise<void> {
+    const userId = await this.userIdForWallet(out.walletId);
```
jakubfilinger-b: This is the one that worries me most. `refund` is the failure path — it runs from the `catch` around `adapter.execute` and from `status === 'failed'`, and its job is to give the player back a hold that has already been debited. Putting `userIdForWallet` in front of the transaction means a `WalletNotFoundError` (or any transient error on that read) now throws *before* the refund transaction opens, so the hold is never returned and the original vendor error is swallowed and replaced by this one. The player is down the money and the error the operator sees points at the wrong thing.

The lookup exists only to populate `after.userId` on the audit row. Read it inside the transaction on `txn` instead — same connection, joins the same commit, and a failure there rolls the whole thing back cleanly rather than skipping it. Or drop `userId` from the payload: `resourceId` is the transaction id and the wallet is one join away, so the row is not much less useful without it.

packages/core/src/wallet/service/swap.service.ts
```diff
@@ -282,10 +295,11 @@ export class SwapService {
     externalId: SwapExecution['externalId'],
     toAmount: string | undefined,
   ): Promise<SwapResult> {
-    if (toAmount === undefined || Number(toAmount) <= 0) {
+    if (toAmount === undefined || !isPositiveMoney(toAmount)) {
       throw new SwapFillAmountMissingError(externalId);
     }
     const toCurrency = this.toCurrencyOf(out);
+    const userId = await this.userIdForWallet(out.walletId);
```
jakubfilinger-b: Same shape as the `refund` one, different consequence. This used to be awaited after the commit, inside `if (settled)`, purely to build the event payload — so a failed lookup cost you an event. Now it gates the settlement itself: a vendor has confirmed a fill, and a failed read of `wallet.userId` means the in-leg is never credited and the out-leg stays `processing`, with the player's debit standing.

On the webhook route (`reconcileSwapStatus`) it also turns into a retry loop, since the handler throws and the vendor re-delivers. Move it onto `txn` inside the transaction.

packages/core/src/wallet/service/swap.service.ts
```diff
@@ -282,10 +295,11 @@ export class SwapService {
     externalId: SwapExecution['externalId'],
     toAmount: string | undefined,
   ): Promise<SwapResult> {
-    if (toAmount === undefined || Number(toAmount) <= 0) {
+    if (toAmount === undefined || !isPositiveMoney(toAmount)) {
```
jakubfilinger-b, line 292: This branch throws before the transaction opens, so the out-leg stays `processing` with the player's balance already debited and — unlike every other terminal path in this file now — no audit row at all. That is the exact gap the PR sets out to close ("a swap moved two balances with no audit row"), still open on the one branch where money is actually stranded.

I'm not asking for an automatic refund here; the reasoning in #161 for not refunding is sound. But a `wallet.swap.fill_rejected` row (externalId, the raw `toAmount` the desk sent, the out-leg id) would give the desk something to work from, and it costs one `record` call. As it stands the only trace is a thrown error and, eventually, #161's `stuck_swap` finding, which says the swap is stuck but not that the vendor sent us garbage.

packages/core/src/wallet/service/swap.service.ts
```diff
@@ -343,6 +372,18 @@ export class SwapService {
         return;
       }
       await creditWalletBalance(txn, out.walletId, out.currency, out.amount);
+      await this.audit.recordInTransaction(txn, {
+        actorType: 'system',
+        action: 'wallet.swap.refunded',
```
jakubfilinger-b, line 393: `wallet.swap.refunded` isn't in `domainEventSchemas` and isn't added to `DirectAuditAction` in `contracts/adapters/audit.ts`, so it resolves through the `(string & {})` arm — which is the arm whose own doc comment says it "constrains nothing at runtime, it just guides", and `DirectAuditAction` says "Add new direct actions here". Please add it, so the action shows up in autocomplete and anyone grepping the union sees swaps are audited.

`wallet.swap.completed` is fine as-is since it is already a declared event name — but that also makes it ambiguous: it is now both a topic the bus emits and a direct action written in-transaction. `AUDIT_TOPICS` doesn't subscribe to it today, so there's no double row; if someone adds it to that list later there will be, silently. Worth a one-line comment here saying the direct write is deliberate (money-path records join the business transaction) and the topic must stay unsubscribed.

packages/core/src/wallet/service/swap.service.ts
```diff
@@ -70,6 +73,12 @@ const SwapLegMetadataSchema = z.object({
   quoteId: z.string().optional(),
 });
 
+// A vendor amount is untrusted input: a malformed string must read as "no fill", never
+// reach the ledger, and never be compared as a float.
+function isPositiveMoney(value: string): boolean {
```
jakubfilinger-b: Small one: `isPositiveMoney` is exactly the guard the deposit and withdrawal paths need for a vendor-supplied amount too, and it is defined as a file-local function here. If there is a shared home for it (`server`'s money helpers, next to `moneyCompare`), that is where the next person will look for it rather than reimplementing `Number(x) > 0` in the next adapter.

jakubfilinger-b (changes requested):
Three good fixes in one PR and I'd take all three in principle. The `Number(toAmount) <= 0` one is worth calling out explicitly, because it was worse than the description says: `Number('NaN') <= 0` is `false`, so a vendor fill of `NaN` did not fail the guard — it went straight on to `creditWalletBalance`. `isPositiveMoney` closes that, and the test pins it.

What I can't sign off on yet is the placement of the two new `userIdForWallet` calls: both sit outside the transaction, on the failure and settlement paths, and both can now abort a money movement to fill in an audit field. Details on those threads.

One scoping note: `getQuote`/`execute` gaining a required `userId` is fine for an adapter *implementing* the port (parameter bivariance, as you say), but anything that *constructs* those inputs — an overlay calling the adapter directly, a consumer's fake in a test — stops compiling. That reads as a `major` to me rather than the `minor` in the changeset, or at least deserves a line in it telling consumers what breaks.

packages/core/src/wallet/service/swap.service.ts
zaxovaiko: Agreed, and this was the one worth blocking on. Moved onto `txn` inside the refund transaction, so the read joins the same commit as the credit and a failure there rolls back with it instead of ahead of it. The vendor error now reaches the caller unchanged. 8c6bef10

packages/core/src/wallet/service/swap.service.ts
zaxovaiko: Same fix. The lookup runs on `txn` now, and the transaction returns the userId so the post-commit event still has it without a second read. A failed read rolls the credit back rather than leaving the in-leg uncredited, and the webhook no longer retries forever on it. 8c6bef10

packages/core/src/wallet/service/swap.service.ts
zaxovaiko: Fair, and cheap. The branch now records `wallet.swap.fill_rejected` with the externalId, the raw `toAmount` the desk sent, and the out-leg id, before throwing. Still no refund - your reasoning in #161 holds - but the desk can see the vendor sent garbage rather than only that the swap is stuck. Covered by the `NaN` test, which also pins the out-leg staying `processing`. 8c6bef10

packages/core/src/wallet/service/swap.service.ts
zaxovaiko: Added both `wallet.swap.refunded` and the new `wallet.swap.fill_rejected` to `DirectAuditAction`. The `wallet.swap.completed` write carries your one-liner: the direct write is deliberate because a money-path record has to commit with its ledger rows, and the topic has to stay out of the audit plugin list or the swap gets two rows. Named the list `SUBSCRIBED_TOPICS` in the comment, which is what it is called in `audit/plugin.ts`. 8c6bef10

packages/core/src/wallet/service/swap.service.ts
zaxovaiko: Moved to `server/db/query-helpers.ts` next to `moneyCompare` and exported, so that is where the next person finds it. One correction on the motivation: the deposit and withdrawal paths do not have this hole. Their guards are zod refinements on request input behind `MoneyAmountSchema`, whose regex already rejects `NaN` before the `Number(v) > 0` refine runs. The hole was specific to a vendor amount arriving past the contract boundary, so I did not rewire those. 8c6bef10

zaxovaiko:
All five threads addressed, plus the scoping note. Pushed 8c6bef10 and 0aaa3cf3.

- Both `userIdForWallet` calls moved onto `txn`, inside the transaction they feed. No money movement is gated on a read that only fills in an audit field any more.
- The rejected-fill branch records `wallet.swap.fill_rejected` before throwing, so the one path that deliberately strands a hold leaves a trace of what the desk actually sent.
- `wallet.swap.refunded` and `wallet.swap.fill_rejected` declared in `DirectAuditAction`; the direct `wallet.swap.completed` write says why it is direct and that its topic must stay out of `SUBSCRIBED_TOPICS`.
- `isPositiveMoney` lives next to `moneyCompare` in the money helpers now.
- Changeset is `major`, with the adapter `userId` break called out as you asked and one you did not catch: `SwapService` now requires an `audit` dependency, which breaks anything constructing it outside the wallet plugin.

The swap routes had no route-level test at all, so the fixes are pinned by one: `swap.execute` over a real database, asserting the ledger legs, the balances, the audit row and the response. `pnpm verify` green including `check:drift`.

0aaa3cf3 is unrelated to the wallet code - it tightens the fix-review-comments workflow in `docs/standards/skills/delivery.md` to require exactly that end-to-end run before a comment on a money or audit path is reported as fixed. Happy to split it out if you would rather keep this PR single-purpose.

PR #163, opened 2026-09-10
fix(wallet): read auto-withdrawal thresholds and caps in the fx pivot

zaxovaiko:
## Summary

Auto-withdrawal thresholds, the daily amount cap and the large-amount heuristic are read in the fx pivot currency (`exchangeRate.pivot`, `USD` by default) through `EXCHANGE_RATE_READER`, and compared exactly instead of with `Number()`. With no rate for the withdrawal's currency, or no fx module loaded, the withdrawal is not auto-approved.

## Why

One `cryptoThreshold` serves every crypto asset but was compared raw against the withdrawal amount, so its meaning depended on the coin: a threshold of `1` written with a stablecoin in mind let `0.4 BTC` auto-approve. The daily amount cap summed trailing payouts across currencies as if they were one unit.

## Alternatives considered

- A threshold per asset. More config for every asset added, and the daily cap would still need a common unit to sum in.

## Risks

- **Behaviour change:** `fiatThreshold`, `cryptoThreshold`, per-player `auto_withdrawal_rule.threshold` and `autoWithdrawal.dailyCapAmount` are pivot amounts. A threshold configured in coin units reads as a tiny pivot amount after upgrade, so withdrawals route to manual review until an operator re-enters it. That is the safe direction, but operators need to know before they upgrade.
- A rate outage stops auto-approval for the affected currency; those withdrawals queue for manual review.
- The review-queue `large_amount` tag still compares the raw amount. It is a display hint, not a decision.

### DIFF
--- a/.changeset/currency-aware-auto-withdrawal.md
+++ b/.changeset/currency-aware-auto-withdrawal.md
@@ -4,8 +4,14 @@
 
 Auto-withdrawal thresholds, the daily amount cap and the large-amount heuristic are now read in the fx pivot currency (`exchangeRate.pivot`, `USD` by default) instead of compared raw against whatever currency the withdrawal is in.
 
-One `cryptoThreshold` serves every crypto asset, so a raw comparison made its meaning depend on the coin: a threshold of `1` written with a stablecoin in mind let `0.4 BTC` auto-approve. The daily amount cap summed trailing payouts across currencies as if they were one unit. Both now convert through `EXCHANGE_RATE_READER`, and every comparison is exact decimal instead of `Number()`.
+One `cryptoThreshold` serves every crypto asset, so a raw comparison made its meaning depend on the coin: a threshold of `1` written with a stablecoin in mind let `0.4 BTC` auto-approve. The daily amount cap summed trailing payouts across currencies as if they were one unit. The withdrawal is now converted once through `EXCHANGE_RATE_READER`, outside the advisory lock, and every comparison is exact decimal instead of `Number()`.
 
-**Behaviour change:** `fiatThreshold`, `cryptoThreshold`, per-player `auto_withdrawal_rule.threshold` and `autoWithdrawal.dailyCapAmount` are pivot amounts. A threshold configured in coin units reads as a tiny pivot amount after upgrade, so withdrawals route to manual review until an operator re-enters it - the safe direction. When no rate is available for the withdrawal's currency, or the fx module is not loaded, the withdrawal is not auto-approved.
+Each auto-approved payout stores its pivot value in the new `wallet_transaction.auto_approval_pivot_amount` column, and the daily amount cap sums those stored values. A rate move after approval cannot shrink what a player already withdrew, and no rate is read while the per-user lock is held.
+
+The `wallet.withdrawal.auto_approved` audit row now carries `pivotAmount` and `pivotCurrency`. `threshold`, `dailyCapAmount` and `cumulativeAmountUsed` are in that pivot currency; `amount` and `currency` stay the raw request.
+
+**Behaviour change:** `fiatThreshold`, `cryptoThreshold`, per-player `auto_withdrawal_rule.threshold` and `autoWithdrawal.dailyCapAmount` are pivot amounts. A threshold configured in coin units reads as a tiny pivot amount after upgrade, so withdrawals route to manual review until an operator re-enters it - the safe direction. When no rate is available for the withdrawal's currency, or the fx module is not loaded, a withdrawal in any currency other than the pivot is not auto-approved; one already in the pivot needs no rate and is still evaluated.
+
+Auto-approvals written before the upgrade have no stored pivot value. While a daily amount cap is configured, a wallet with one in its trailing 24 hours routes to manual review until it ages out, and `cumulativeAmountUsed` is recorded as `null` rather than a partial total.
 
 The review-queue `large_amount` tag still compares the raw amount; it is a display hint, not a decision.
--- a/.changeset/currency-aware-auto-withdrawal.md
+++ b/.changeset/currency-aware-auto-withdrawal.md
@@ -1,0 +1,11 @@
+---
+'@openora/core': minor
+---
+
+Auto-withdrawal thresholds, the daily amount cap and the large-amount heuristic are now read in the fx pivot currency (`exchangeRate.pivot`, `USD` by default) instead of compared raw against whatever currency the withdrawal is in.
+
+One `cryptoThreshold` serves every crypto asset, so a raw comparison made its meaning depend on the coin: a threshold of `1` written with a stablecoin in mind let `0.4 BTC` auto-approve. The daily amount cap summed trailing payouts across currencies as if they were one unit. Both now convert through `EXCHANGE_RATE_READER`, and every comparison is exact decimal instead of `Number()`.
+
+**Behaviour change:** `fiatThreshold`, `cryptoThreshold`, per-player `auto_withdrawal_rule.threshold` and `autoWithdrawal.dailyCapAmount` are pivot amounts. A threshold configured in coin units reads as a tiny pivot amount after upgrade, so withdrawals route to manual review until an operator re-enters it - the safe direction. When no rate is available for the withdrawal's currency, or the fx module is not loaded, the withdrawal is not auto-approved.
+
+The review-queue `large_amount` tag still compares the raw amount; it is a display hint, not a decision.
--- a/packages/core/src/contracts/schemas/platform-config.ts
+++ b/packages/core/src/contracts/schemas/platform-config.ts
@@ -134,10 +134,12 @@ export const WalletConfigSchema = z
 
 export type WalletConfig = z.infer<typeof WalletConfigSchema>;
 
+const DEFAULT_EXCHANGE_RATE_PIVOT = 'USD';
+
 export const ExchangeRateConfigSchema = z
   .object({
     /** Comparison currency the fx module derives a cross rate against. Absent = 'USD'. */
-    pivot: CurrencyCodeSchema.default('USD'),
+    pivot: CurrencyCodeSchema.default(DEFAULT_EXCHANGE_RATE_PIVOT),
     freshTtlMs: z.number().int().positive().default(60_000),
     hardMaxAgeMs: z
       .number()
@@ -149,6 +151,11 @@ export const ExchangeRateConfigSchema = z
   .strict();
 
 export type ExchangeRateConfig = z.infer<typeof ExchangeRateConfigSchema>;
+
+/** The pivot every fx conversion targets. One resolver, so no module converts against another. */
+export function resolveExchangeRatePivot(config: ExchangeRateConfig | undefined): string {
+  return (config?.pivot ?? DEFAULT_EXCHANGE_RATE_PIVOT).toUpperCase();
+}
 
 export const NotificationsConfigSchema = z
   .object({
--- a/packages/core/src/wallet/__tests__/wallet-auto-withdrawal.int.test.ts
+++ b/packages/core/src/wallet/__tests__/wallet-auto-withdrawal.int.test.ts
@@ -210,6 +210,7 @@ describe('WalletService.withdraw auto-approval (real PG)', () => {
       reviewedBy: null,
       reviewReason: 'auto-approved',
       providerName: DEFAULT_PAYMENT_PROVIDER,
+      autoApprovalPivotAmount: '40.000000000000000000',
     });
     expect(events.emit.mock.calls.map(([topic]) => topic)).toEqual([
       'wallet.withdrawal.requested',
@@ -225,6 +226,8 @@ describe('WalletService.withdraw auto-approval (real PG)', () => {
           threshold: '1000.000000000000000000',
           thresholdSource: 'global',
           kycStatus: 'verified',
+          pivotAmount: '40',
+          pivotCurrency: 'USD',
           cumulativeCountUsed: 0,
         }),
       }),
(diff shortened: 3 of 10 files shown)


packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2405,18 +2425,34 @@ export class WalletService {
...
         ),
-      );
-    const amountUsed = row?.total ?? '0';
-    const countUsed = Number(row?.n ?? 0);
-    // Cap comparison is a review-queue decision, not a ledger write - moneyToNumber is the
-    // documented single conversion point (see the helper's own doc comment).
+      )
+      .groupBy(walletTransaction.currency);
+    const countUsed = rows.reduce((total, row) => total + Number(row.n), 0);
+    const [pivotAttempt, ...pivotUsed] = await Promise.all([
```
jakubfilinger-b: This puts an `EXCHANGE_RATE_READER` call inside the advisory-locked transaction, and the comment 80 lines up says the opposite is the design: "Threshold/KYC/risk/velocity gates run outside any lock; only the cap check below needs serializing." Before this PR `autoApprovalCaps` was one SQL statement on the passed `txn`; now it is `1 + N` calls into the fx reader while `withAdvisoryXactLock` is held.

Two concrete consequences, both from `ExchangeRateReaderService`:

1. A hard-stale or missing quote is fetched from the vendor **synchronously** (`resolveLeg` -> `fetchLeg`), bounded only by `providerTimeoutMs`. So a cold or expired BTC quote holds the per-user advisory lock and a pooled connection for the whole vendor round trip, on the withdrawal hot path.
2. `resolveLeg` reads `exchange_rate_quote` through the reader's own `this.drizzle`, i.e. a *second* connection from the same pool, taken while the outer transaction still holds the first. With pool size N, N concurrent auto-withdrawals each holding one connection and each waiting for another is a pool-exhaustion deadlock, not just slowness.

The fix looks cheap: the cap check only needs the *conversion* to be consistent, not to be inside the lock. Resolve the pivot values before `db.transaction(...)` — the trailing-payout currencies can be read outside the lock too, since the lock is there to serialize the read-and-flip of the cap marker, not the rate — and pass the converted numbers in. That also removes the vendor call from the transaction entirely.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2405,18 +2425,34 @@ export class WalletService {
...
+      this.toPivotAmount(amount, currency),
+      ...rows.map((row) => this.toPivotAmount(row.total, row.currency)),
+    ]);
+    const isUnpriced = pivotAttempt === null || pivotUsed.some((value) => value === null);
+    const amountUsed = pivotUsed.reduce<string>(
```
jakubfilinger-b: `amountUsed` is returned to the caller and lands in the `wallet.withdrawal.auto_approved` audit row as `cumulativeAmountUsed`, next to `amount` and `currency` which are still the raw withdrawal values. So the AML trail now carries two different units in one payload with nothing saying which is which. Worth either renaming the field to say it's a pivot amount, or carrying the pivot code alongside it, so someone reconstructing a decision a year from now doesn't read `cumulativeAmountUsed: 1500` as 1500 BTC.

Separately, `value ?? '0'` folds an unpriced currency into the sum as zero. When `cfg.dailyCapAmount` is undefined, `isUnpriced` never forces `amountExceeded`, so the withdrawal can auto-approve and record a `cumulativeAmountUsed` that silently omits every leg the reader couldn't price. Narrow, but it is exactly the row a reviewer would later trust. Either surface the unpriced legs in the decision or don't report a total you know is incomplete.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2405,18 +2425,34 @@ export class WalletService {
...
-      );
-    const amountUsed = row?.total ?? '0';
-    const countUsed = Number(row?.n ?? 0);
-    // Cap comparison is a review-queue decision, not a ledger write - moneyToNumber is the
-    // documented single conversion point (see the helper's own doc comment).
+      )
+      .groupBy(walletTransaction.currency);
+    const countUsed = rows.reduce((total, row) => total + Number(row.n), 0);
+    const [pivotAttempt, ...pivotUsed] = await Promise.all([
+      this.toPivotAmount(amount, currency),
```
jakubfilinger-b: `toPivotAmount(amount, currency)` was already computed in `evaluateAutoApproval` at :1930 for the threshold gate. Converting it a second time here means two reads of a value that can move between them — a quote that goes hard-stale and refreshes in the gap has the threshold gate and the cap gate judging the same withdrawal at two different rates. Pass the pivot amount down from `evaluateAutoApproval` instead; it also drops one call from inside the lock.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -561,11 +562,14 @@ function namespacedIdempotencyKey(namespace: string, rawKey: string): string {
...
-// ponytail: a flat, currency-agnostic threshold - a starter heuristic for the review
-// queue, not a compliance risk engine. Revisit with a real risk service if/when one exists.
-// moneyToNumber is the sanctioned JS conversion point for this heuristic comparison only.
+// ponytail: a flat threshold - a starter heuristic, not a compliance risk engine. Auto-approval
+// reads it in the fx pivot; the review-queue tag still compares the raw amount, so it is only
+// a display hint there. Revisit with a real risk service if/when one exists.
 const LARGE_WITHDRAWAL_THRESHOLD = '5000';
 
+// The fx module's own fallback when `exchangeRate.pivot` is unset.
+const DEFAULT_FX_PIVOT = 'USD';
```
jakubfilinger-b: The comment is honest about what this is — "the fx module's own fallback" — which is the problem: `DEFAULT_PIVOT` already exists in `fx/exchange-rate/plugin.ts` and this is a second copy that nothing keeps in step. If fx ever changes its default, wallet converts against one pivot while the reader resolves legs against another, and the mismatch is silent because `convert` just returns a number. Export the fx constant (or put it on the config schema as a real default) and import it here.

packages/core/src/wallet/__tests__/wallet-auto-withdrawal.int.test.ts
```diff
@@ -748,10 +768,70 @@ describe('WalletService.withdraw auto-approval - crypto rail (real PG)', () => {
...
+    const w = await seedWallet({ currency: 'BTC', balance: '10' });
+
+    const result = await svc.withdraw({
+      userId: w.userId,
+      amount: '2',
+      currency: 'BTC',
+      destinationAddress: 'bc1qexample',
+      ...NO_CLIENT_META,
+    });
+
```
jakubfilinger-b, line 829: This covers the reader being absent (`rates: 'unbound'`), but not the case the changeset actually warns about: the reader is bound and returns `null` for that currency, which is what a rate outage or an unlisted asset looks like in production. `fixedRates()` already returns `null` for anything outside `USD_PER_UNIT`, so a case withdrawing in, say, `ETH` would pin that path for the price of three lines. Same for the daily cap — `isUnpriced` forcing `amountExceeded` has no test at all.

jakubfilinger-b (changes requested):
The direction is right — one flat threshold across every crypto asset was never safe, and moving the comparisons off `Number()` onto `moneyCompare` is overdue. Two things I'd want resolved before this lands, both about *where* the conversion now happens rather than the pivot idea itself, plus a data question.

The blocking one is the rate lookup inside the advisory-locked transaction (detail on the thread at `autoApprovalCaps`). The rest are smaller.

On the data side: `wallet_auto_withdrawal_rule.threshold` rows written before this change are now silently reinterpreted as pivot amounts, and the changeset is the only place that says so. There's no migration, no flag on the row, and nothing an operator can query to tell a stale threshold from a deliberate one — a per-player rule of `5` (BTC, trusted player) becomes a $5 rule and just stops auto-approving, which is safe but invisible. Can we either stamp the rows (a `thresholdCurrency` column, or a `migratedAt` marker) or at least have the reconciliation cycle report rules whose pivot value looks implausible? Otherwise the only signal an operator gets is a queue that quietly fills up.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2405,18 +2425,34 @@ export class WalletService {
...
-    const amountUsed = row?.total ?? '0';
-    const countUsed = Number(row?.n ?? 0);
-    // Cap comparison is a review-queue decision, not a ledger write - moneyToNumber is the
-    // documented single conversion point (see the helper's own doc comment).
+      )
+      .groupBy(walletTransaction.currency);
+    const countUsed = rows.reduce((total, row) => total + Number(row.n), 0);
+    const [pivotAttempt, ...pivotUsed] = await Promise.all([
+      this.toPivotAmount(amount, currency),
+      ...rows.map((row) => this.toPivotAmount(row.total, row.currency)),
```
klaudia-blazyczek-blurify: Previous auto-approved withdrawals are converted with the current FX rate here. If BTC falls, past withdrawals count less and a player can pass the daily cap. Save the pivot amount or quote at approval and sum saved values.

packages/core/src/wallet/service/wallet.service.ts
```diff
@@ -2405,18 +2425,34 @@ export class WalletService {
...
-      moneyToNumber(amountUsed) + moneyToNumber(amount) > moneyToNumber(cfg.dailyCapAmount);
+      (isUnpriced ||
+        moneyCompare(moneyAdd(amountUsed, pivotAttempt ?? '0'), cfg.dailyCapAmount) > 0);
     const countExceeded = cfg.dailyCapCount !== undefined && countUsed + 1 > cfg.dailyCapCount;
     return { exceeded: amountExceeded || countExceeded, amountUsed, countUsed };
   }
 
+  private async toPivotAmount(amount: string, currency: string): Promise<string | null> {
+    const pivot = this.platformConfig?.exchangeRate?.pivot ?? DEFAULT_FX_PIVOT;
+    if (currency.toUpperCase() === pivot.toUpperCase()) {
```
klaudia-blazyczek-blurify: This returns a USD amount when `rates` is not bound, so USD withdrawals can still auto-approve. The PR says an absent FX module must fail closed; require `rates` first or narrow that statement and test the exception.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `7d0870a0` - the withdrawal is converted once in the pre-lock gates, and the cap under the lock is now a single sum over the stored `auto_approval_pivot_amount`, so no reader call and no second connection while the lock is held.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `7d0870a0` - the audit row now carries `pivotAmount` and `pivotCurrency` next to the raw `amount` and `currency`. With stored values the only unpriced leg left is an auto-approval from before the upgrade, and for those `cumulativeAmountUsed` is `null` instead of a partial total, while a configured amount cap counts as exceeded.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `7d0870a0` - `evaluateAutoApproval` returns `pivotAmount`, and the threshold, the heuristic and the cap all judge that one value.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Done in `7d0870a0` - moved to one `resolveExchangeRatePivot` in `schemas/platform-config.ts`, used by both the fx plugin and wallet, and the schema default reads the same constant.

packages/core/src/wallet/__tests__/wallet-auto-withdrawal.int.test.ts
zaxovaiko: Done in `7d0870a0` - added the bound-reader case with `ETH` returning `null`, a trailing auto-approval with no stored pivot value against the amount cap, and one proving the cap sums the value stored at approval.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Good catch, done in `7d0870a0` - each auto-approved payout stores its pivot value in `wallet_transaction.auto_approval_pivot_amount` and the cap sums those. There's a test where the stored value is above today's rate and the next withdrawal stays pending.

packages/core/src/wallet/service/wallet.service.ts
zaxovaiko: Hm, I'd keep the behaviour - a withdrawal already in the pivot needs no rate, and requiring the fx module would take auto-approval away from operators who only run the pivot currency. So I narrowed the changeset instead and added a test for that exception, in `7d0870a0`.

zaxovaiko:
@jakubfilinger-b on the stale thresholds - agreed it's invisible today. It fails safe and the changeset calls it out, but a `thresholdCurrency` column or a reconciliation report changes the rule model, so I'd take it as a separate follow-up rather than grow this PR.

.changeset/currency-aware-auto-withdrawal.md
zaxovaiko: Remove it

jakubfilinger-b (approved):
Rechecked the follow-up. The blocking lock/FX issue and the inline audit, consistency, and coverage points are addressed. Treating legacy threshold traceability as a separate follow-up.

packages/core/src/contracts/schemas/platform-config.ts
```diff
@@ -134,10 +134,12 @@ export const WalletConfigSchema = z
 
 export type WalletConfig = z.infer<typeof WalletConfigSchema>;
 
+const DEFAULT_EXCHANGE_RATE_PIVOT = 'USD';
```
zaxovaiko: I think this must be configurable from database level or UI

PR #164, opened 2026-09-10
chore(tooling): enforce the standards this repo already writes down

zaxovaiko:
## Summary

Two pieces of agent tooling that turned out to be the same problem: rules this repo already writes down, with nothing mechanically enforcing them.

Trims the vendored `anti-slop` oxlint plugin to the ten rules that enforce a standard already written here, and adds the gates the agent layer never had: a prose-reference lint in `pnpm verify`, an approval gate that survives bypass mode, a `cleaner` subagent that applies rather than reports, documentation obligations wired into delivery, and a dispatch-only docs audit that n8n schedules.

## Why

`pnpm verify` checked generated artifacts and code shape. Nothing read prose. Every `SKILL.md` is a one-line pointer into `docs/standards/skills/`, so a renamed doc, anchor, roster agent or `pnpm` script silently turned a skill into a no-op, and no gate would notice.

Measured across 24 headless runs of all six skills on Sonnet 5 and Opus 5, in a sandbox with a local git remote and a stubbed `gh` (nothing reached GitHub or Jira):

- Opus never read the standards doc its skill points at. Six runs, zero reads. Sonnet read it in five of six. An instruction only one model follows is not a standard.
- `review` fanned out to all three roster reviewers on Sonnet and to none on Opus, which also emitted no verdict.
- Two Sonnet runs tried to commit a `Co-Authored-By: Claude` trailer, which `delivery.md` forbids. Only a contributor's personal global hook stopped it.

`check:refs` found seven real defects, all fixed here: two ADRs both numbered 0032, ADR-0024 still describing `packages/foundation/*` that ADR-0025 deleted, `tools/build-contract.ts` cited in two agent briefs after its removal, a wrong react hooks path, a dead anchor, and `/pre-pr` claiming `verify` skips a check it actually ends with.

On the lint rules: upstream states its rules are one engineer's preferences and are meant to be vendored, not depended on. Taking all eighteen treated that as a package deal - 4,380 lines, of which 358 could never run and 1,094 existed to enforce blank lines. A probe of five slop patterns that `types.md` already forbids was caught 8 times by the kept rules, 0 times by this repo's own config, and passed `tsc --strict` clean.

## Alternatives considered

A `/plan` skill instead of the hook. Rejected: it would restate what built-in plan mode already does and be exactly as unenforceable as the prose it replaces. A `PreToolUse` denial holds in every permission mode, including `--dangerously-skip-permissions`.

A GitHub Actions `schedule:` for the docs audit. Rejected in favour of n8n, which already schedules the other openora automations; two schedulers for one job is one too many. `docs-audit.yml` is `workflow_dispatch` only.

Rewriting the ten kept lint rules by hand, per ADR-0015's preference. Rejected: `no-widen-then-assert` is 366 lines because the AST work is genuinely fiddly, and reimplementing it produces the same behaviour with fresh bugs.

Checking ADR supersession symmetry in `check:refs`. Dropped: this repo supersedes partially, so a superseded ADR legitimately keeps `Status: Accepted`. No mechanical rule separates that from rot.

## Risks

Two concerns in one PR, against `delivery.md`'s "one PR per concern". Merged rather than rebased so the two lines of work stay separable in the history.

The plan gate is off unless `OPENORA_PLAN_GATE` is set, so no existing workflow changes. The `cleaner` subagent edits, unlike the three reviewers - it is scoped to the branch diff, never changes behaviour, and ends on `pnpm verify`.

`docs-audit.yml` needs a `CLAUDE_CODE_OAUTH_TOKEN` repository secret before it can run at all; the n8n workflow that dispatches it is created inactive and has never executed. The job is read-only by construction - no Edit or Write in its `--allowedTools` - so a scheduled model cannot rewrite prose in a regulated codebase.

Deferred: 692 anti-slop warnings across the eight rules still ramping at `warn`. That is a triage-per-rule job and its own PR. One of them, in `audit.service.ts`, parses half an RG limit-change payload through a schema and half through `typeof` ternaries that turn a malformed value into `null` in an audit row - worth pulling out first.

Also deferred: an integration-test flake under parallel load. The mail wait is fixed here and local failures drop from 61 to 7, but two causes remain unexplained and predate this branch.


.oxlintrc.json
```diff
@@ -27,7 +28,26 @@
...
+    "anti-slop/no-reflect-apply": "error",
+    "anti-slop/no-reflect-get": "error",
+    "anti-slop/no-runtime-typeof": "warn",
+    "anti-slop/no-shape-in-symbol-names": "warn",
+    "anti-slop/no-unknown-parameters": "warn",
+    "anti-slop/no-unknown-returns": "warn",
+    "anti-slop/no-unknown-type-aliases": "error",
+    "anti-slop/no-unsafe-dictionary-type": "warn",
+    "anti-slop/no-widen-then-assert": "error",
+    "anti-slop/require-readable-spacing": "warn",
```
jakubfilinger-b, line 49: This rule alone is 6.3k of the 7k warnings, and the plan is to autofix it in a separate PR "at a quiet moment". That leaves an interim state where `pnpm check:lint` prints seven thousand lines, and I don't think anyone reads the seven-thousand-and-first — including the warnings from the rules that were already there and were previously actionable.

It also sits badly with the policy this PR writes into `enforcement.md`: "a `warn` is known debt, not a suggestion to ignore". 6.3k sites of a purely stylistic rule will be ignored, by construction, and the sentence loses force for the rules where it matters.

The sequencing that gets the same end state without the blind window is to register this one as `off` here, land the autofix PR, then flip it straight to `warn` (or `error`, since after an autofix the codebase is clean of it, which is your own bar for `error`). Nothing else in the PR has to change, and the conflict-with-open-branches argument for deferring the fix is unaffected.

Separate question while you're here: blank lines between statements are normally the formatter's business, and this repo already runs `oxfmt`. Is there a reason this is a lint rule rather than an `.oxfmtrc.json` setting? If both end up with an opinion about the same whitespace, `fix:lint` and `fix:format` can disagree.

.oxlintrc.json
```diff
@@ -27,7 +28,26 @@
...
-    "oss-module-shape/no-inline-z-enum-outside-contract": "error"
+    "oss-module-shape/no-inline-z-enum-outside-contract": "error",
+    "oxc/no-accumulating-spread": "error",
+    "anti-slop/no-array-filter-map": "warn",
+    "anti-slop/no-reduce-accumulator-copy": "error",
+    "anti-slop/no-chained-type-assertions": "warn",
+    "anti-slop/no-conditional-empty-object-spread": "warn",
+    "anti-slop/no-known-value-widening": "warn",
+    "anti-slop/no-module-mocking": "warn",
+    "anti-slop/no-object-parameters": "warn",
```
jakubfilinger-b, line 39: This is the one I'd want a second opinion on before the follow-ups start, and it applies to `no-unknown-parameters` at :44 too. The Risks section says clearing these will change "parts of the public API (`EventHandler`, `AdminGuard.assert`)".

Those are published surfaces that consumers implement against, so a change there is a breaking release for them regardless of how good the reason is. I'm not against the rules — I'm against the ordering of the argument. "The linter flags it" should not be what carries a public signature change; the change should stand on its own (this is the better shape for consumers, here is what breaks, here is the migration), and the rule flip should follow it rather than motivate it.

Concretely: when those follow-ups land, could they be framed as API PRs with their own changesets and a stated migration, with the `warn` -> `error` flip as a one-line afterthought? If a rule turns out not to be worth a breaking change, the honest outcome is that the rule stays `warn` on the public surface, not that the surface moves.

jakubfilinger-b (commented):
I'm in favour of the premise. "Review catches some of it, but not consistently" is the honest version of where we are, and type-evidence rules are the right subset to mechanise — `as` with no stated reason and `unknown` parameters are exactly the patterns that read as fine in a diff and cost you later.

The execution is careful in the ways that matter: the licence is vendored alongside the code, the upstream commit and every local change are recorded, the nested `eslint-stylistic` vendor carries its own LICENSE and UPSTREAM.md, the rule tests actually run in `test:tools`, and the published consumer config is untouched. I checked that last one — `packages/core/oxlint/oxlintrc.json` is not in the diff, so a consumer sees nothing. And since `check:lint` is a bare `oxlint .` with no `--deny-warnings`, none of the 7k warnings can fail anyone's build. Good.

Not blocking anything. Three thoughts, one of which I'd act on before merge (the `require-readable-spacing` sequencing) and one of which is bigger than this PR (the public-API follow-ups).

.oxlintrc.json
```diff
@@ -37,7 +57,14 @@
     "coverage",
     "generated",
     "*.generated.ts",
-    "tools/templates"
+    "tools/templates",
+    "tools/lint/anti-slop/**",
+    ".agents/**",
+    ".claude/**",
```
jakubfilinger-b, line 63: These ignore entries aren't mentioned anywhere in the description, which took me a detour to work out. They're the generated agent mirrors — nothing under them is committed on `dev`, so this is housekeeping to stop oxlint walking generated files in a working tree, not a suppression. Worth a clause in the body saying so.

One forward-looking note: `.github/hooks/**` is on the list. Nothing lives there today, but hooks are executable scripts, and if one ever lands it will be unlinted with no visible reason why. If the intent is "generated agent instructions", it might be safer to scope these to the generated file patterns rather than to whole directories that could later hold real code.

docs/standards/enforcement.md
```diff
@@ -8,6 +8,7 @@ Detail for the enforcement line in `conventions`. Read this when a gate fails or
...
 - oxlint config is split: the published `@openora/core/oxlint/oxlintrc.json` holds the universal, stack-agnostic rules (base rules, `typescript/no-explicit-any`, `typescript/no-non-null-assertion`, `typescript/consistent-type-definitions`, `import/no-cycle`, `import/no-duplicates`) - the single source of truth a consumer extends via `"extends": ["./node_modules/@openora/core/oxlint/oxlintrc.json"]`. The root `.oxlintrc.json` here `extends` that shared config and adds only OSS-internal rules (`oss-boundaries/*`, `oss-module-shape/*`, `unicorn/filename-case`) that need the local `jsPlugins`.
+- Low-evidence TypeScript patterns are lint-enforced by the vendored `anti-slop` oxlint JS plugin (`tools/lint/anti-slop/`, upstream commit and local changes in its README). Every generic rule is registered in the root `.oxlintrc.json`. A rule is `error` once the codebase is clean of it and `warn` until the pull request that fixes its remaining sites flips it, so a `warn` is known debt, not a suggestion to ignore. Its rule tests run in `pnpm test:tools`.
```
jakubfilinger-b: "A rule is `error` once the codebase is clean of it and `warn` until the pull request that fixes its remaining sites flips it" is a good, checkable policy — worth having written down.

What it doesn't say is what stops the debt growing while a rule sits at `warn`. With ~7k warnings in the output, a newly introduced violation of a `warn` rule is invisible: nobody diffs the warning list, and nothing fails. So the twelve rules parked at `warn` can quietly get worse between now and the PR that is supposed to clear them, which makes each of those PRs bigger than it needs to be.

A per-rule warning baseline that CI compares against (fail if the count for rule X went up) would make the policy self-enforcing. Even without tooling, recording today's count per rule in this document — you already have the five biggest — gives the next person a number to check against rather than a vibe.

jakubfilinger-b (dismissed):
Rechecked the updated branch. The remaining review notes are non-blocking follow-ups.

zaxovaiko:
Some fixes will be done in another PR.

PR #165, opened 2026-09-11
feat: lifecycle transactional emails and the email-change flow

klaudia-blazyczek-blurify:
## Summary

Adds the mail template keys and triggers the approved mockups core didn't have: a welcome mail, a two-step OTP email-change flow (request + confirm) with paired old-address/new-address notices, and dedicated withdrawal-completed/withdrawal-failed mails. Every responsible-gambling mail (cooling-off, self-exclusion, limit change) now carries who triggered it - player, admin, or the system sweep. Mail greetings use the player's username instead of their free-text display name.

## Why

Mapping the approved mockups to code surfaced several gaps, some more than a text mismatch:

- No email-change flow existed - `IdentityService.changeEmail` swapped the address synchronously with no confirmation and no notification. Built the OTP-confirmed two-step flow instead: `requestEmailChange` mails a code to the new address (behind a fresh password/2FA check), `confirmEmailChange` verifies it and swaps the login email.
- `wallet.withdrawal.failed` only fired when an admin reviewer was attached. An auto-approved payout that failed, or a payment-provider webhook rejecting one later, silently returned the funds with no notification at all.
- `wallet.withdrawal.completed` already had a template key but no `email` option wired into the notification map, so no withdrawal-success mail was ever sent.
- RG mails gave no indication of who acted - a player reading "your cooling-off has ended" couldn't tell whether they lifted it themselves, an admin did, or it just expired. Every `rgCoolingOffActivated`/`rgCoolingOffLifted`/`rgLimitUpdated`/`rgSelfExclusionActivated`/`rgSelfExclusionLifted` mail now carries `initiatedBy` (`player` | `admin` | `system`), including a new mail sent by the expiry sweep itself when a cooling-off lapses naturally (that path never went through `liftCoolingOff`, so nobody was telling the player their break had ended).
- Mail greetings read `user.name`, a free-typed display field a player can set to anything. `user.username` is the required, unique, always-present identifier and is what greetings resolve from now, across every template that addresses a recipient by `userId`.
- A review pass on this branch surfaced real bugs in the code above it, fixed here since they touch the same paths: the `emailChanged` notice was missing the recipient's anti-phishing code; the email-change rate limit was keyed on the target address alone, so any signed-in caller could exhaust a stranger's budget and block their real change; the withdrawal-failed audit row was written by a best-effort event instead of the same transaction as the refund credit; and the new-address confirmation notice read the account row live at delivery time instead of snapshotting it, which could relabel the notice onto a later address on a fast second change.

## Alternatives considered

- A confirmation link instead of an OTP for the email change - not chosen, the requirement calls for a code.
- A separate "generic transactional" template key - not chosen, the base layout already covers untemplated copy.
- Threading a real actor into `liftCoolingOff`/`liftSelfExclusion`'s `initiatedBy` instead of hardcoding `'admin'` - not chosen, both are today admin-only back-office actions with no self-service lift route, so there is no other value to pass; revisit if a self-service lift ships.

## Risks

- Breaking change for any `@openora/core` consumer pinned to a canary: the old single `changeEmail` route/schema is replaced by `requestEmailChange`/`confirmEmailChange`, and `wallet.withdrawal.failed`'s `adminId` is now nullable (event version 2 -> 3). A consumer reading `adminId` without a null check breaks at runtime; there is no deprecation shim, so downstream needs a coordinated canary bump.
- Breaking change to the mail contract: `EmailTemplateData['emailChanged']` gained required `occurredAt`/`isNewAddress`, `emailChangeConfirmation` gained required `oldEmail`, and `rgCoolingOffActivated`/`rgCoolingOffLifted`/`rgLimitUpdated`/`rgSelfExclusionActivated`/`rgSelfExclusionLifted` all gained a required `initiatedBy`. A custom `EmailTemplateRenderer` implementation needs to handle every one of these.
- The email-change flow now requires the caller's current password (plus a TOTP code if 2FA is enrolled) and revokes every other session once confirmed - a live session alone must never be enough to move an account's recovery address.
- The mail-greeting source change is repo-wide, not scoped to the templates touched here: any operator whose rendered copy showed `user.name` sees `user.username` instead from this canary on, for every mail resolved by `userId`.


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -2108,46 +2111,155 @@ export class IdentityService {
...
           ip,
           userAgent,
         });
       }
+      // To the OLD address - the row now holds the new one. Fire-and-forget like
+      // `dispatchMail` in notifications/plugin.ts: the swap already committed, so a
+      // mail-enqueue hiccup must not turn it into a reported failure. `randomUUID()`
+      // keeps the key unique so the queue's dedupe window can't swallow a repeat notice.
+      this.mailDispatch
+        ?.toAddress({
```
zaxovaiko: `emailChanged` goes out via `toAddress`, and `mail/mail.service.ts` always resolves an address recipient with `antiPhishingCode: null`, so this one lands without the code while `withdrawalCompleted`/`withdrawalFailed` carry it. Of all our mails this is the one people will actually forge - let's pull `antiPhishingCode` into the `before` select and pass it through.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -2108,46 +2111,155 @@ export class IdentityService {
...
     const userId = await this.currentUserId(headers);
+    await assertRateLimit(
+      this.limiter,
+      `change-email:${userId ?? 'anonymous'}`,
+      VERIFY_EMAIL_RATE_LIMIT,
+    );
     const newEmail = input.newEmail.toLowerCase();
-    const res = await this.api.changeEmail({
+    // Separate budget caps codes per target inbox, regardless of caller.
+    await assertRateLimit(this.limiter, `change-email-target:${newEmail}`, VERIFY_EMAIL_RATE_LIMIT);
```
zaxovaiko: The `change-email-target` and `change-email-ip` buckets are spent before the `!userId` throw, so an anonymous caller can burn them - 5 requests is exactly the target limit, enough to keep an address unclaimable for its real owner. Let's move the `!userId` check above both. Same in `confirmEmailChange`, which doesn't check `userId` at all. Also `changePassword` only has the caller-keyed bucket, so the comment about matching its convention isn't quite right.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -2108,46 +2111,155 @@ export class IdentityService {
...
+      const playerId = await this.identityReader.getPlayerIdByUserIdSafe(userId);
+      this.events.emit('identity.email.changed', {
+        userId,
+        playerId,
+        previousEmail: before.email,
+        newEmail,
+        ip,
+        userAgent,
+      });
+      await this.trustedDevices?.revokeAllForUser(userId, userId);
```
zaxovaiko, line 2613: `revokeAllForUser`/`revokeAllSessions` are awaited before the notice is enqueued - if either throws, the swap is already committed, the sessions stay alive and the user never gets warned. Let's enqueue the notice right after the `identity.email.changed` emit, it's fire-and-forget anyway.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -2108,46 +2111,155 @@ export class IdentityService {
...
     const headers = nodeHeadersToHeaders(reqHeaders);
+    const meta = extractClientMeta(reqHeaders);
+    const { ip } = meta;
+    // Session resolved first so the rate limit keys on the caller, not the target
+    // address - same as `changePassword`'s `${userId ?? 'anonymous'}` key.
     const userId = await this.currentUserId(headers);
+    await assertRateLimit(
+      this.limiter,
+      `change-email:${userId ?? 'anonymous'}`,
+      VERIFY_EMAIL_RATE_LIMIT,
```
zaxovaiko: This is a send route but uses `VERIFY_EMAIL_RATE_LIMIT` - shouldn't it be `EMAIL_VERIFICATION_RATE_LIMIT`? Each request mints a fresh code with a fresh 3-attempt counter, which is exactly the case that constant's own comment describes. `confirmEmailChange` is already on the right one.

packages/core/src/engagement/notifications/plugin.ts
```diff
@@ -393,6 +421,25 @@ export default {
       });
     }
 
+    // Welcome mail, mail-only like the security login alert below. Keyed on the user, not
```
zaxovaiko: Small one - the comment says keying on the user means a second welcome can never go out, but `idempotencyKey` is a BullMQ `jobId` and those get evicted after 24h/10k. What actually makes it once-per-account is `sendEmailVerification` skipping verified addresses. Maybe worth saying that instead.

zaxovaiko (commented):
Traced both new email routes end to end and the `wallet.withdrawal.failed` fan-out - the security model holds, the OTP is bound to the `current-new` address pair so a stolen session alone can't spend one, and the two dead withdrawal triggers are genuinely fixed. Left a few comments, the anti-phishing one on `emailChanged` is the one I'd really like fixed before merge.

One thing for the ticket rather than the code: the "generic transactional" variant isn't there, and the reasoning only lives in the PR description. Worth recording it there so the AC can be closed honestly.

packages/core/src/pam/identity/service/identity.service.ts
klaudia-blazyczek-blurify: Good catch - it was silently defaulting to null since `toAddress` had no path to carry a code at all. Threaded `antiPhishingCode` through `MailToAddressInput` -> `MailRecipientSchema`'s address kind -> `resolveRecipient`, pulled it into the `before` select in `confirmEmailChange`, and pass it on the notice. 2f5df3df.

packages/core/src/pam/identity/service/identity.service.ts
klaudia-blazyczek-blurify: Right on both counts. Moved the `!userId` throw above the target/IP buckets in both requestEmailChange and confirmEmailChange, and fixed the misleading comment (changePassword only has the one pre-authenticated bucket, no target/IP split). Added a rate-limit regression test proving an anonymous caller can no longer touch either bucket. 2f5df3df.

packages/core/src/pam/identity/service/identity.service.ts
klaudia-blazyczek-blurify: Moved the mailDispatch.toAddress enqueue to right after the identity.email.changed emit, before revokeAllForUser/revokeAllSessions - so the notice is already in flight if either revoke call throws. Added a test that spies revokeAllSessions to throw and asserts the notice still went out. 2f5df3df.

packages/core/src/pam/identity/service/identity.service.ts
klaudia-blazyczek-blurify: Yep, mixed up the two constants - swapped requestEmailChange to EMAIL_VERIFICATION_RATE_LIMIT (fresh-code-per-send semantics) on all three of its buckets, left confirmEmailChange on VERIFY_EMAIL_RATE_LIMIT since that one's guessing the code, not sending it. 2f5df3df.

packages/core/src/engagement/notifications/plugin.ts
klaudia-blazyczek-blurify: Agreed, rewrote it: idempotencyKey is a BullMQ jobId (short-window dedupe), the real once-ever guarantee is sendEmailVerification skipping already-verified addresses, so identity.email.verified can't fire twice. 2f5df3df.

.changeset/email-change-otp-and-withdrawal-failed-notice.md
zaxovaiko, line 1: We can remove it.

packages/core/src/server/auth/auth.ts
```diff
@@ -157,10 +170,41 @@ export function createAuth(options: AuthOptions): BetterAuthType {
...
+            await dispatchOtpMail({
+              to: email,
+              template: {
+                key: 'emailChangeConfirmation',
+                data: { otp, oldEmail: maskEmail(session.user.email), newEmail: email },
+              },
+              recipientName: (session.user as { username?: string | null }).username ?? null,
+              // The new inbox has no user row of its own yet, so it cannot resolve one
+              // itself - carry the caller's own code, unaffected by the pending swap.
+              antiPhishingCode:
```
zaxovaiko: Hm, I think `antiPhishingCode` here is always `null` - better-auth only returns fields it knows about on the session user, and `antiPhishingCode` isn't in `additionalFields`. The test mocks `getSession`, so it doesn't catch it. To be honest, I wouldn't make it work either: the new inbox hasn't proven ownership yet, so a typo in `newEmail` would hand a stranger the code plus the username, and then they could send the real player a very convincing phishing mail. Let's drop `antiPhishingCode` and `recipientName` from this mail, along with the comment and the test assertion.

.changeset/email-change-otp-and-withdrawal-failed-notice.md
klaudia-blazyczek-blurify: Makes sense given the new no-changeset-in-feature-PR policy - removed the file.

packages/core/src/server/auth/auth.ts
klaudia-blazyczek-blurify: Good catch - antiPhishingCode isn't in additionalFields so it was always null in the real session, just not in the test's own mock. Agreed on the security point too. Dropped both fields plus the comment and the test assertion.

zaxovaiko (approved):

PR #166, opened 2026-09-14
perf(tooling): cut the per-edit and per-commit hook latency

zaxovaiko:
## Summary

The `PostToolUse` edit hook drops from 4.4s to 0.33s per TypeScript edit, and the pre-commit hook from 95s (17 minutes on a cold turbo cache) to 0.81s. The git hooks lose their `pnpm` wrappers, and the consumer template gets the same edit-hook fix.

## Why

Measured across a day of agent sessions: 757 minutes of wall-clock went to tool calls against 458 minutes of model generation, and the hooks were a large share of it. Almost none of that time was the checks doing work.

- `pnpm exec oxfmt` takes 2.25s where `node_modules/.bin/oxfmt` takes 0.12s. The edit hook spawned two such processes; husky already exports `node_modules/.bin` onto PATH, so the git hooks never needed the wrapper either.
- The pre-commit hook ran the full `pnpm verify`. Nothing it caught was not already caught by the pre-push commitlint hook, the `/pre-pr` gate, or CI, and it charged minutes per commit for the privilege.
- The edit hook's typecheck stage was dead code: it gated on a `typecheck` package script, and every package here names that script `check:types`, so it had never once run. Wiring it up would have added 16.4s to every edit, so it is gone and typechecking stays in `pnpm verify`, `/pre-pr`, and CI.

Separately, a defect in the new hook that the first commit introduced and the second fixes: `grep` exits 1 when it matches nothing, husky runs hooks under `sh -e`, and the staged-file filter is a command substitution, so a commit touching no lintable file aborted with no diagnostic.

## Alternatives considered

Keeping a typecheck on the edit hook by correcting the script name. Rejected on cost: 16.4s on every edit to duplicate a gate that already runs before the PR.

Measured and left alone: `post-checkout` / `post-merge` regen (6s, now 0.91s), the `guard-*` PreToolUse scripts (0.07s each), and context size, which turned out not to drive turn latency (2.3s at 0k tokens against 4.5s at 100k, flat above that).

## Risks

Formatting and linting at commit time now see the staged files rather than the whole repository, so a pre-existing violation elsewhere no longer blocks an unrelated commit. `pnpm verify` and CI are unchanged and still cover the full tree.

`pnpm verify` passes. Verified by hand: an `any` in an edited file exits 2 with the capped oxlint report, an `any` in a staged file fails the commit, formatter fixes land in the staged blob, a docs-only commit passes, and a malformed commit message still fails `commit-msg`.


.husky/pre-commit
```diff
@@ -1,10 +1,25 @@
...
-if [ -n "$staged" ]; then
+# `|| true`: grep exits 1 when the commit touches no lintable file, and husky runs
+# this hook under `sh -e`, so an unguarded no-match would abort the commit.
+lintable=$(printf '%s\n' "$staged" | grep -E '\.(ts|tsx|js|jsx|mjs|cjs)$' || true)
+if [ -n "$lintable" ]; then
+  printf '%s\n' "$lintable" | tr '\n' '\0' | xargs -0 "$BIN/oxfmt" || exit 1
+  printf '%s\n' "$lintable" | tr '\n' '\0' | xargs -0 "$BIN/oxlint" --fix || exit 1
+  node tools/lint/ensure-final-newline.mjs || exit 1
+  # Re-add ONLY the staged set: the fixers rewrote files, and a blanket `git add`
+  # would drag unrelated unstaged work into this commit.
```
klaudia-blazyczek-blurify: `git add` on the next line stages the whole working-tree version of each path. If a file has staged and unstaged hunks, the unstaged hunk is committed after formatting. Preserve unstaged hunks, or refuse to fix partially staged files. This conflicts with `docs/standards/git-delivery.md`.

.husky/pre-commit
```diff
@@ -1,10 +1,25 @@
...
+"$BIN/tsx" tools/lint/verify-oss-hygiene.ts || exit 1
 
-if [ -n "$staged" ]; then
+# `|| true`: grep exits 1 when the commit touches no lintable file, and husky runs
+# this hook under `sh -e`, so an unguarded no-match would abort the commit.
+lintable=$(printf '%s\n' "$staged" | grep -E '\.(ts|tsx|js|jsx|mjs|cjs)$' || true)
+if [ -n "$lintable" ]; then
+  printf '%s\n' "$lintable" | tr '\n' '\0' | xargs -0 "$BIN/oxfmt" || exit 1
+  printf '%s\n' "$lintable" | tr '\n' '\0' | xargs -0 "$BIN/oxlint" --fix || exit 1
+  node tools/lint/ensure-final-newline.mjs || exit 1
```
klaudia-blazyczek-blurify: This script writes a final newline to every tracked text file, not only `$staged`. A commit with one TypeScript file can therefore modify unrelated unstaged files. Limit it to staged paths. This conflicts with `docs/standards/git-delivery.md`.

.husky/pre-commit
```diff
@@ -1,10 +1,25 @@
...
+#
+# Binaries come straight from node_modules/.bin: `pnpm exec` adds ~2.2s of
+# package-manager startup per call, the binary itself runs in ~0.12s.
+BIN="$(git rev-parse --show-toplevel)/node_modules/.bin"
+
 staged=$(git diff --cached --name-only --diff-filter=ACMR)
+[ -z "$staged" ] && exit 0
 
-pnpm verify
+"$BIN/tsx" tools/lint/verify-oss-hygiene.ts || exit 1
```
klaudia-blazyczek-blurify, line 27: This removes the type and boundary checks from pre-commit, but `docs/standards/enforcement.md` still says that pre-commit runs both. Update the documented gate or keep the checks.

klaudia-blazyczek-blurify (changes requested):

.husky/pre-commit
zaxovaiko: Correct, and this one was mine - the script walks `git ls-files` and ignored its arguments, so a commit containing one TypeScript file could rewrite any tracked file in the repo that was missing a trailing newline. It now takes path arguments and the hook passes the staged set; with no arguments it still scans everything, which is what `pnpm fix:format` needs. Verified: an unstaged file with no final newline is left untouched by a commit that does not include it. Fixed in 425bffeb.

.husky/pre-commit
zaxovaiko: Correct. The hook now refuses a commit where one file holds both staged and unstaged hunks, and names the files. I went with refusing over preserving the unstaged hunks: doing that properly means a stash dance around the fixers, and silently committing work the author deliberately left out is the worse failure to risk. The message points at `--no-verify` plus a manual `pnpm verify` for anyone who wants the partial commit anyway. Verified the working tree survives the refusal intact. Fixed in 425bffeb.

.husky/pre-commit
zaxovaiko: Correct - and it was already stale before this branch, since the hook ran `pnpm verify` rather than those two checks. ADR-0015 carried the same claim, so both are updated: `enforcement.md` now describes the staged-file gate and points at `/pre-pr` for the full local run, and I added a line documenting the partial-stage refusal. Fixed in 425bffeb.

.husky/pre-commit
```diff
@@ -1,10 +1,43 @@
...
+
 staged=$(git diff --cached --name-only --diff-filter=ACMR)
+[ -z "$staged" ] && exit 0
+
+# The fixers below rewrite whole files in the working tree and the re-add at the end
+# stages whole files, so a file carrying both staged and unstaged hunks cannot go
+# through this hook without its unstaged hunk landing in the commit. Refuse instead:
+# silently committing work the author deliberately left out is the worse failure.
+unstaged=$(git diff --name-only)
+if [ -n "$unstaged" ]; then
```
klaudia-blazyczek-blurify, line 18: `$unstaged` can contain several paths. Passed as one `grep -x` pattern, it only matches a single line containing all paths, so a partially staged file is missed whenever another file also has unstaged changes. Compare the path lists one file at a time before running the fixers. This breaks the partial-stage protection documented in `docs/standards/enforcement.md`.

klaudia-blazyczek-blurify (changes requested):

.husky/pre-commit
zaxovaiko: I tested this rather than reasoning from the shell docs, and the protection holds - a multi-line string is not one pattern here.

POSIX defines the `grep` pattern operand as a newline-separated *list* of patterns, and both implementations agree:

```
unstaged=$'a.ts\nb.ts\nc.ts'
printf 'b.ts\nz.ts\n' | grep -Fx "$unstaged"
# BSD grep (macOS):   b.ts
# GNU grep (debian):  b.ts
```

Three patterns matched independently. If they were concatenated into a single pattern, that would print nothing.

I also ran the exact case you describe - one file partially staged while a second file has unstaged changes:

```
staged:   tmp-probe-a.ts
unstaged: tmp-probe-a.ts tmp-probe-b.ts

pre-commit: staged and unstaged changes in the same file:
packages/core/src/tmp-probe-a.ts
exit=1
```

Caught and named correctly. `-x` is also what keeps it safe when one path is a substring of another.

Fair hit on the testing, though: my original check only covered a single unstaged file, so the multi-file path was genuinely unverified until you raised it.

The one edge neither of us covered is a filename containing a literal newline. Git quotes those identically on both sides of the comparison by default (`core.quotePath`), so they still match; it would only break with that setting disabled, which I do not think is worth guarding.

Leaving the code as is - happy to reopen if you can get a partial stage past it.

klaudia-blazyczek-blurify (approved):

PR #167, opened 2026-09-14
feat(casino): game tag catalog management

okapitula:
## Summary

Adds game tags to the casino catalog: backoffice CRUD for tags, tag assignment on game edits, and visible tags on public game and lobby reads.

## Why

Operators need to label and badge games (e.g. "New", "Hot") independently of categories. Categories group games for navigation. Tags are lightweight labels with their own visibility and badge styling, so an operator can prepare a label before showing it to players.

- New `game_tag` and `game_tag_game` tables. A tag has a unique name, a `visible`/`invisible` visibility, and badge colours.
- Admin tag routes (list, get, create, update, delete) are guarded by `game-config`, and `game-config` gains the `delete` action.
- `updateGame` accepts `tagIds`. The tag check takes `FOR KEY SHARE`, so a concurrent tag delete returns not-found instead of a foreign-key error.
- Public `listGames`, `getGame` and the lobby return only visible tags. Admin reads return every tag.
- `gaming.tag.created`, `updated` and `deleted` events are audited against the `game_tag` resource. `gaming.game.updated` snapshots now carry `tagIds`.
- Partial `badgeSettings` updates merge into the stored colours.

## Alternatives considered

- **Reuse categories with a flag:** rejected. Categories drive navigation and translations, while tags need per-tag visibility and badge styling, and a flag would put both jobs on one entity.
- **Let the admin API set a tag's type:** rejected. System tags can't be deleted, so anyone with game-config create could make a tag nobody can remove. API-created tags are always custom, and system tags come from a direct database insert.

## Risks

- **Schema shapes:** `GameSchema` and the lobby `GameSummarySchema` gain a required `tags` array. `gaming.game.updated` snapshots gain `tagIds`, which defaults to `[]` when older events are replayed.
- **Permissions:** `game-config` gains `delete`, so every role with game-config `read_write` can delete custom tags. That covers the seeded casino-manager and sportsbook-manager roles.
- **Migration:** `0006` only adds enums, tables and indexes, so there's no rollout ordering concern.
- **System tags:** nothing seeds them yet. They exist only when inserted directly into the database.


marek-chmielowski-blurify (commented):
Automated code review notes (efficiency/duplication) — non-blocking.

packages/core/src/casino/lobby/service/lobby.service.ts
```diff
@@ -169,7 +178,10 @@ export class LobbyService {
             .innerJoin(gameProvider, eq(game.providerId, gameProvider.id))
             .where(and(inArray(game.id, gameIds), playableGameCondition()))
         : [];
-    const categories = await categoriesByGameIds(db, gameIds, true);
+    const [categories, tags] = await Promise.all([
+      categoriesByGameIds(db, gameIds, true),
+      tagsByGameIds(db, gameIds),
```
marek-chmielowski-blurify: **[efficiency]** `getCategoryGames` calls `tagsByGameIds` with the full lobby-category link list before filtering to only playable games, duplicating an existing over-fetch pattern (`categoriesByGameIds` already had the same issue pre-PR). A lobby category with many delisted/deactivated games now costs two unnecessary joined-row fetches per stale link instead of one.

packages/core/src/casino/shared/game-catalog.ts
```diff
@@ -98,3 +111,34 @@ export async function mappingsByProviderIds(
   }
   return grouped;
 }
+
+export async function tagsByGameIds(
```
marek-chmielowski-blurify, line 132: **[duplication]** `tagsByGameIds` re-implements the same select+join+where+orderBy-then-group-into-Map loop already present in `categoriesByGameIds` in this same file. Worth extracting a shared `groupBy`-style helper so both (and any future batched game-lookup) share one grouping implementation instead of copy-pasting it again.

packages/core/src/casino/gaming/service/game-tag.service.ts
```diff
@@ -0,0 +1,232 @@
...
+export const GameTagNameTakenError = makeConflictError(
+  'GameTagNameTakenError',
+  'A game tag with this name already exists',
+);
+export const GameTagSystemDeletionError = makeConflictError(
+  'GameTagSystemDeletionError',
+  'System game tags cannot be deleted',
+);
+
+type Actor = {
```
marek-chmielowski-blurify: **[duplication]** `GameTagService` re-declares a local `Actor` type identical to the one already independently duplicated in three sibling service files (`game-category.service.ts`, `game-provider.service.ts`, `gaming.service.ts`). A shared type would avoid a 4th copy that can silently drift out of sync.

packages/core/src/casino/lobby/service/lobby.service.ts
okapitula: Fixed in ccc0eac7. `getCategoryGames` now passes only the ids of games that survived the playable filter to `categoriesByGameIds` and `tagsByGameIds`, so delisted or deactivated links no longer trigger extra joined-row fetches. Response shape and ordering are unchanged.

packages/core/src/casino/shared/game-catalog.ts
okapitula: Fixed in ccc0eac7. Extracted a private `groupRows` helper in `game-catalog.ts` that keeps the query row order; `categoriesByGameIds`, `tagsByGameIds` and `mappingsByProviderIds` (which had the same loop) all use it now.

packages/core/src/casino/gaming/service/game-tag.service.ts
okapitula: Fixed in ccc0eac7. Added one `CatalogActor` type in `casino/shared/game-catalog.ts` and removed the local copies from the game-tag, game-category, game-provider and gaming services. It could not live in its own file under `service/` because the module-shape lint only allows `*.service.ts` there. The lobby service keeps its own `Actor` since its `actorId` is optional, so it is a different shape.

packages/core/src/casino/gaming/drizzle/migrations/0006_young_wrecker.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TYPE "public"."game_tag_type" AS ENUM('system', 'custom');--> statement-breakpoint
+CREATE TYPE "public"."game_tag_visibility" AS ENUM('visible', 'invisible');--> statement-breakpoint
+CREATE TABLE "game_tag" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"name" text NOT NULL,
```
zaxovaiko, line 5: I think name is redundant here. It should be translated on consumer's app anyway.

packages/core/src/casino/gaming/drizzle/migrations/0006_striped_betty_ross.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TYPE "public"."game_tag_type" AS ENUM('system', 'custom');--> statement-breakpoint
+CREATE TYPE "public"."game_tag_visibility" AS ENUM('visible', 'invisible');--> statement-breakpoint
+CREATE TABLE "game_tag" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"name" text NOT NULL,
+	"type" "game_tag_type" DEFAULT 'custom' NOT NULL,
+	"visibility" "game_tag_visibility" DEFAULT 'invisible' NOT NULL,
+	"badge_settings" jsonb DEFAULT '{"badgeColor":"#3377ff","textColor":"#ffffff"}'::jsonb NOT NULL,
```
zaxovaiko: The same with color.

packages/core/src/casino/gaming/schema/index.ts
```diff
@@ -117,6 +123,28 @@ export const game = pgTable(
...
 );
 
+export const gameTag = pgTable(
+  'game_tag',
+  {
+    id: uuid().primaryKey().defaultRandom(),
+    name: text().notNull(),
+    type: gameTagTypeEnum().notNull().default('custom'),
+    visibility: gameTagVisibilityEnum().notNull().default('invisible'),
+    badgeSettings: zodJsonb(GameTagBadgeSettingsSchema, 'game_tag.badge_settings')()
```
zaxovaiko: Let's remove `name` and `badgeSettings`.  That metadata must be kept in the frontend client, not in openora.

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -128,11 +131,18 @@ function toGame(row: {
...
+    tx
+      .select({ categoryId: gameCategoryGame.categoryId })
+      .from(gameCategoryGame)
+      .where(eq(gameCategoryGame.gameId, row.id))
+      .orderBy(asc(gameCategoryGame.categoryId)),
+    tx
+      .select({ tagId: gameTagGame.tagId })
+      .from(gameTagGame)
+      .where(eq(gameTagGame.gameId, row.id))
+      .orderBy(asc(gameTagGame.tagId)),
```
zaxovaiko, line 157: Should we add some limitation for number of fetched records?

zaxovaiko (changes requested):
Good in general, but let's remove `name` and `badgeSettings` from table.

packages/core/src/casino/gaming/schema/index.ts
okapitula: Fixed in 39ccb883. Replaced `badgeSettings` with a nullable `metadata` jsonb, same as `game` and `gameProvider`. Core no longer stores badge colours or ships default ones. An operator who wants to manage badge styling from their back-office can keep it there, and updates replace the whole object. Note that `metadata` is returned on public reads, so it shouldn't hold anything internal.

I kept `name`. It isn't the player-facing label: it's the operator's back-office label and the tag's unique key. The consumer app still translates it, using `name` as the key, the same way `gameCategory.name` works today. Without it, a tag would just be a UUID, which isn't stable between environments.

For context, the PR has two kinds of tags:
- **Custom tags** are what operators create through the admin API (create, update, delete). A tag created through the API is always `custom`.
- **System tags** can only be added directly in the database, and they can't be deleted through the API. This way nobody with `game-config` access can create a tag that no one can remove.

Each tag has its own `visible`/`invisible` setting, so an operator can prepare a custom tag before players see it.

packages/core/src/casino/gaming/drizzle/migrations/0006_young_wrecker.sql
okapitula: Kept as the back-office label and unique key, see my reply on `schema/index.ts`.

packages/core/src/casino/gaming/drizzle/migrations/0006_striped_betty_ross.sql
okapitula: Removed in 39ccb883. Colours are now covered by the free-form `metadata` field, see my reply on `schema/index.ts`.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: That query sits inside `gameAuditSnapshot`. It reads the tag links of one game, and `updateGame` already caps `tagIds` and `categoryIds` at 50 each, so it can't grow without limit. It's also the audit snapshot, so cutting it short would leave the audit record incomplete. The admin tag list is already paginated.

packages/core/src/contracts/schemas/game.ts
```diff
@@ -50,3 +50,21 @@ export const GameCategorySummaryWithTranslationsSchema = GameCategorySummarySche
...
+export const GAME_TAG_VISIBILITIES = ['visible', 'invisible'] as const;
+export const GameTagVisibilitySchema = z.enum(GAME_TAG_VISIBILITIES);
+export type GameTagVisibility = z.infer<typeof GameTagVisibilitySchema>;
+
+export const GameTagSummarySchema = z.object({
+  id: UuidSchema,
+  name: z.string(),
+  type: GameTagTypeSchema,
+  visibility: GameTagVisibilitySchema,
+  metadata: z.unknown().nullable(),
```
zaxovaiko: Do we need this metadata? Especially as an `unknown` wildcard

okapitula: Fair point, fixed in 897cfae1. I'd keep the field, though: it lets operators style tags from the backoffice without a frontend deploy.

It's now a record of JSON values. That's the best balance between flexibility and usability, and it's the same pattern as the lobby section config. The column uses the same schema through `zodJsonb`, so writes are validated too.

zaxovaiko (approved):

PR #169, opened 2026-09-14
feat(gaming): admin catalog stats route

okapitula:
## Summary

Adds `getCatalogStats` (`GET /backoffice/gaming/stats`) to the gaming admin contract. One call returns total, active and inactive counts for game providers, game categories and games, plus a playable count for games.

## Why

A backoffice catalog page shows four KPI tiles: providers total, inactive providers, games total and active games. Until now the only way to get those numbers was four paginated calls with `limit: 1`, reading `total` from each: `listAdminProviders` twice (once with `isActive: false`) and `listAdminGames` twice (once with `isActive: true`). Each call runs a row query plus a `count(*)`, and `listAdminGames` also joins providers and loads categories. Any operator with a catalog admin screen needs the same numbers, so the route belongs in core.

- The response is `{ providers, categories, games }`. Each has `{ total, active, inactive }`, counting the row's own `isActive` flag the same way the admin list filters do. `games` also has `playable`: active games whose provider is active too, using the same `playableGameCondition` the public catalog uses.
- One aggregate query per table (`count(*)` and `count(*) filter (where ...)`), run in parallel. No rows are loaded.
- Guarded by `game-config` `view`, like the admin list routes.
- Read-only: no migration, event or audit entry.

## Alternatives considered

- **Keep the four list calls:** rejected. It costs four row queries plus four counts, and the games list also joins providers and loads categories, for four numbers.
- **Count a game as active only when its provider is active:** rejected. The tile would disagree with the `isActive` filter on the games list it links to, and `inactive` would merge "game off" with "provider off". A separate `playable` count keeps both numbers.
- **One SQL statement with scalar subqueries:** not worth it. The three per-table queries are already cheap and run in parallel on a read-only admin path.

## Risks

None.


packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -169,6 +190,27 @@ export class GamingService {
...
+    const db = this.drizzle.db;
+    const [[providers], [categories], [games]] = await Promise.all([
+      db
+        .select({ total: count(), active: countWhere(eq(gameProvider.isActive, true)) })
+        .from(gameProvider),
+      db
+        .select({ total: count(), active: countWhere(eq(gameCategory.isActive, true)) })
+        .from(gameCategory),
+      db
+        .select({ total: count(), active: countWhere(playableGameCondition()) })
```
zaxovaiko: `games.active` counts `playableGameCondition()` here - game active AND provider active - but `listAdminGames` with `isActive: true` filters on `eq(game.isActive, true)` alone (line 241). Since this route is meant to replace that call, the tile and the list it links to disagree whenever a provider is off: the tile says 1, the list shows 2.

Providers and categories do not have the gap - their counts match the list filters in `game-provider.service.ts:126` and `game-category.service.ts:97` exactly.

It also makes `games.inactive` merge "the game is off" with "the provider is off", so whoever reads it cannot tell which one to fix, and disabling a single provider silently moves all its games into that bucket.

Either name the field for what it measures (`playable`), return both numbers, or make the admin list filter use the same condition.

okapitula: Agreed. `games.active`/`inactive` now count the game's own `isActive` flag, so they match the `listAdminGames` filter, and a new `games.playable` counts active games under an active provider. That keeps "game off" and "provider off" separate. The integration test covers an active game under an inactive provider (active, not playable). Fixed in 3f33da1a.

zaxovaiko (approved):

PR #170, opened 2026-09-14
chore(tooling): salvage the prose lint, mail-wait fix and cleaner from #164

zaxovaiko:
Salvages the parts of #164 that stand on their own, so that branch can be closed without losing them. The anti-slop oxlint rules, the readable-spacing churn, the vendored eslint-stylistic drop, the plan gate and the dispatch-only docs audit are all left behind deliberately.

Three commits, each independently revertable:

**`chore(tooling): gate prose pointers and fix the ones that had rotted`**

`tools/lint/verify-references.ts` walks tracked markdown and fails on a dead relative link, an anchor matching no heading, a backticked repo path that does not exist, a `pnpm <script>` that is neither a script nor a `node_modules/.bin` binary, a skill whose name disagrees with its directory, a roster entry with no subagent file, and two ADRs sharing a number. Added to `verify` as `check:refs`. It found seven rotted pointers, all fixed in the same commit: the duplicate ADR-0032, ADR-0024's description of packages ADR-0025 folded away, two subagents pointing at a deleted `tools/build-contract.ts`, a dead hooks path in the quickstart, a moved anchor in `architecture.md`, and a missing `.tpl` suffix in the consumer guide.

**`test: wait long enough for a mail job, and say what went wrong when it does not`**

`waitForEmail` allowed 5s. Mail is not sent inline: `MailService.enqueueToUser` puts a job on the real BullMQ queue per ADR-0039, so nothing reaches the capture until a worker has round-tripped through Redis. Raised to 20s, and the failure message now distinguishes captured-but-unmatched from nothing-captured.

**`feat(tooling): make docs follow code, and add a cleaner that applies`**

The three reviewer subagents are findings-only by charter, so nothing in the roster applied a mechanical fix. `cleaner` is the applier: formatting first, then comments that restate the line under them, then code it can prove unreferenced with `git grep`, then abstractions the branch added that have exactly one caller. Scoped to the branch diff, never changes behaviour, ends on `pnpm verify`, and preserves validation, authz, audit writes and `ponytail:` markers.

## Conflicts resolved against current `dev`

- `.rulesync/commands/pre-pr.md`: `dev` now documents `check:drift` as regenerating the catalog and has dropped the `system-design.md` table claim. Kept dev's facts and the branch's correction that `verify` already ends with `check:drift`.
- `.husky/pre-commit`: #166 rewrote the hook into a fast staged-files gate that does run `oxfmt` and `oxlint --fix`. The branch's replacement comment asserted that nothing rewrites files, which is no longer true, so dev's version stands and the matching paragraph was cut from the commit message.

## Verification

`check:lint`, `check:format`, `check:shape`, `check:hygiene` and `test:tools` pass locally (78 tool tests, including the new reference ones). `pnpm check:refs` is clean across 155 markdown files.

Not run locally: `build`, `check:boundaries`, `check:deprecations`, `test:unit`, `test:integration`, `check:drift`. The integration tier is the one the mail-wait change touches, so CI is the first real exercise of it.

Closes #164.


turbo.json
```diff
@@ -77,6 +77,15 @@
       "inputs": ["**/*", "!.turbo/**", "!**/node_modules/**", "!**/dist/**", "!**/.next/**"],
       "outputs": []
     },
+    "//#check:refs": {
+      "inputs": [
+        "docs/**/*.md",
```
klaudia-blazyczek-blurify: `check:refs` reads every tracked Markdown file, but these inputs omit root files such as `README.md` and `.github/**/*.md`. Turbo can reuse a cached pass after one of them changes, so its links are not checked. Include all Markdown read by the script (for example `**/*.md`, with exclusions as needed).

docs/standards/skills/delivery.md
```diff
@@ -41,12 +43,13 @@ Promotion follows `feature -> dev -> stage -> tag`.
 1. Determine the current branch and target.
 2. Inspect `git status -s` and stage only this work's files.
 3. Leave foreign or pre-existing edits untouched and report them.
-4. Create a lowercase conventional commit without AI trailers or sensitive data.
-5. Run `pnpm verify`.
-6. Report the commit SHA and stop for explicit per-action push confirmation.
-7. Push only after that confirmation.
-8. Reuse an existing matching PR if present; otherwise read `.github/pull_request_template.md`, complete its actual sections, then create the PR with `gh`.
-9. Report the PR URL.
+4. Pre-flight, before the commit so both land in it rather than a follow-up: run `cleaner` over the diff, and run `docs` scoped to the diff if the change touched a module surface, an adapter contract, a config field, a standard, or an "is" claim in an ADR.
```
klaudia-blazyczek-blurify, line 46: `docs` and `cleaner` scope their work with `git diff <base>...HEAD`, but this workflow runs them before the commit. That excludes staged and unstaged changes, so a new branch can get an empty scope and skip its actual work. Include working-tree changes or run these steps after the commit.

.rulesync/subagents/cleaner.md
```diff
@@ -0,0 +1,47 @@
...
+
+You are not a reviewer. Do not hunt for bugs, do not restructure, do not improve a design you merely disagree with. If you think something is wrong rather than merely surplus, report it and leave it alone.
+
+## Scope
+
+The branch diff and nothing else: `git diff <base>...HEAD --name-only`, default base `origin/dev`. A file the branch did not touch is out of scope even when it is worse than the ones that are. Pre-existing mess is someone else's PR.
+
+## Order
+
+1. **Mechanical first.** `pnpm fix:format`, then `pnpm exec oxlint <changed files> --fix`. Free, deterministic, no judgment. Everything below is judgment, so spend it on less.
```
klaudia-blazyczek-blurify: `pnpm fix:format` runs the root formatter across the workspace, not only the collected branch files. It can therefore edit unrelated work despite the diff-only guarantee. Pass the changed files to the formatter or remove the global formatting step.

klaudia-blazyczek-blurify (changes requested):

turbo.json
zaxovaiko: Fixed in a32fb0a7. Inputs are now `**/*.md` with `node_modules` and `dist` excluded, so the task hashes every file the script actually reads.

You were right about the size of it: the script walks `git ls-files`, and 49 of the 155 tracked markdown files sat outside the two declared globs - `README.md`, `CONTRIBUTING.md`, `SECURITY.md`, `CODE_OF_CONDUCT.md`, `.github/**`, `.changeset/**` and every package changelog.

Checked rather than assumed: appending a newline to `README.md` moved the task hash from `975b101fa4aa1e90` to `bb324620972828d5`, where before it did not move at all.

docs/standards/skills/delivery.md
zaxovaiko: Fixed in a32fb0a7, taking your first option: the scope now includes working-tree changes.

Both subagents read `git diff $(git merge-base <base> HEAD) --name-only` instead of `<base>...HEAD`. Diffing against the merge base with no second ref covers the index and the working tree, so the uncommitted work they are called on is in scope.

The change is in `cleaner.md` and `docs.md` rather than here, since each subagent owns its own scope definition and restating it in the workflow would give the rule two homes. Step 4 keeps its pre-commit position, which is what makes the cleanup land in the same commit.

Worth noting the failure was total, not partial: on a branch whose work was entirely uncommitted - the normal state at step 4 - the scope was empty, so both subagents reported nothing to do and the workflow read as if they had run.

.rulesync/subagents/cleaner.md
zaxovaiko: Fixed in a32fb0a7. Step 1 now passes the changed files: `pnpm exec oxfmt <changed files>`, then `node tools/lint/ensure-final-newline.mjs <changed files>`, then the `oxlint --fix` that was already scoped.

Both tools take paths already - `ensure-final-newline.mjs` handles an argument list explicitly, and its own comment says the pre-commit hook passes the staged set for exactly this reason, so `fix:format` was the only thing standing in the way.

The contradiction was internal too: the global format sat one step from a correctly scoped `oxlint <changed files> --fix` and eight lines above the file's own rule against touching a file outside the diff.

PR #173, opened 2026-09-15
feat(compliance): add per-provider geo-blocking

okapitula:
## Summary

Operators can now block every game of a game provider in a country. Admin routes under `/compliance/provider-geo-rules` add, remove and list provider rules. Every change is audited, and new rounds are checked against these rules alongside the existing global and per-game rules.

## Why

Per-game geo rules (#154) need one rule per game. A provider's licence restriction usually covers its whole catalogue. Blocking it game by game means many rules to maintain and audit, and any game added to the provider later is not blocked.

## Alternatives considered

- Bulk-create per-game rules for every game of the provider: rejected. Games added later are not covered, and the audit trail shows many unrelated rule changes instead of one decision.
- Let the gaming module pass the provider id to the geo check: rejected. Compliance looks up the game's current provider itself, so a caller cannot supply a wrong or stale provider, and an unknown game is refused.
- A generic helper shared by the game and provider rule methods: deferred. The new methods follow the #154 pattern, and merging the two can be its own refactor.

## Risks

- `GameGeoDenialReason` gains `provider_block` and `game_not_found`. Consumer code that handles every reason explicitly needs updating.
- `checkGame` on the geo-check port (`GAME_GEO_CHECK`) now denies a game id that does not exist instead of allowing it. `startRound` is unaffected because it rejects unknown games first; other callers of the port will see the new denial.
- When the player's country cannot be determined and the game's provider has any rule, the start is refused. A flaky or unbound GeoIP adapter therefore blocks every game of that provider, not just the games with per-game rules.


packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -197,25 +212,47 @@ export class ComplianceService {
...
+                countryCode ? eq(gameGeoRule.countryCode, countryCode) : undefined,
+              ),
+            ),
+        )}`,
+      })
+      .from(game)
+      .where(eq(game.id, input.gameId));
+
+    if (!rules) {
+      return { allowed: false as const, countryCode, reason: 'game_not_found' as const };
```
marek-chmielowski-blurify: This branch returns `countryCode` (the resolved value) instead of `null`, unlike the `geo_unresolved` branch below which nulls it out. The current caller (`startRound`) never hits this since `getGame()` already throws on a missing game first, but this is a new behavior on the shared `GAME_GEO_CHECK` port — any other consumer that reads `countryCode` alongside `reason` could mistake `game_not_found` for a normal geo-resolved decision. Worth nulling it here too for consistency, or documenting why it's intentional.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -197,25 +212,47 @@ export class ComplianceService {
     }
 
     const { countryCode } = globalDecision;
-
+    // Without a resolved country any rule counts, so an unresolved lookup fails closed.
+    const [rules] = await this.drizzle.db
```
marek-chmielowski-blurify, line 241: This now runs two correlated EXISTS subqueries against `provider_geo_rule` and `game_geo_rule` on every round start, purely to detect the `game_not_found` case — which `startRound` can never actually hit since `getGame()` already validates the game exists before calling `checkGame()`. That's extra query cost on the hot path (every bet) to support a branch the only real caller can't reach.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -525,4 +562,120 @@ export class ComplianceService {
       : await this.drizzle.db.select().from(gameGeoRule);
     return rows.map((row) => serializeRow(row, { dateFields: ['createdAt', 'updatedAt'] }));
   }
+
+  async upsertProviderGeoRule(
```
marek-chmielowski-blurify: `upsertProviderGeoRule`/`deleteProviderGeoRule`/`listProviderGeoRules` duplicate `upsertGameGeoRule`/`deleteGameGeoRule`/`listGameGeoRules` almost line-for-line (existence check, advisory-lock key, before/after select, `onConflictDoUpdate`, serialization). Might be worth factoring into a shared helper parameterized on the rule scope, otherwise a future fix to this shape (e.g. the advisory-lock flow) risks landing in one copy and not the other.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -197,25 +212,47 @@ export class ComplianceService {
...
+    const [rules] = await this.drizzle.db
+      .select({
+        providerRule: sql<boolean>`${exists(
+          this.drizzle.db
+            .select({ id: providerGeoRule.id })
+            .from(providerGeoRule)
+            .where(
+              and(
+                eq(providerGeoRule.providerId, game.providerId),
+                countryCode ? eq(providerGeoRule.countryCode, countryCode) : undefined,
```
marek-chmielowski-blurify: The `countryCode ? eq(...) : undefined` guard is duplicated identically for the provider and game subqueries. Could be computed once above and reused in both `where` clauses to avoid the two copies drifting if the unresolved-country semantics ever change.

packages/core/src/compliance/service/compliance.service.ts
okapitula: Good catch. `game_not_found` now returns `countryCode: null`: the denial isn't about location, so it shouldn't carry one. Fixed in 62f0ea6.

packages/core/src/compliance/service/compliance.service.ts
okapitula: The `game` read isn't only there for `game_not_found`: the port takes just `gameId` and `ipAddress`, so that row is how the check finds the game's provider. It's still one query per check, the same count as before this branch, and both EXISTS subqueries hit the unique `(provider_id, country_code)` / `(game_id, country_code)` indexes.

Having `startRound` pass the provider id in would drop the read, but the compliance check would then trust caller input, and a wrong provider id would let a provider-blocked game through. `game_not_found` stays as a fail-closed guard for other consumers of the port.

packages/core/src/compliance/service/compliance.service.ts
okapitula: Pulled the shared serialization into `serializeGeoRule` (62f0ea6). I kept the transaction and lock flow written out per rule type on purpose. The queries differ by table, and sharing more would take either type casts or a helper built from per-table query callbacks, which nothing else in the repo does. The chat ban and mute writes follow the same explicit per-table shape. Only the lock keys and serialization would really be shared, and both already live in one place.

packages/core/src/compliance/service/compliance.service.ts
okapitula: The two copies filter different columns (`providerGeoRule.countryCode` / `gameGeoRule.countryCode`), so it can't be a single value, but the check now lives in one local `matchesCountry(column)` helper used by both subqueries. Fixed in 62f0ea6.

packages/testing/src/__tests__/game-geo-blocking.e2e.test.ts
```diff
@@ -49,6 +50,41 @@ async function listRounds() {
   return (await readJson(response)) as Array<{ id: string; status: string }>;
 }
 
+function startFromBlockedCountry(gameId: string) {
+  return player.request('/gaming/rounds/start', {
+    method: 'POST',
+    headers: { 'content-type': 'application/json', 'x-real-ip': '198.51.100.40' },
```
jakubfilinger-b, line 57: Blocker: this test relies on a client-supplied `x-real-ip`, but `createApp` preserves that header and passes it into GeoIP enforcement. In a direct deployment, or behind a proxy that does not strip and replace it, a player can spoof an allowed-country IP and bypass the provider block. Please derive the default IP from the socket and trust forwarding headers only behind explicit trusted-proxy configuration, then add a spoofing regression test.

packages/core/src/compliance/service/compliance.service.ts
```diff
@@ -197,25 +216,45 @@ export class ComplianceService {
     }
 
     const { countryCode } = globalDecision;
-
+    // Without a resolved country any rule counts, so an unresolved lookup fails closed.
```
jakubfilinger-b: Please remove this narrative comment or express the rule through a named helper: `docs/standards/comments.md` permits comments only for facts the code cannot express, while the branch immediately below already defines this behavior.

packages/core/src/audit/plugin.ts
```diff
@@ -121,6 +121,26 @@ export async function mapEventToRecord(
     };
   }
 
+  if (
```
jakubfilinger-b, line 124: The new audit topics have no topic-specific persistence test: service tests only assert mocked event emission and schema tests parse the payload, but nothing proves `mapEventToRecord` stores the intended actor, resource, and before/after state. Please cover this mapper or assert the persisted audit records in the lifecycle E2E.

jakubfilinger-b:
Blocker: provider-country enforcement can be bypassed by spoofing `x-real-ip`; the other two comments cover a convention cleanup and missing audit-persistence coverage.

packages/testing/src/__tests__/game-geo-blocking.e2e.test.ts
okapitula: Right, but this is already on `dev`: `createApp` trusts a client-supplied `x-real-ip`, so per-game rules, the registration geo check and rate limits are all exposed today. Fixing it breaks every deployment behind a proxy and needs a design for trusting forwarded headers. I would do that in a separate PR with a spoofing test.

packages/core/src/compliance/service/compliance.service.ts
okapitula: Removed in 542abfc7.

packages/core/src/audit/plugin.ts
okapitula: Added in 542abfc7.

zaxovaiko:
@codex

packages/core/src/compliance/contract/index.ts
```diff
@@ -275,6 +301,21 @@ export const complianceContract = {
...
+    .route({ method: 'PUT', path: '/compliance/provider-geo-rules' })
+    .input(UpsertProviderGeoRuleInputSchema)
+    .output(ProviderGeoRuleSchema),
+
+  deleteProviderGeoRule: oc
+    .route({ method: 'DELETE', path: '/compliance/provider-geo-rules/{id}' })
+    .input(DeleteProviderGeoRuleInputSchema)
+    .output(ProviderGeoRuleSchema),
+
+  listProviderGeoRules: oc
```
zaxovaiko, line 324: Remember about pagination.

packages/core/src/compliance/contract/index.ts
```diff
@@ -275,6 +301,21 @@ export const complianceContract = {
     .input(ListGameGeoRulesInputSchema)
     .output(z.array(GameGeoRuleSchema)),
 
+  upsertProviderGeoRule: oc
```
zaxovaiko: Is there any limit of number of created rules?

packages/testing/src/__tests__/game-geo-blocking.e2e.test.ts
```diff
@@ -71,48 +114,31 @@ beforeAll(async () => {
...
-    await app.container.get(DRIZZLE).db.delete(gameRound).where(eq(gameRound.gameId, gameId));
-    await app.container.get(DRIZZLE).db.delete(gameGeoRule).where(eq(gameGeoRule.gameId, gameId));
-    await app.container.get(DRIZZLE).db.delete(game).where(eq(game.id, gameId));
-  }
-  if (app && providerId) {
-    await app.container.get(DRIZZLE).db.delete(gameProvider).where(eq(gameProvider.id, providerId));
+  if (app) {
+    const drizzle = app.container.get(DRIZZLE).db;
+    for (const { gameId, providerId } of seededGames) {
+      await drizzle.delete(gameRound).where(eq(gameRound.gameId, gameId));
```
zaxovaiko, line 124: Maybe we could add cascade delete? I assume the tables are connected with foreign keys. 
Of course, cascade delete isn't always great, but we need to remember to clean up properly.

zaxovaiko (dismissed):

packages/core/src/compliance/contract/index.ts
okapitula: The unique `(provider_id, country_code)` index allows one rule per country code per provider, and the code must be two uppercase letters, so a provider can have at most 676 rules. A hard cap would just get in the way of an operator who blocks most countries for a provider.

packages/testing/src/__tests__/game-geo-blocking.e2e.test.ts
okapitula: The tables have no foreign keys on purpose. The rule tables belong to compliance and `game`/`game_provider` belong to gaming, and the database standard forbids foreign keys across modules. Also, core has no way to delete a game or provider, so no rules are left behind in practice. The deletes in `afterAll` only clean up the rows this test inserts itself.

packages/core/src/compliance/contract/index.ts
okapitula: Added in 7ef1c266. Both geo rule lists (game and provider) now take `page`/`limit` and return `{ items, total, page, limit }`. They also filter by `gameIds[]` / `providerIds[]` (up to 100 ids).

zaxovaiko (approved):

PR #174, opened 2026-09-15
feat(gaming): read-only game catalog port for plugins

okapitula:
## Summary

Adds `GAME_CATALOG_READER`, a read-only port owned and registered by the gaming module. Plugins outside `casino/gaming` can use it to read playable games, active categories and active providers without importing the gaming tables.

## Why

Overlay lobby sections (`defineLobbySection`) need games, categories and providers. Until now the only way to get them was to import `@openora/core/casino/schema/gaming` and write their own Drizzle queries. That copied the playability rule into consumer code. One overlay checked only `game.isActive`, so games from a disabled provider still showed in its lobby sections. Each of those schema reads also blocks extracting gaming into its own service (ADR-0017).

The port applies core's own `playableGameCondition()` (game and provider both active), so consumers can't drift from core's visibility rule. It returns plain objects typed from the `game.ts` contract schemas, and every method answers a whole list of ids in one query:

- `getPlayableGames(ids)`, `getActiveCategories(ids, { withGameCount? })`, `getActiveProviders(ids)` and `getCategoryIdsByGame(ids)` return a `Map` in the order the ids were passed. Unknown, inactive, unplayable or malformed ids are left out rather than throwing.
- `listPlayableGamesInCategory(id, { limit })`, `listActiveCategoriesWithGameCount()` and `listActiveProviders()` return ordered lists.
- `gameCount` counts playable games only.

The gaming services now share their provider and category column sets and the `countWhere` helper through `casino/shared/game-catalog.ts`. The selected columns haven't changed.

## Alternatives considered

- **Export `GamingService`, `GameProviderService` and `GameCategoryService`:** module-structure rules forbid importing another module's service, and those services also carry audited admin writes. Their reads are paged or fetch a single row, so a lobby section would end up doing one query per game.
- **Expose `playableGameCondition()` on a public subpath:** consumers would still query gaming's tables directly, which keeps the extraction blocker and the duplicated queries.

## Risks

None.


packages/core/src/casino/gaming/adapters/game-catalog-reader.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    );
+    return inIdOrder(
+      ids,
+      new Map(
+        [...categoryIdsByGame].map(([gameId, categoryIds]) => [gameId, new Set(categoryIds)]),
+      ),
+    );
+  }
+
+  async listActiveCategoriesWithGameCount() {
```
zaxovaiko: What about pagination? I assume there won't be a lot of records here and in active providers, but still would be good to limit the fetching.

zaxovaiko (dismissed):
lgtm

packages/core/src/casino/gaming/adapters/game-catalog-reader.service.ts
okapitula: Done 963b5cbf

marek-chmielowski-blurify (approved):

PR #175, opened 2026-09-16
chore(agents): commit the root AGENTS.md with code review rules

zaxovaiko:
## Summary

Tracks the generated root `AGENTS.md` and adds a `Code Review Rules` section to its source, `.rulesync/rules/overview.md`, so an automated Codex review applies the repo's money, compliance, audit, authorization, contract and testing standards.

## Why

Codex review reads `AGENTS.md` from the checkout and never runs `pnpm install`, so the gitignored root brief never reached it. Only the nested module files were visible. `check:drift` now regenerates `AGENTS.md` and fails when the committed copy is stale, so the file stays generated from `.rulesync/`.

## Alternatives considered

- A GitHub Action running the Codex CLI after `pnpm gen:agents`, which keeps the file ignored. The official action only accepts an API key, which is billed separately, and a ChatGPT login stored as a secret is unsupported for CI and risky in a public repo.
- Committing every generated mirror (skills, subagents, `CLAUDE.md`). Codex review reads only `AGENTS.md`, so the rest stays ignored.

## Risks

A change under `.rulesync/rules/` now needs `pnpm gen:agents` and the regenerated `AGENTS.md` in the same commit, or `pnpm verify` fails at the drift step.


package.json
```diff
@@ -23,7 +23,7 @@
     "check:hygiene": "tsx tools/lint/verify-oss-hygiene.ts",
     "check:refs": "tsx tools/lint/verify-references.ts",
     "check:deprecations": "depretec --fail-on-found",
-    "check:drift": "tsx tools/gen/sync-tsconfig-paths.ts --check && tsx tools/gen/gen-catalog.ts",
+    "check:drift": "tsx tools/gen/sync-tsconfig-paths.ts --check && tsx tools/gen/gen-catalog.ts && rulesync generate && git diff --exit-code -- AGENTS.md",
```
klaudia-blazyczek-blurify: If a later PR deletes the tracked `AGENTS.md`, `rulesync generate` recreates it as an untracked file and `git diff --exit-code -- AGENTS.md` still succeeds. Check that the file is tracked too, otherwise hosted reviewers lose the root instructions again.

klaudia-blazyczek-blurify (changes requested):
Only one comment. Apart from that lgtm

package.json
zaxovaiko: Good catch, reproduced it. Done in `08777267` - `check:drift` now runs `git ls-files --error-unmatch AGENTS.md` before the diff, so an untracked `AGENTS.md` fails the check.

PR #178, opened 2026-09-16
feat(promo): grant creation with a durable idempotency guard

zaxovaiko:
## Summary

Stacked on #177. Review that one first; this diff is only the last three commits. Adds `promo_grant` - the row a player's bonus actually is - and `GrantService.grant()` bound to the `BONUS_GRANTS` port. BF-592.

## Why

**The grant row is the bonus balance.** There is no separate balance table and no second `wallet_balance` row: `promo_grant.bonus_balance` holds the funds, and they cross into the real balance exactly once, at conversion, as a `wallet_transaction` of type `bonus`.

That is the load-bearing decision in this PR, so the reasoning in full. `wallet_balance` is unique on `(wallet_id, currency)`, and it is read from roughly three dozen places - reconciliation, the custody sweep, swap, deposit, withdrawal, auto-withdrawal, the balance stream, admin reporting. Discriminating a bonus row by a `kind` column means every one of those reads needs a "real money only" predicate, and the one that gets missed is a player withdrawing bonus money as if it were their own.

Reconciliation is the sharper argument. It compares internal balances against on-chain custody. Bonus money was never deposited and has no chain side, so holding it in `wallet_balance` manufactures a permanent unexplained surplus on every reconciliation run.

Keeping it out also makes withdrawal correct by construction rather than by formula: there is nothing to subtract, because there is nothing there.

**The idempotency guard is the unique index, not a read.** `unique (user_id, source, source_ref)` with `onConflictDoNothing`, then a select for the existing row. A read-then-write check would let two concurrent replays of the same deposit both pass and grant the bonus twice; the index cannot. There is a test that fires three concurrent duplicates and asserts exactly one row and one `created: true`.

**`terms` is a snapshot, never a lookup,** and it snapshots the weight *rows*, not a reference to the profile. Editing an offer - or its weight profile - must not change a bonus a player already holds. Storing a profile id would have left that property true only by discipline; storing the resolved rows makes it true by construction. This is the fix for the one correction the design review raised against the original shape.

**Input is validated at the boundary.** A caller is another module, but a malformed multiplier or a non-positive amount has to be refused before it reaches the ledger rather than corrected after. Zod parse on entry.

**`expires_at` comes from Postgres**, `now() + make_interval(days => ...)`, so there is no clock to inject and no drift between the row's `created_at` and its expiry.

## Alternatives considered

**A second `wallet_balance` row with a `kind` discriminator** - rejected above, on the reconciliation and the thirty-odd unguarded reads.

**A `promo_grant_balance` table** - one row per grant per currency, which is one row per grant, which is a column.

**Four terminal timestamps** (`completed_at`, `expired_at`, `forfeited_at`, `cancelled_at`) - `status` already says which terminal state a grant closed in, so `closed_at` is one column carrying the same information.

**A column-add migration on top of the existing one.** The `0001` migration is regenerated in place instead. The table has never existed in a deployed database, so a follow-up `ALTER TABLE` would carry no information a reader could use.

## Input bounds and attribution

Three bounds sit at the boundary rather than after the fact. A wagering multiplier of zero
produces a grant that converts the moment it is created, so there is a floor, and a ceiling above
it so a fat-fingered multiplier is refused rather than written. Currency is uppercased at entry:
a grant stored as `usd` against a wallet holding `USD` is invisible to every bet in that currency
and quietly expires.

`BonusGrantArgs` carries an actor. `manual` is in the source enum because an admin hands a player
a bonus through it, and before this the only record of that said the system did it, with a null
actor.

The replay path used to return success for any row matching the idempotency key without checking
it was the same grant, so two payouts sharing one source reference would have reported a credit
that never happened. It compares the currency and the amounts now and refuses on a mismatch.

Three `CHECK` constraints hold the money invariants where no caller can route around them: a bonus
balance cannot go negative, wagering progress cannot pass its requirement, and a forfeit reason
cannot exist without a forfeit. The consumption path is written against these rather than trusted
to respect them.

ADR-0040 records the balance-model decision this PR implements.

## Risks

**No wallet credit yet.** `GrantService` writes the grant row and moves no money into a wallet; conversion is what moves money, and it arrives with the consumption path. A grant created today is a correct record of an obligation with no settlement behind it.

**No events emitted.** `promo.bonus.granted` is declared in #177 and nothing subscribes. Emitting it from inside the caller's transaction would require the transactional outbox, which is optional and throws when unbound, so the emit belongs to whoever owns the commit - the deposit subscription, once offers exist.

**No offer table,** so terms arrive from the caller. Eligibility (`ok: false` with `ineligible` / `offer_inactive` / `currency_unsupported`) is declared on the port and unreachable until offers land.

This PR also registers both promo migration sets in the shared test-database bootstrap. They were missing; the module's self-migrating integration tests hid it, but anything booted through the shared test app would have failed on a missing relation.


packages/core/src/contracts/adapters/bonus-grants.ts
```diff
@@ -13,6 +13,19 @@ import { createToken, type Token } from './token.js';
...
+ * The terms a grant is created under, snapshotted onto the grant row. Editing an offer must
+ * not change a bonus a player already holds, so nothing here is read from config again later.
+ */
+export type BonusGrantTerms = {
+  /** Wagering requirement as a multiple of the granted amount, as a decimal string. */
+  wageringMultiplier: string;
+  /** Days from the grant until it expires and what is left of it is forfeited. */
+  expiryDays: number;
+  /** Weight profile that scores this grant's bets. */
+  weightProfileId: string;
```
klaudia-blazyczek-blurify, line 21: This is only a pointer to mutable configuration, not a terms snapshot. `weightedContribution` reloads the profile rows by this id for every bet, so editing or deleting a profile changes or stops wagering for grants that already exist. Store the resolved weights here, or point to an immutable profile version.

klaudia-blazyczek-blurify (changes requested):

klaudia-blazyczek-blurify:
Are you planning on adding more for our "Stage 0"?

packages/core/src/promo/bonus/contract/index.ts
```diff
@@ -30,4 +30,41 @@ export const WagerWeightProfileSchema = z.object({
 
 export type WagerWeightProfile = z.infer<typeof WagerWeightProfileSchema>;
 
+/**
+ * `pending` is a grant that is claimed but not yet funded, `cancelled` its only exit - nothing
+ * was credited, so there is nothing to lose. Everything after funding ends in `completed`,
+ * `expired` or `forfeited`.
+ */
+export const BONUS_GRANT_STATUSES = [
```
klaudia-blazyczek-blurify, line 38: These grant value sets already exist in `contracts/schemas/promo.ts`, which is the shared event and adapter contract. Keeping another tuple, schema, and type here lets the database, input port, and event payloads drift. Import the shared values instead.

packages/core/src/promo/bonus/schema/index.ts
```diff
@@ -67,5 +82,82 @@ export const promoWeight = pgTable(
...
+    // zeroed by conversion, expiry or forfeiture.
+    bonusBalance: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE })
+      .notNull()
+      .default('0'),
+    wageringRequired: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull(),
+    wageringProgress: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE })
+      .notNull()
+      .default('0'),
+    status: promoGrantStatusEnum().$type<BonusGrantStatus>().notNull().default('active'),
+    forfeitReason: promoForfeitReasonEnum().$type<BonusForfeitReason>(),
```
klaudia-blazyczek-blurify, line 126: This only rejects a reason on a non-forfeited grant. It still allows `{ status: "forfeited", forfeitReason: null }`, even though every forfeit needs a reason for the audit trail. Enforce both directions in the database constraint.

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,184 @@
...
+// a fat finger rather than an offer. Both are refused before anything reaches the ledger.
+const MAX_WAGERING_MULTIPLIER = '1000';
+
+const grantArgsSchema = z.object({
+  userId: UuidSchema,
+  currency: CurrencyTickerInputSchema,
+  amount: MoneyAmountSchema.refine(isPositiveMoney, 'must be greater than zero'),
+  source: BonusGrantSourceSchema,
+  sourceRef: z.string().min(1),
+  actor: z.union([
```
klaudia-blazyczek-blurify, line 44: `source` and `actor` are validated independently, so `{ source: "manual", actor: { type: "system" } }` is accepted. That records an admin-issued bonus as a system action. Require an admin actor for a `manual` grant.

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * Creates the bonuses a player holds. Bound to BONUS_GRANTS, and always called on the caller's
+ * transaction handle so the grant commits with whatever earned it.
+ */
+export class GrantService implements BonusGrantCommands {
+  constructor(private readonly audit: AuditWritePort) {}
+
+  async grant(tx: DrizzleTx, rawArgs: BonusGrantArgs): Promise<BonusGrantOutcome> {
+    const args = grantArgsSchema.parse(rawArgs);
+    const wageringRequired = moneyScaleBy(args.amount, args.terms.wageringMultiplier);
```
klaudia-blazyczek-blurify, line 71: `MoneyAmountSchema` allows 20 integer digits, and the 1000 multiplier can produce a 23-digit `wageringRequired`. For example, `99999999999999999999` times `2` reaches the insert and overflows `numeric(38,18)`. Validate the derived amount, or cap the amount relative to the multiplier.

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,184 @@
...
+    const terms = await this.snapshotTerms(tx, args.terms);
```
klaudia-blazyczek-blurify, line 72: This reads the live weight profile before checking the idempotency key. If a grant already exists and its profile was later deleted, an exact retry throws here instead of returning the original grant. Resolve an existing key before requiring live configuration, while keeping the insert race-safe.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/contracts/adapters/bonus-grants.ts
zaxovaiko: This one is already covered in `3bf82f22` - `snapshotTerms` copies the profile rows into `terms.weights`, and wagering scores from the snapshot, never from the live profile. Editing or deleting the profile leaves an existing grant untouched (there are tests for both).

packages/core/src/promo/bonus/contract/index.ts
zaxovaiko: Done in `97923d04` - dropped the copies, the module imports them from `contracts/schemas/promo.ts` now.

packages/core/src/promo/bonus/schema/index.ts
zaxovaiko: Done in `97923d04` - the check is now `(status = 'forfeited') = (forfeit_reason is not null)`, so both directions are enforced.

packages/core/src/promo/bonus/service/grant.service.ts
zaxovaiko: Done in `97923d04` - a `manual` grant is refused unless the actor is an admin.

packages/core/src/promo/bonus/service/grant.service.ts
zaxovaiko: Done in `97923d04` - the derived `wageringRequired` is validated against `MoneyAmountSchema` before the insert.

packages/core/src/promo/bonus/service/grant.service.ts
zaxovaiko: Done in `97923d04` - an existing key is resolved before the profile is read. The insert keeps `onConflictDoNothing`, so the concurrent case is still race-safe.

zaxovaiko:
Stage 0 is complete in #177 - both modules, the `BONUS_GRANTS`, `WAGER_TRACKING` and `BONUS_WAGERING` ports, `context` on `WalletDebitArgs` and the promo event names. Everything from #178 up is on my side of the split.

zaxovaiko:
- `97923d04` -> `a9eb1fc2`
- `3bf82f22` -> `5e7a3cf6`

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,207 @@
...
+        'must be greater than zero',
+      ).refine(
+        (v) => moneyCompare(v, MAX_WAGERING_MULTIPLIER) <= 0,
+        `must not exceed ${MAX_WAGERING_MULTIPLIER}`,
+      ),
+      expiryDays: z.number().int().positive(),
+      weightProfileId: UuidSchema,
+    }),
+  })
+  .refine((a) => a.source !== 'manual' || a.actor.type === 'admin', {
```
klaudia-blazyczek-blurify: [BLOCK] This also accepts a non-manual grant with an admin actor, then writes an automated bonus as that admin's permanent audit action. That makes the regulated audit trail false. Require an admin actor exactly for `manual` and a system actor for every other source, with the reverse-case test.

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,207 @@
...
+    offerId: UuidSchema.optional(),
+    terms: z.object({
+      wageringMultiplier: MoneyAmountSchema.refine(
+        isPositiveMoney,
+        'must be greater than zero',
+      ).refine(
+        (v) => moneyCompare(v, MAX_WAGERING_MULTIPLIER) <= 0,
+        `must not exceed ${MAX_WAGERING_MULTIPLIER}`,
+      ),
+      expiryDays: z.number().int().positive(),
```
klaudia-blazyczek-blurify: [WARN] `expiryDays` has no upper bound before it is passed to PostgreSQL. For example, 2147483648 passes this schema but exceeds `make_interval`'s integer days argument and fails during the insert. Set a product-safe maximum here and cover the boundary.

packages/core/src/promo/bonus/service/grant.service.ts
```diff
@@ -0,0 +1,207 @@
...
+      .leftJoin(promoWeight, eq(promoWeight.profileId, promoWeightProfile.id))
+      .where(eq(promoWeightProfile.id, terms.weightProfileId));
+
+    if (rows.length === 0) {
+      throw new WagerWeightProfileNotFoundError(terms.weightProfileId);
+    }
+    const weights = rows.flatMap((r) =>
+      r.scope && r.contributionPercent
+        ? [{ scope: r.scope, scopeRef: r.scopeRef, contributionPercent: r.contributionPercent }]
+        : [],
```
klaudia-blazyczek-blurify: [WARN] An empty or all-zero profile reaches this return and creates an active grant that no wager can ever progress, so the player's bonus can only expire. Reject profiles with no positive usable weight before the insert and cover that case.

klaudia-blazyczek-blurify (changes requested):

zaxovaiko:
Fixed both of klaudia's remaining points, in 73de4c98:
- a non-manual grant could still name an admin actor, misattributing an automated bonus in the audit trail - the actor/source refinement now requires the pairing both ways, with a reverse-case test.
- expiryDays now has a sane upper bound (3650 days) before it reaches make_interval.
- also added the third one from the same review round: a weight profile with no positive weight (empty or all-zero) is now refused at grant time, since no wager could ever progress it - covered both the "no rows" and "all zero" cases.

This is stacked on #177, which I can't get to actually merge tonight (see my comment there) - GitHub's async merge-async endpoint for this stacked PR isn't honoring the repo's admin bypass. Stopping the stack at #177 for now; this branch is rebased and green on its side once that unblocks.

PR #181, opened 2026-09-17
feat: add per-IP login rate limit alongside per-account lockout

jakubfilinger-b:
## What

Adds a coarse per-IP throttle on `/identity/login`, separate from the existing per-account
lockout (`user.failedLoginAttempts` / `lockoutUntil`) and the existing per-email request
throttle (`login:<email>`, 10/5min).

## Why

The per-email throttle and per-account lockout both key on the email/account being attacked,
so they see repeated guessing against *one* account but are blind to credential stuffing
spread thin across *many* accounts from the same source (a botnet, a leaked-credential list).
A coarse per-IP counter closes that gap without touching the per-account behavior at all.

## How

- `RATE_LIMIT_KEYS.LOGIN_IP` (`'login-ip'`) added to the rate-limit key registry.
- `IdentityServiceOptions.loginIpRateLimit` (`enabled?`, `limit?`, `windowMs?`) lets a
  downstream operator tune or disable the threshold without a core change. Default is
  100 requests / 15 min - well above the 10/5min per-email budget so a shared network
  (office NAT) is never blocked.
- Fails closed (`onUnavailable: 'deny'`), matching the other credential-guessing keys
  (`login:`, `pwreset:`, `verify2fa:`, ...) - an unthrottled window during a limiter outage
  is worse than a transient 429 on this surface.
- Skipped entirely when the request carries no IP, rather than bucketing under a shared
  `'unknown'` key - that shared key would let one IP-less client's traffic exhaust the
  budget for every other IP-less client, a self-inflicted DoS.
- Not reset on a successful login (unlike the per-email key), so a successful login on one
  account can't be used to clear the IP's failure budget.

## Tests

`identity.rate-limit.int.test.ts` (real Redis): per-IP exhaustion across different accounts,
bucket isolation between IPs, the no-IP skip, the `enabled: false` override, fail-closed
behavior on backend unavailability.

`identity.service.int.test.ts` (real Postgres + Redis): locking one account never blocks a
different account signing in from the same IP.

99/99 tests green (`pnpm -F @openora/core test:integration`). `check:types`, `check:lint`,
`check:format`, `check:shape` clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -851,6 +865,19 @@ export class IdentityService {
     const email = input.email.toLowerCase();
 
     await assertRateLimit(this.limiter, makeLoginRateLimitKey(email), LOGIN_RATE_LIMIT);
+    // Skipped when the IP is unknown (no trusted proxy header) rather than bucketed under
+    // a shared 'unknown' key - that shared key would let one client's traffic exhaust the
+    // budget for every other client with no IP, which is a self-inflicted DoS.
+    const ipRateLimitOptions = this.options?.loginIpRateLimit;
+    if (ip && (ipRateLimitOptions?.enabled ?? true)) {
```
damianrzepka, line 872: This limiter uses an IP extracted from request headers. The runtime preserves a client-supplied `X-Real-IP`, so an attacker can rotate it on every attempt and create a fresh `login-ip:*` bucket. Sending only `X-Forwarded-For` also leaves the IP `null`, which skips this gate. Establish the IP at the trusted proxy/socket boundary before this path consumes it, and add an E2E proving spoofed headers cannot evade the limit.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -851,6 +865,19 @@ export class IdentityService {
     const email = input.email.toLowerCase();
 
     await assertRateLimit(this.limiter, makeLoginRateLimitKey(email), LOGIN_RATE_LIMIT);
+    // Skipped when the IP is unknown (no trusted proxy header) rather than bucketed under
+    // a shared 'unknown' key - that shared key would let one client's traffic exhaust the
+    // budget for every other client with no IP, which is a self-inflicted DoS.
+    const ipRateLimitOptions = this.options?.loginIpRateLimit;
```
damianrzepka, line 871: `loginIpRateLimit` accepts arbitrary numeric values. With `windowMs: -1`, Redis deletes the key after each `INCR`, so every request starts at count 1 and this limiter never exhausts. Validate positive, finite integer `limit` and `windowMs` values at the configuration boundary, with hostile-value coverage.

packages/core/src/contracts/adapters/identity.ts
```diff
@@ -11,8 +11,23 @@ export type IdentityLockoutOptions = {
...
 
+/**
+ * Coarse per-IP throttle on login attempts, separate from the per-account lockout
+ * above. Protects against credential stuffing across many accounts from one source
+ * (eg a botnet or a leaked-credential list), which the per-account counter can't see
+ * because each guess lands on a different account. Must stay well above the
+ * per-account threshold so a shared network (office NAT) never blocks legitimate
+ * users signing into their own accounts.
+ */
+export type IdentityLoginIpRateLimitOptions = {
```
damianrzepka, line 22: This new public option type is not re-exported from `@openora/core/contracts`, unlike `IdentityLockoutOptions`. Please add it to the contracts barrel so consumers can name the configuration type through the public subpath.

damianrzepka (changes requested):
NO-GO — the per-IP limiter can be bypassed through client-controlled IP headers, and invalid configuration values can disable it.

jakubfilinger-b:
Superseded by #184 (same change, renamed the branch).

PR #183, opened 2026-09-17
feat(gaming): bulk enable/disable, tag and category actions for games and providers

okapitula:
## Summary

Adds bulk catalog actions for the backoffice: enable/disable, add tags, and add categories across many games at once. A call can target chosen games, every game of chosen providers, or both, and each call is recorded in the audit trail.

## Why

Operators manage catalogs of thousands of games. Onboarding a provider, badging a batch of new releases, or pulling a studio's games meant one `PATCH /backoffice/gaming/games/{id}` per game. The audit trail then held hundreds of unrelated rows with nothing tying them to one decision.

- Three new routes, guarded by `game-config` `update`:
  - `POST /backoffice/gaming/games/bulk/active` `{ gameIds?, providerIds?, isActive }`
  - `POST /backoffice/gaming/games/bulk/tags` `{ gameIds?, providerIds?, tagIds }`
  - `POST /backoffice/gaming/games/bulk/categories` `{ gameIds?, providerIds?, categoryIds }`
- A target needs at least one non-empty list: `gameIds` (max 500) or `providerIds` (max 50). The affected set is every game named directly or owned by a named provider. A game reached both ways is changed and counted once.
- `providerIds` on `bulk/active` flips the provider's own `isActive` **and** every game of that provider. `gameIds` flips only those games. Enabling with `isActive: true` returns `unplayableGameIds`: games now active under a provider that is still disabled.
- Tags and categories are add-only. Existing links are kept, and re-adding a link is a no-op.
- Missing items are handled per item. Unknown game or provider ids come back in `notFound` and the rest of the call applies. An unknown tag or category id rejects the whole call with 404 and writes nothing.
- A call matching more than 5,000 games returns 400 (`GameBulkTooManyGamesError`) and writes nothing.
- **Audit:**
  - Each call that changes something emits one `gaming.games.bulk_updated` summary event.
  - For tag and category calls, the summary lists the exact `(gameId, ids)` links inserted, taken from the insert itself, so `before`/`after` show only what each game was actually missing.
  - `bulk/active` also emits a `gaming.provider.updated` event with a real before/after snapshot for each provider whose flag changed, so a search by provider id finds the change.
  - Rows from one call share a `correlationId`.
  - A repeat call that changes nothing writes no audit row.
- **Locking and SQL:**
  - Every bulk path locks games before providers, in id order, the same order `updateGame` uses. This avoids deadlocks with `updateGame`, `updateProvider`, and other bulk calls.
  - Games are re-scanned after the provider lock, so a game moved onto a provider mid-call is picked up. The call retries from a savepoint, at most 5 attempts.
  - Link inserts use `unnest` array parameters, so the bind-parameter count stays constant whatever the batch size.

## Alternatives considered

- **Replace semantics for tags and categories (like `updateGame`):** rejected. A bulk replace across a provider's games would silently strip labels that were set on individual games. Add-only is the safe bulk default; removal stays a per-game edit.
- **All-or-nothing on unknown game or provider ids:** rejected. A stale id in a 500-item selection shouldn't block the other 499. Unknown tag or category ids still reject the call, because they would fail every item.
- **One audit row per changed game:** rejected. A single provider-scoped call could write thousands of rows. The summary row keeps the exact per-game diff, and per-provider rows cover lookups by provider id.
- **Separate calls for games and providers:** rejected. One request with both lists is one decision, one transaction, and one audit record.

## Risks

- **Re-enabling cascades:** enabling a provider through `bulk/active` also re-enables games that were disabled one by one, for example a game not certified in a jurisdiction. This is intentional and fully audited, but operators should know about it.
- **Audit search:** per-game history of a bulk change is in the summary row's `changedGameIds`/`addedLinks`, not in `resourceId`. Searching the audit log by one game id won't show it; searching by provider id or `correlationId` will.
- **Lock scope:** a call near the 5,000-game cap row-locks those games for the length of the transaction, so a concurrent `updateGame` on them waits.
- **Contract:** additive only. There are 3 new routes and a new `gaming.games.bulk_updated` topic, and `gaming.provider.updated` gains an optional `bulkOperationId`. There's no schema change and no migration.
- **Changeset:** follows in a separate PR.


packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -306,6 +306,61 @@ export const UpdateGameInputSchema = z.object({
...
+function hasBulkTarget(target: { providerIds?: string[]; gameIds?: string[] }) {
+  return (target.providerIds?.length ?? 0) > 0 || (target.gameIds?.length ?? 0) > 0;
+}
+
+function bulkTargetRefinement() {
+  return {
+    message: 'Provide at least one non-empty providerIds or gameIds',
+    path: ['gameIds'],
+  };
+}
```
zaxovaiko: I think it can be an object directly for simplicity.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -385,4 +440,19 @@ export const gamingAdminContract = {
   getCatalogStats: oc
     .route({ method: 'GET', path: '/backoffice/gaming/stats' })
     .output(CatalogStatsSchema),
+
+  bulkSetGamesActive: oc
```
zaxovaiko: (optional) Tbh, I would prefer to name it just `setGamesActive` without the `bulk` prefix.

packages/core/src/contracts/schemas/events.ts
```diff
@@ -539,11 +540,46 @@ export const domainEventSchemas = {
...
+        operation: z.literal('set_active'),
+        actorId: UuidSchema,
+        bulkOperationId: UuidSchema,
+        target: z.object({ gameIds: z.array(UuidSchema), providerIds: z.array(UuidSchema) }),
+        isActive: z.boolean(),
+        changedGameIds: z.array(UuidSchema),
+        changedProviderIds: z.array(UuidSchema),
+        notFound: z.object({ gameIds: z.array(UuidSchema), providerIds: z.array(UuidSchema) }),
+      })
+      .extend(authContextBase.shape),
```
zaxovaiko: Can we extract them? Or actualy, don't we have duplicates here?

zaxovaiko (dismissed):
Overall, looks good. Short comments about naming.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Done, now an object.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Renamed, prefix dropped.

packages/core/src/contracts/schemas/events.ts
okapitula: Extracted into shared schemas.

zaxovaiko (approved):

PR #184, opened 2026-09-17
feat: add per-IP login rate limit alongside per-account lockout

jakubfilinger-b:
## What

Adds a coarse per-IP throttle on `/identity/login`, separate from the existing per-account
lockout (`user.failedLoginAttempts` / `lockoutUntil`) and the existing per-email request
throttle (`login:<email>`, 10/5min).

## Why

The per-email throttle and per-account lockout both key on the email/account being attacked,
so they see repeated guessing against *one* account but are blind to credential stuffing
spread thin across *many* accounts from the same source (a botnet, a leaked-credential list).
A coarse per-IP counter closes that gap without touching the per-account behavior at all.

## How

- `RATE_LIMIT_KEYS.LOGIN_IP` (`'login-ip'`) added to the rate-limit key registry.
- `IdentityServiceOptions.loginIpRateLimit` (`enabled?`, `limit?`, `windowMs?`) lets a
  downstream operator tune or disable the threshold without a core change. Default is
  100 requests / 15 min - well above the 10/5min per-email budget so a shared network
  (office NAT) is never blocked.
- Fails closed (`onUnavailable: 'deny'`), matching the other credential-guessing keys
  (`login:`, `pwreset:`, `verify2fa:`, ...) - an unthrottled window during a limiter outage
  is worse than a transient 429 on this surface.
- Skipped entirely when the request carries no IP, rather than bucketing under a shared
  `'unknown'` key - that shared key would let one IP-less client's traffic exhaust the
  budget for every other IP-less client, a self-inflicted DoS.
- Not reset on a successful login (unlike the per-email key), so a successful login on one
  account can't be used to clear the IP's failure budget.

## Tests

`identity.rate-limit.int.test.ts` (real Redis): per-IP exhaustion across different accounts,
bucket isolation between IPs, the no-IP skip, the `enabled: false` override, fail-closed
behavior on backend unavailability.

`identity.service.int.test.ts` (real Postgres + Redis): locking one account never blocks a
different account signing in from the same IP.

99/99 tests green (`pnpm -F @openora/core test:integration`). `check:types`, `check:lint`,
`check:format`, `check:shape` clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -851,6 +865,19 @@ export class IdentityService {
     const email = input.email.toLowerCase();
 
     await assertRateLimit(this.limiter, makeLoginRateLimitKey(email), LOGIN_RATE_LIMIT);
+    // Skipped when the IP is unknown (no trusted proxy header) rather than bucketed under
+    // a shared 'unknown' key - that shared key would let one client's traffic exhaust the
+    // budget for every other client with no IP, which is a self-inflicted DoS.
+    const ipRateLimitOptions = this.options?.loginIpRateLimit;
+    if (ip && (ipRateLimitOptions?.enabled ?? true)) {
```
damianrzepka: The limiter key is derived from `X-Real-IP`, but `extractClientMeta` accepts that header verbatim and `createApp` only falls back to the socket address when neither forwarding header is present. A caller can rotate `X-Real-IP` for each login (or send only `X-Forwarded-For` so `ip` is null), so it never exhausts a bucket and the new credential-stuffing protection is bypassed. Please derive the address at a trusted runtime/proxy boundary that overwrites client-supplied forwarding headers, and cover both spoof paths in an end-to-end test.

packages/core/src/contracts/adapters/identity.ts
```diff
@@ -11,8 +11,23 @@ export type IdentityLockoutOptions = {
...
 
+/**
+ * Coarse per-IP throttle on login attempts, separate from the per-account lockout
+ * above. Protects against credential stuffing across many accounts from one source
+ * (eg a botnet or a leaked-credential list), which the per-account counter can't see
+ * because each guess lands on a different account. Must stay well above the
+ * per-account threshold so a shared network (office NAT) never blocks legitimate
+ * users signing into their own accounts.
+ */
+export type IdentityLoginIpRateLimitOptions = {
```
damianrzepka, line 26: `IdentityLoginIpRateLimitOptions` is a new public configuration type, but it is not re-exported by the `@openora/core/contracts` barrel alongside `IdentityServiceOptions` and `IdentityLockoutOptions`. Downstream operators therefore cannot import the option type from the supported contract entry point. Please export it there and add a consumer-level type import assertion.

packages/core/src/pam/identity/__tests__/identity.rate-limit.int.test.ts
```diff
@@ -136,6 +136,88 @@ describe('IdentityService - rate limiting (real Redis)', () => {
...
+    }
+    const svc = withTemplateRenderer({ drizzle, events, limiter });
+
+    await expect(
+      svc.login(
+        { email: 'other-account@x.dev', password: 'password123' },
+        { 'x-real-ip': '203.0.113.42' },
+        new Headers(),
+      ),
+    ).rejects.not.toMatchObject({ code: 'TOO_MANY_REQUESTS' });
```
damianrzepka: This assertion accepts any rejection other than 429. Here `createAuth().api.signInEmail` is an unconfigured mock, so the test passes on a TypeError instead of proving that a login from a separate IP succeeds. Stub a successful sign-in response and assert the returned session.

damianrzepka (changes requested):
NO-GO — the per-IP credential-stuffing protection can be bypassed by spoofing the request IP.

packages/core/src/contracts/adapters/identity.ts
jakubfilinger-b: Good catch, fixed - `IdentityLoginIpRateLimitOptions` is now re-exported alongside `IdentityServiceOptions`/`IdentityLockoutOptions` in the contracts barrel (`f1fdbe80`).

packages/core/src/pam/identity/__tests__/identity.rate-limit.int.test.ts
jakubfilinger-b: Fair - that assertion would pass on the unconfigured `signInEmail` throwing too, not just on a real success. Stubbed a success response and now assert the actual returned session (`f1fdbe80`).

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Confirmed as a real characteristic, but it's not introduced by this PR: `extractClientMeta` reads `X-Real-IP` unconditionally (no `trustForwarded`-style gate, unlike `X-Forwarded-For`), and this per-IP gate reuses it identically to the four other per-IP throttles already merged in this same file - `register-ip:`, `verify-email-ip:`, `change-email-ip:`, `confirm-email-change-ip:`. If a caller can spoof `X-Real-IP` against a live deployment of this codebase, all five throttles are equally bypassable today, not just this new one.

The documented model (see `extractClientMeta`'s own comment) is that the header is only meaningful "when Node receives requests exclusively through a trusted reverse proxy ... configured to strip and replace the header" - that's an operational requirement on the deployment, not something `extractClientMeta` itself enforces. I've made that assumption explicit at both the option type and the call site (`9c9c69b3`) rather than leaving it implicit.

Properly closing the gap - validating a trusted-proxy boundary inside `extractClientMeta` itself, or refusing to trust `X-Real-IP` without an explicit opt-in the way `X-Forwarded-For` already requires - is a bigger, systemic change touching all five call sites, not something this PR can fix in isolation without reshaping a shared kernel helper. I'd rather track that as its own focused issue than bolt a partial fix onto this diff.

Given it matches existing, already-shipped behavior exactly, I don't think it should block this specific PR - but happy to open that follow-up issue now if you'd like it tracked, and open to being told I'm wrong about scope here.

jakubfilinger-b:
Addressed the barrel-export and test-assertion findings directly (\`f1fdbe80\`). Replied inline on the IP-spoofing finding with the scope analysis - it's a real characteristic, but pre-existing and shared by 4 other per-IP throttles already in this file (\`register-ip:\`, \`verify-email-ip:\`, \`change-email-ip:\`, \`confirm-email-change-ip:\`), not something this diff introduces. Made the trust assumption explicit in the code (\`9c9c69b3\`) rather than leaving it implicit. Over to you on whether that resolves the NO-GO or you'd rather see it tracked as a separate issue against \`extractClientMeta\` before this merges.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -851,6 +865,26 @@ export class IdentityService {
...
+    // budget for every other client with no IP, which is a self-inflicted DoS.
+    //
+    // `ip` comes from extractClientMeta(), which reads X-Real-IP unconditionally (see its
+    // doc comment) - a deployment MUST terminate at a reverse proxy that overwrites
+    // X-Real-IP with the real peer address (eg nginx `proxy_set_header X-Real-IP
+    // $remote_addr`) before this protection means anything; the same assumption already
+    // applies to register-ip:/verify-email-ip:/change-email-ip:/confirm-email-change-ip:
+    // elsewhere in this file.
+    const ipRateLimitOptions = this.options?.loginIpRateLimit;
+    if (ip && (ipRateLimitOptions?.enabled ?? true)) {
```
zaxovaiko: Hm, skipping the gate when `ip` is null makes it easy to turn off. If a request sends only `X-Forwarded-For` (no `X-Real-IP`), `create-app.ts` does not fill in the socket address, `extractClientMeta` returns `ip: null` since `trustForwarded` is not set, and the whole `login-ip:` check is silently skipped. The same happens with no attacker at all behind a load balancer that only sets `X-Forwarded-For` (AWS ALB, GCP LB), so the protection is just off in those deployments. The sibling keys use `ip ?? 'unknown'`, so they never switch off. Let's either fall back to a shared bucket there too, or at least log a warning when the gate is skipped, so an operator can see it is not active.

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -851,6 +865,26 @@ export class IdentityService {
     const email = input.email.toLowerCase();
 
     await assertRateLimit(this.limiter, makeLoginRateLimitKey(email), LOGIN_RATE_LIMIT);
```
zaxovaiko: Shouldn't the per-IP check run before the per-email one? Right now a request from an IP that is already over its limit still consumes one unit of `login:<email>` before it gets the 429, so a single IP can keep draining the 10/5min budget of any player it targets, with no cap. With the IP gate first, that is bounded by the IP budget.

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Agreed, and this was actually the wrong tradeoff - skipping is easier to disable than the shared-bucket risk it was avoiding. Switched to `ip ?? 'unknown'`, matching the sibling keys (`983e1130`).

packages/core/src/pam/identity/service/identity.service.ts
jakubfilinger-b: Good catch. Swapped the order - IP gate now runs first, so an already-over-limit IP is rejected before it can touch the targeted email's budget at all. Added a test asserting the email's 10/5min budget stays fully intact after that (`983e1130`).

zaxovaiko (dismissed):
More comments than actual code 😄

packages/core/src/pam/identity/service/identity.service.ts
```diff
@@ -850,6 +864,31 @@ export class IdentityService {
...
+    // simply not sending the header.
+    //
+    // `ip` comes from extractClientMeta(), which reads X-Real-IP unconditionally (see its
+    // doc comment) - a deployment MUST terminate at a reverse proxy that overwrites
+    // X-Real-IP with the real peer address (eg nginx `proxy_set_header X-Real-IP
+    // $remote_addr`) before this protection means anything; the same assumption already
+    // applies to the sibling keys above.
+    const ipRateLimitOptions = this.options?.loginIpRateLimit;
+    if (ipRateLimitOptions?.enabled ?? true) {
+      await assertRateLimit(this.limiter, makeLoginIpRateLimitKey(ip ?? 'unknown'), {
```
damianrzepka, line 882: This still keys the new protection from `X-Real-IP`, which `extractClientMeta` accepts verbatim. A direct caller can rotate that header for each login and receive a fresh `login-ip` bucket; the comment documents a reverse-proxy requirement but the runtime does not enforce it. Please derive the address at a trusted ingress boundary (or make trusted-proxy handling explicit and validated) and add runtime-level coverage for rotated `X-Real-IP`.

jakubfilinger-b: You were right to push back - documenting the trust assumption did not make it hold. Fixed at the ingress in `d022433d`, so it now covers every per-IP key, not just `login-ip:`.

`createApp` now decides the client address before any handler runs (`server/runtime/client-address.ts`). Forwarding headers are kept only when the socket peer is a trusted proxy. Any other peer gets its socket address written over `X-Real-IP` and its `X-Forwarded-For` dropped, so rotating either header gets a direct caller nothing. The trusted list is explicit and checked at boot: `CreateAppConfig.trustedProxies` first, then `TRUSTED_PROXIES` (comma-separated, empty = trust no proxy), then loopback + RFC 1918/4193. The default keeps an nginx-on-the-same-host or private-network deployment working. A malformed entry fails `createApp` before anything is bound. Requests with no socket at all (in-process `app.request`) are left as they are, so the existing IP-rotating test harness still works.

Tests:
- runtime level (`create-app.test.ts`): a public peer rotating `X-Real-IP` and `X-Forwarded-For` over three requests always resolves to its socket address. A configured proxy's `X-Real-IP` is honoured, while a private peer outside an explicit list is not. A bad entry refuses to boot.
- unit (`client-address.test.ts`): IPv4-mapped peers, CIDR parsing and the config/env/default precedence.

One thing the runtime cannot enforce: the proxy itself has to overwrite `X-Real-IP` (e.g. nginx `proxy_set_header X-Real-IP $remote_addr`) and not pass the client's value through. That is documented on the option. I also merged `dev` in, because the branch was behind.

packages/core/src/server/runtime/client-address.ts
```diff
@@ -0,0 +1,92 @@
...
+// instead of silently trusting nobody (or everybody behind the proxy it meant to name).
+export function parseTrustedProxies(entries: readonly string[]): TrustedProxies {
+  const list = new BlockList();
+  for (const raw of entries) {
+    const entry = raw.trim();
+    const [address = '', prefixText] = entry.split('/');
+    const family = isIP(address);
+    const prefix = prefixText === undefined ? undefined : Number(prefixText);
+    const maxPrefix = family === 4 ? 32 : 128;
+    if (
```
damianrzepka: A trailing slash is accepted as a /0 CIDR here: Number("") is 0, so trustedProxies: ["10.0.0.1/"] makes BlockList trust every IPv4 peer. A direct caller can then retain a rotating X-Real-IP and bypass the new login-IP throttle. Please reject an empty (and extra) CIDR suffix and cover it with a test.

packages/core/src/server/runtime/client-address.ts
```diff
@@ -0,0 +1,92 @@
...
+): void {
+  if (!peerAddress) {
+    return;
+  }
+  if (!trustedProxies.contains(peerAddress)) {
+    headers['x-real-ip'] = normalizeAddress(peerAddress);
+    delete headers['x-forwarded-for'];
+    return;
+  }
+  if (!headers['x-real-ip'] && !headers['x-forwarded-for']) {
```
damianrzepka: A trusted proxy that sends only X-Forwarded-For reaches this branch with no X-Real-IP, but the downstream extractClientMeta() is not invoked with trustForwarded. Its IP is therefore null, and all logins through that proxy share login-ip:unknown; 100 attempts block every user for 15 minutes. Normalize the validated XFF value into X-Real-IP here, or stop accepting XFF-only proxy configuration, and add coverage.

packages/core/src/server/runtime/client-address.ts
jakubfilinger-b: Confirmed - `Number('')` is 0, so `10.0.0.1/` became a /0 and every IPv4 peer was trusted. Fixed in `71c062e0`. Entries now have to match an anchored `^address(/digits)?$` pattern, so an empty suffix, a second `/…` segment, a sign or a space all fail `createApp` at boot. `10.0.0.1/`, `10.0.0.0/8/9`, `10.0.0.0/-1` and `10.0.0.0/ 8` are in the refusal table in `client-address.test.ts`, and the runtime boot test now uses `10.0.0.1/`.

packages/core/src/server/runtime/client-address.ts
jakubfilinger-b: Agreed - an XFF-only balancer (e.g. AWS ALB) put every login behind it into `login-ip:unknown`. I took the first option: normalise it at the ingress (`71c062e0`).

Behind a trusted peer, `X-Real-IP` now always holds a valid address:
- the proxy's own `X-Real-IP`, if it sent a parseable one;
- otherwise the client from `X-Forwarded-For`, read right to left, skipping trusted hops and stopping at the first untrusted one. Entries the client wrote ahead of the proxy's are never read. A malformed hop ends the walk at the last address that could still be read;
- otherwise the proxy's socket address.

So `extractClientMeta()` needs no `trustForwarded` here, and `ip` is never null for a request that came in over a socket.

Coverage:
- unit: XFF-only, client-prepended entries, chained proxies, a malformed hop, garbage `X-Real-IP`;
- runtime (`create-app.test.ts`): two clients behind an XFF-only proxy on `10.0.0.2` resolve to two distinct addresses, not null.

damianrzepka (approved):

PR #189, opened 2026-09-17
feat(promo): append-only bonus grant ledger

zaxovaiko:
## Summary

Stacked on #178. Adds `promo_grant_entry`, the append-only bonus ledger, and has `GrantService` open it with the grant's first movement. BF-592.

## Why

`wallet_transaction` is the real-money ledger and stays that way: one row per bet, per win, per deposit, for the full amount. Under the separate-balance model a whole class of movement never touches the real balance - a stake paid from bonus funds, a win attributed back to the grant that funded it, a conversion, an expiry, a forfeit - and the acceptance criterion is that *every* credit, debit, conversion and forfeiture is recorded. Those movements need a ledger of their own.

**`bonus_amount` is a signed delta, and that is the design.** The sum of a grant's entries always equals its `bonus_balance`. One identity, asserted in the tests rather than left as an intention:

```sql
select sum(bonus_amount) from promo_grant_entry where grant_id = $1  -- = promo_grant.bonus_balance
```

If it ever breaks, money moved outside the ledger, and that is the one class of bug in this area that cannot be recovered from after the fact.

**`user_id` is denormalised off the grant.** The win path has to answer "which grant funded this provider round" on every credit, and the index `(user_id, external_round_id) where external_round_id is not null` answers it with one read and no join. The round id lives on the entry rather than on the grant because a player can hold several grants and a grant spans many rounds - the round is a property of the movement, not of the bonus.

**`real_amount` sits beside `bonus_amount` on the same row.** A stake split half from cash and half from bonus wins back proportionally, and the proportion is exactly these two numbers. Storing only the bonus half would mean recomputing the split from the wallet ledger at settlement time, across a module boundary, for a number this row already knows.

**The entry is written only on the created path.** A replayed grant returns the existing row and touches nothing. The idempotency guard has to cover the ledger as well as the grant, or a retried deposit doubles a player's recorded bonus while the grant row stays correct - the worst kind of drift, because the two disagree silently.

**The grant foreign key is `on delete restrict`, not `cascade`.** A ledger that a single `DELETE FROM promo_grant` can erase is not a ledger. A grant that has to go away gets a `forfeit` or `expire` entry; the history stays.

## Alternatives considered

**Recording progress increments as ledger rows.** The table would grow one row per bet per active grant, the largest table in the system at real slot volume. It also is not what the requirement asks for: a wagering-progress increment is not a credit, a debit, a conversion or a forfeiture. Only money movements are logged; progress lives on the grant. The cost is losing per-bet attribution when a player disputes why their bonus never completed - at which point the `stake` rows this table already carries are the answer, since a bonus-funded stake *is* a money movement.

**A cross-module FK to the wallet transaction.** `wallet_transaction_id` is a bare uuid with no reference, per the module-boundary rule. `grant_id` does carry a real FK, because that one is inside this module.

**Enforcing append-only with a database trigger.** Not done, and worth naming rather than leaving implied: nothing in this repository uses one, including `audit_log`, whose immutability is the stricter requirement. `on delete restrict` closes the erasure path that actually existed; no code path updates an entry, and a trigger would be a local exception to a house rule.

## Risks

Nothing writes a `stake`, `win`, `reversal`, `convert`, `forfeit` or `expire` entry yet - those arrive with the consumption path and the lifecycle sweep. The type enum is declared in full up front so neither of those PRs has to touch this migration.


packages/core/src/promo/bonus/contract/index.ts
```diff
@@ -67,4 +67,24 @@ export const BONUS_GRANT_SOURCES = [
...
 
+/**
+ * One movement on a grant's own ledger. Append-only: a correction is another row, never an edit,
+ * because the spec demands every credit, debit, conversion and forfeiture be recorded immutably.
+ *
+ * `stake` and `win` carry the provider round, which is how a win finds the grant that funded the
+ * bet. `reversal` undoes a voided round: the bonus stake goes back and its wagering progress with
+ * it, because money returned without progress returned is free wagering bought by a rollback.
+ */
+export const BONUS_GRANT_ENTRY_TYPES = [
```
klaudia-blazyczek-blurify, line 78: `BONUS_GRANT_ENTRY_TYPES` already lives in `contracts/schemas/promo.ts`, alongside the event and adapter contracts. A second tuple, schema, and type here can drift from the ledger enum and event payloads. Import the shared values instead.

packages/core/src/promo/bonus/schema/index.ts
```diff
@@ -158,6 +160,57 @@ export const promoGrant = pgTable(
...
 
+export const promoGrantEntryTypeEnum = pgEnum('promo_grant_entry_type', BONUS_GRANT_ENTRY_TYPES);
+
+/**
+ * The bonus ledger: one row per movement on a grant, append-only. `wallet_transaction` stays the
+ * real-money ledger; a movement that never touches the real balance belongs here, and the spec
+ * still requires it be recorded immutably.
+ *
+ * `bonusAmount` is a signed delta, so the sum of a grant's entries always equals its
+ * `bonus_balance`. If that identity ever breaks, money moved outside the ledger.
```
klaudia-blazyczek-blurify, line 171: `onDelete: restrict` only protects the parent grant. The application role can still `UPDATE` or `DELETE` a `promo_grant_entry`, which breaks the immutable ledger and the `sum(entries) = bonusBalance` invariant. Enforce append-only writes at the database boundary, not only by convention.

packages/core/src/promo/bonus/schema/index.ts
```diff
@@ -158,6 +160,57 @@ export const promoGrant = pgTable(
...
+    currency: text().notNull(),
+    type: promoGrantEntryTypeEnum().$type<BonusGrantEntryType>().notNull(),
+    // Signed: negative on a stake, positive on a grant or a win, negative on a conversion.
+    bonusAmount: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull(),
+    // The real-money half of the same bet. The win split needs both sides of the stake.
+    realAmount: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull().default('0'),
+    wageringDelta: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE })
+      .notNull()
+      .default('0'),
+    balanceAfter: decimal({ precision: MONEY_PRECISION, scale: MONEY_SCALE }).notNull(),
```
klaudia-blazyczek-blurify, line 196: the target schema also needs `(user_id, created_at)` for player bonus history. `(grant_id, created_at)` only serves one grant, while the round index serves settlement; a history query will otherwise scan and sort the ledger. Add the user-history index as well.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/contract/index.ts
zaxovaiko: fixed in the rebase onto #178. Deleted the local redeclarations, `schema/index.ts` and everything else already import `BONUS_GRANT_ENTRY_TYPES`/`BonusGrantEntryType` from `@openora/core/contracts`. Done in `3a56ca72`.

packages/core/src/promo/bonus/schema/index.ts
zaxovaiko: fair, restrict only stops a cascading delete of the parent. Added a trigger that denies UPDATE/DELETE on `promo_grant_entry`, same pattern as `audit_log_append_only` in the audit module (`postSql` in `migrate.ts`, not hand-edited into the migration). Done in `3a56ca72`.

packages/core/src/promo/bonus/schema/index.ts
zaxovaiko: added `(user_id, created_at)`. Done in `3a56ca72`.

zaxovaiko:
- `3a56ca72` -> `229d9af4`
- `567f99a6` -> `4d1d2e9d`

packages/core/src/promo/bonus/schema/index.ts
```diff
@@ -157,6 +159,60 @@ export const promoGrant = pgTable(
...
+ * real-money ledger; a movement that never touches the real balance belongs here, and the spec
+ * still requires it be recorded immutably.
+ *
+ * `bonusAmount` is a signed delta, so the sum of a grant's entries always equals its
+ * `bonus_balance`. If that identity ever breaks, money moved outside the ledger.
+ *
+ * `userId` is denormalised from the grant so the win-attribution lookup by round is one indexed
+ * read with no join.
+ */
+export const promoGrantEntry = pgTable(
```
klaudia-blazyczek-blurify: [BLOCK] The ledger is immutable, but `promo_grant.bonus_balance` can still be updated directly. After a normal grant, setting that balance to `0` succeeds while the entries still sum to the original credit. This breaks the stated financial invariant. Add a deferred constraint trigger that checks the entry sum against the grant balance at commit, and cover direct balance drift.

packages/core/src/promo/bonus/schema/index.ts
```diff
@@ -157,6 +159,60 @@ export const promoGrant = pgTable(
...
+  'promo_grant_entry',
+  {
+    id: uuid().primaryKey().defaultRandom(),
+    // Same module, so a real FK. Restrict, not cascade: a ledger a single DELETE can erase is
+    // not a ledger, and a grant that has to go away gets a forfeit or expire entry instead.
+    grantId: uuid()
+      .notNull()
+      .references(() => promoGrant.id, { onDelete: 'restrict' }),
+    userId: uuid().notNull(),
```
klaudia-blazyczek-blurify: [WARN] `userId` and `currency` are denormalised but are not tied to the referenced grant. A future writer can attach grant A to player B or another currency, so round lookup and player history can use the wrong grant. Enforce the three fields together with a composite FK to a unique `(id, user_id, currency)` parent key, or an equivalent insert trigger.

packages/core/src/promo/bonus/migrate.ts
```diff
@@ -4,6 +4,22 @@
...
+// application role can still UPDATE or DELETE a promo_grant_entry row directly, which breaks
+// the immutable ledger and the "sum(entries) = bonusBalance" invariant. Same pattern as the
+// audit log: a trigger, not a convention, is the boundary drizzle-kit cannot express.
+const APPEND_ONLY_SQL = [
+  `CREATE OR REPLACE FUNCTION promo_grant_entry_deny_mutation() RETURNS trigger LANGUAGE plpgsql AS $$
+   BEGIN
+     RAISE EXCEPTION 'promo_grant_entry is append-only: % is not permitted', TG_OP;
+   END;
+   $$`,
+  `DROP TRIGGER IF EXISTS promo_grant_entry_append_only ON promo_grant_entry`,
```
klaudia-blazyczek-blurify: [WARN] `runMigrations` executes these post-SQL statements separately, so every migration drops the protection before recreating it. A concurrent connection can update or delete a ledger row in that gap. Create the trigger atomically, or create it only when absent, so the table is never committed without append-only protection.

klaudia-blazyczek-blurify (changes requested):

zaxovaiko:
Fixed all three of klaudia's points, in b3b8e31e:
- [BLOCK] a direct update of promo_grant.bonus_balance could drift from the ledger's entry sum with nothing to catch it. Added a deferred constraint trigger (checked at commit, so grant creation - one insert, one entry, same transaction - still passes) on both promo_grant_entry and promo_grant(bonus_balance), plus a test that a bare balance write is now refused.
- [WARN] the append-only trigger's CREATE/DROP/CREATE was three separate round trips through runMigrations' postSql, leaving the table briefly uncovered on every migration run. Combined into one multi-statement string, which Postgres runs as one implicit transaction.
- [WARN] promo_grant_entry's userId/currency are now tied to their grant with a composite FK against a new (id, userId, currency) unique index on promo_grant, so an entry can't be attached to the wrong player or currency.

Same as #178 - stacked on #177, which isn't merging tonight for reasons unrelated to this code (see #177). Stopping the stack there; this branch is rebased and green on its side.

PR #190, opened 2026-09-18
feat(promo): bet-time bonus consumption and proportional win settlement

zaxovaiko:
## Summary

Stacked on #189. The bet-time half of the engine: a stake decides which balance pays it, wagering advances on exactly one grant, a met requirement converts to real money, and a win or a void settles back in the proportion the stake was paid in. BF-592.

## Why

This is the part that moves a player's money, so the reasoning is written out rather than implied.

**The split, and where it sits.** `debit()` now takes the real balance first and the bonus only for what it could not cover - the operator's stated order. The engine call sits **below** the wallet's existing duplicate-provider-reference guard, which is the whole dedupe story: a replayed wager returns at that guard and can never reach the code that spends bonus funds or advances progress. A second check inside the engine would be a copy that drifts out of sync with the first. There is a test that sends the same `providerRef` twice and asserts one contribution, one spend, one ledger entry.

**One bet feeds exactly one grant.** Attribution is a single locked read: the active grant in the bet's own currency, earliest expiry first, with a stable tiebreak so two concurrent bets take the lock in the same order. The bonus **spend** comes from that same grant - there is no waterfall across several. That is what keeps a win attributable: a stake drawn from three grants would have to be split three ways on the way back, for a case the product rule says cannot happen.

**A grant in another currency does not participate.** No implicit conversion, no cross-currency scoring. Failing closed here costs a player nothing they were promised; the opposite mistake releases a bonus early and cannot be taken back.

**The conditional update is the concurrency guard.** Bonus is spent with `UPDATE ... WHERE bonus_balance >= $amount AND status = 'active'`; zero rows means another bet got there first. Progress advances with `LEAST(required, progress + weighted)` and flips the status in the same statement, so the cap and the completion decision cannot disagree. A test fires twenty concurrent bets against a balance that covers ten and asserts exactly ten succeed, the balance lands on zero and never below, and the ledger still sums to it.

**The engine never calls back into the wallet.** A completed grant is reported as `convertedAmount`; the wallet, which already holds the transaction, performs the credit and writes the `wallet_transaction` of type `bonus`. Conversion happens only at bet time - expiry and forfeit convert nothing, and a win does not complete a requirement - so there is exactly one conversion site and no dependency cycle between the two modules.

**The win path finds its grant through the ledger.** One indexed read of the round's `stake` entries gives the bonus and real halves of the stake; the win splits in that proportion, truncating, so a rounding remainder lands on the real balance rather than the bonus one. A round with no bonus stake finds no rows and credits normally.

**A voided round returns the bonus stake and takes back the progress it bought.** Money back with progress left standing is free wagering bought by a rollback.

**`wallet_transaction` still records the full amount** of every bet and every win. The split lives only in `promo_grant_entry`, so gross-flow reporting stays honest.

**A defect the end-to-end pass turned up.** A stake paid entirely from bonus funds reached `debitWalletBalance` for zero, matched no row, and was reported as refused - *after* the bonus had already been spent. The wallet said no while the grant had already paid. A stake with no real part now touches no real balance at all.

Three findings from the security pass, all of them paths that released money the terms had not:

- **The in-core bet path never named its round.** `startRound` debited before inserting `game_round`, so no round id reached the engine, and `endRound` credited the win without one. Every bonus-funded win on the only bet path core ships was therefore credited to the real balance, bypassing the requirement outright. The round id is now minted before the debit and carried on both sides, with the game context the weighting needs.
- **`settle()` failed open.** A movement the grant refused looked exactly like a round that drew no bonus, and both credited the full amount as real. A win on a completed grant does belong to the real balance - its funds converted at completion - but a reversal does not, and that case now stops with a named error instead of paying out.
- **The conversion was unaudited.** The single movement that turns a bonus into withdrawable money wrote a ledger row and nothing else. It now writes an audit record with the balance before and after and the grants that completed, on the same transaction.

Narrower, from the same pass: a round drawing on several grants splits across all of them in proportion rather than taking whichever row Postgres returned first; a partial void reverses only the progress the returned part bought, and a repeated reversal callback refunds nothing; and the balance a debit decides on is read under a row lock, because the `wallet` lock above it does not cover `wallet_balance`.

## Alternatives considered

**Spending across several grants when the first cannot cover the stake.** Rejected: it makes a win unattributable and multiplies every downstream rule - the max-win cap, the forfeit-takes-winnings rule - by the number of grants a stake happened to touch. A stake the chosen grant cannot cover is refused instead.

**Publishing a domain event and letting the bonus module subscribe.** The bus is post-commit and best-effort; a lost message is lost wagering progress on a player's money.

**Reversing a void by writing a compensating entry only.** The progress has to come back off the grant as well, or a provider that voids rounds is a wagering generator.

**Enforcing `max_bet` here.** Deferred, deliberately: it is a field of an offer, and there is no offer table yet. It belongs with the configuration surface that sets it, enforced in this same transaction against the terms snapshot.

## Risks

Cross-currency wagering, a conversion cap on winnings, and whether a cash bet should still count once bonus funds are spent are all open product questions. The defaults here are: no cross-currency participation, no cap, and the full stake counts. Each is a field on the `terms` snapshot away from being configurable, and `terms` is jsonb, so none of them needs a migration.

The wallet-commands constructor is now at five positional parameters and should take a named object for its optional ports; that refactor touches call sites the next PR deletes, so it rides with the teardown rather than here.

Wallet's older `bonus_credit` rollover model is still present and still runs on the same bet. The two are disjoint - one reads `wallet_balance`, the other the grant - but `debit()` now has two bonus branches, which is exactly one more than should exist. The next PR removes the old one.


packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,354 @@
...
+    const advanced = await this.advanceProgress(tx, grant.id, weighted);
+    const completed = advanced?.status === 'completed';
+    const convertedAmount = completed ? spent.balanceAfter : ZERO;
+    const balanceAfter = completed ? ZERO : spent.balanceAfter;
+
+    if (completed) {
+      await this.closeCompleted(tx, grant.id);
+    }
+
+    await tx.insert(promoGrantEntry).values({
```
klaudia-blazyczek-blurify, line 96: this writes a stake, win, or reversal entry and changes the grant without a transactional audit record or declared domain event. Only conversion writes through `AUDIT_WRITER`, so the other bonus money movements have no actor or before/after audit trail.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,354 @@
...
+      grantId: grant.id,
+      userId: args.userId,
+      currency: args.currency,
+      type: 'stake',
+      bonusAmount: `-${args.fromBonus}`,
+      realAmount: moneySubtract(args.stake, args.fromBonus),
+      wageringDelta: advanced ? applied : ZERO,
+      balanceAfter: spent.balanceAfter,
+      ...(args.externalRoundId === undefined ? {} : { externalRoundId: args.externalRoundId }),
```
klaudia-blazyczek-blurify, line 105: a bonus-funded bet can omit `externalRoundId`. Its stake then cannot be found by `settle()`, and a later win bypasses the bonus ledger and is credited as real money. Require a stable round identity before spending bonus funds.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,354 @@
...
+            ? moneySubtract(realShare, returned)
+            : moneyDivide(moneyScaleBy(realShare, row.real), totalReal);
+      if (moneyCompare(slice, ZERO) <= 0 && moneyCompare(realSlice, ZERO) <= 0) {
+        continue;
+      }
+
+      const applied = reversal
+        ? await this.reverseRound(tx, row.stake, slice)
+        : await this.creditBonus(tx, row.stake.grantId, slice);
+      if (applied === null) {
```
klaudia-blazyczek-blurify, line 191: `creditBonus()` returns `null` for expired and forfeited grants as well as completed ones, but this branch treats all three as cash settlements. A delayed win after expiry or forfeiture is therefore paid to the real balance. Distinguish `completed` from terminal states that must not release funds.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,354 @@
...
+  private async roundStakes(
+    tx: DrizzleTx,
+    userId: Uuid,
+    round: NonNullable<PromoGrantEntry['externalRoundId']>,
+  ): Promise<RoundStake[]> {
+    // What the round took, and what earlier callbacks already gave back, in one pass: a second
+    // rollback has to net against the first rather than be waved through or refused outright.
+    const onStake = sql`${promoGrantEntry.type} = 'stake'`;
+    const onReversal = sql`${promoGrantEntry.type} = 'reversal'`;
+    return tx
```
klaudia-blazyczek-blurify, line 302: two reversal callbacks with different provider references can both read the same outstanding stake before either reversal is inserted. Both then add it back to the grant, so one bonus stake can be refunded twice. Serialize settlement per round or make the outstanding amount an atomic database guard.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,354 @@
...
+        grantId: promoGrantEntry.grantId,
+        bonusStake: sql<string>`(-sum(case when ${onStake} then ${promoGrantEntry.bonusAmount} else 0 end))::text`,
+        realStake: sql<string>`sum(case when ${onStake} then ${promoGrantEntry.realAmount} else 0 end)::text`,
+        weightedTotal: sql<string>`sum(case when ${onStake} then ${promoGrantEntry.wageringDelta} else 0 end)::text`,
+        bonusReturned: sql<string>`sum(case when ${onReversal} then ${promoGrantEntry.bonusAmount} else 0 end)::text`,
+        realReturned: sql<string>`sum(case when ${onReversal} then ${promoGrantEntry.realAmount} else 0 end)::text`,
+      })
+      .from(promoGrantEntry)
+      .where(
+        and(
```
klaudia-blazyczek-blurify, line 313: this lookup uses only user and `externalRoundId`. It ignores the callback currency and the provider identity was dropped from the port, so equal round IDs from different providers or currencies can settle the wrong grant. Persist and query a provider-qualified round key, including currency.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Hm, not adding one here. Per `LDG-05` in the design doc only grant, conversion, forfeiture and expiry get an audit row, and conversion already writes `promo.bonus.converted` in the wallet transaction. Stake, win and reversal are per-bet movements, and `promo_grant_entry` is their append-only ledger (trigger-enforced since #189).

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Already covered by #177 - `externalRoundId` is required on `BonusWagerArgs`, and the wallet only calls `wager()` when the bet carries both a `context` and a round id; without them it is a plain real-balance debit with no bonus draw. Done in `9b04420e` - also dropped the now-dead conditional spread here.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Done in `9b04420e` - a `completed` grant still pays to the real balance (its bonus already converted), but `expired` and `forfeited` pay neither side, per question 8's default that a forfeit takes the winnings too. Tests for both states.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Done in `9b04420e` - `settle()` takes a `pg_advisory_xact_lock` on `(userId, providerName, currency, externalRoundId)` before reading the outstanding stake, so a second concurrent callback waits for the first. Concurrent-reversal test included.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Done in `9b04420e` - added `promo_grant_entry.provider_name`, and the round lookup filters on `(userId, currency, providerName, externalRoundId)`. `providerName` comes from `providerRef`, not `context.provider`. Test with two providers sharing one round id.

zaxovaiko:
- `9b04420e` -> `642a0f1c`

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,404 @@
...
+    // A win slice that could not go back onto its grant because the grant is no longer active,
+    // and that grant did not complete (it expired or was forfeited). Question 8's default takes
+    // the winnings from that bonus along with the bonus balance, so this slice belongs to
+    // neither balance - it is not added to realShare below.
+    let forfeited = ZERO;
+    for (const [index, row] of outstanding.entries()) {
+      const last = index === outstanding.length - 1;
+      // Each grant takes the part of each share its own funding paid for; the last takes whatever
+      // truncation left over, so the parts always add back up to the share.
+      const slice = last
```
klaudia-blazyczek-blurify: [BLOCK] The last grant's slice here is `moneySubtract(share, settled)` - the total share minus only what *succeeded* so far. If an earlier (non-last) grant's `creditBonus`/`reverseRound` refuses (its status left `'active'`) but the last grant's own attempt succeeds, the last grant absorbs the refused grant's slice into `settled` (since `settled` never counted it), while `forfeited` (line 227) also counts that same slice. The final `realShare` at line 252 (`applyAmount - settled - forfeited`) then subtracts it twice, can go negative, and `wallet-commands.service.ts:374` feeds that straight into `creditWalletBalance` - a WIN can subtract money from the player's real balance (even below zero) while the `wallet_transaction` row already written records the full positive win.

Fix: track what each row was actually *assigned* separately from what succeeded (`settled`), and use that for the `last` remainder instead of `share - settled`; clamp the returned `realShare` at zero as a backstop.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      const applied = reversal
+        ? await this.reverseRound(tx, row.stake, slice)
+        : await this.creditBonus(tx, row.stake.grantId, slice);
+      if (applied === null) {
+        // A win on a grant that has already completed belongs to the real balance: its bonus
+        // funds converted at completion. A reversal does not - returning a bonus-funded stake as
+        // real money releases it without the wagering it was granted under.
+        if (reversal) {
+          throw new GrantNotReversibleError();
```
klaudia-blazyczek-blurify: [BLOCK] A reversal against a grant that is `completed`, `expired`, or `forfeited` throws `GrantNotReversibleError` unconditionally, aborting the whole `credit()` transaction - including the `wallet_transaction` row already inserted for this callback. For a `completed` grant specifically: the bet finished the requirement, `convertBonus` already turned it into withdrawable cash, and a routine provider void/rollback for that same round can then never settle - it throws on every retry forever, the player's own real-money share of that stake is never returned, and the wagering progress the round bought is never reversed. This inverts the PR's own stated rule ("money back with progress standing is free wagering bought by a rollback").

Fix: reuse the win branch's `grantStatus` check here too - a `completed` grant's bonus slice should fall through to `realShare` (already converted, belongs to the player), and only `expired`/`forfeited` should forfeit it; neither case should throw and abort the transaction.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,404 @@
...
+    // player's own money would pay an un-wagered bonus stake out as withdrawable cash.
+    const outstanding = stakes.map((stake) => ({
+      stake,
+      bonus: reversal ? moneySubtract(stake.bonusStake, stake.bonusReturned) : stake.bonusStake,
+      real: reversal ? moneySubtract(stake.realStake, stake.realReturned) : stake.realStake,
+    }));
+    const totalBonus = outstanding.reduce((sum, row) => moneyAdd(sum, row.bonus), ZERO);
+    const totalStake = outstanding.reduce((sum, row) => moneyAdd(sum, row.real), totalBonus);
+
+    if (moneyCompare(totalBonus, ZERO) <= 0 || moneyCompare(totalStake, ZERO) <= 0) {
```
klaudia-blazyczek-blurify: [BLOCK] This early return fires whenever `totalBonus <= 0`, which includes every round funded 100% from the real balance (the common case for a player who holds an active grant but bets cash). For such a round, a `bet_reversal` (void) returns here without ever calling `reverseRound`, so the wagering progress that bet advanced in `wager()` (which weights and counts the *full* stake regardless of `fromBonus`, per this PR's own "the full stake counts" default) is never taken back. A player can bet real cash, get the round voided by the provider, and keep the progress - repeating this completes and converts a bonus without ever placing an at-risk real-money wager.

Fix: gate this early return on whether the round has any stake row at all (`totalStake <= 0`), not on `totalBonus`, so a pure-cash round's void still reaches the progress-reversal path for any grant that drew wagering credit from it.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,404 @@
...
+        // funds converted at completion. A reversal does not - returning a bonus-funded stake as
+        // real money releases it without the wagering it was granted under.
+        if (reversal) {
+          throw new GrantNotReversibleError();
+        }
+        // creditBonus only refuses once the grant left 'active'. Completed is the good case
+        // above; expired and forfeited took the winnings from that bonus with them too, so the
+        // slice must not fall through to the real balance below.
+        if ((await this.grantStatus(tx, row.stake.grantId)) !== 'completed') {
+          forfeited = moneyAdd(forfeited, slice);
```
klaudia-blazyczek-blurify: [BLOCK] When a win settles against a grant that is `expired` or `forfeited` (not `completed`), the slice is folded into `forfeited` and `continue` skips the `promoGrantEntry` insert below - no row is written anywhere for that money. `wallet-commands.service.ts:341` already wrote the `wallet_transaction` row for the *full* gross win amount before this ever runs, so the gap between what the provider reported and what the player received ends up recorded nowhere: no ledger entry, no audit record.

Fix: insert a `promoGrantEntry` recording the withheld amount before the `continue` (a schema type for this already exists, unused), so every forfeiture is traceable the same way `convertBonus` already makes every conversion traceable.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -153,7 +165,7 @@ export class WalletCommandsService implements WalletCommands {
     const debitCurrency = balanceKey(currency ?? row.currency);
     const debitRow = { ...row, currency: debitCurrency };
 
-    const available = await readWalletBalance(txn, row.id, debitCurrency);
+    const available = await readWalletBalanceForUpdate(txn, row.id, debitCurrency);
```
klaudia-blazyczek-blurify: [WARN] `debit()` locks `wallet_balance` here (`readWalletBalanceForUpdate`) before it locks `promo_grant` (inside `bonusWagering.wager()` a few lines down). `credit()`'s settlement path takes the opposite order: it locks `promo_grant` first (inside `bonusWagering.settle()`), and only afterward locks `wallet_balance` via `creditWalletBalance` (line 374). An ordinary concurrent bet and win/void callback on the same grant/currency can now deadlock, and Postgres aborts one of the two transactions (`40P01`) - a real bet or a real payout spuriously fails and has to be retried.

Fix: make both paths acquire the two locks in the same order, e.g. have `credit()` touch/lock the `wallet_balance` row before calling into `bonusWagering.settle()`.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -178,29 +190,120 @@ export class WalletCommandsService implements WalletCommands {
...
-        userId,
+    // The older fungible rollover model, still running alongside the grant engine until the
+    // chat-gift and rain paths move across. It reads `wallet_balance`, which grant money never
+    // enters, so the two cannot double-count the same bet.
+    const completedBonusCredits =
+      type === 'bet'
+        ? await this.applyBonusRolloverProgress(txn, {
+            userId,
+            currency: debitCurrency,
+            amount,
```
klaudia-blazyczek-blurify: [BLOCK] `applyBonusRolloverProgress` is called with the full stake `amount`, not `realPart` (computed above at line 195 - the portion actually taken from the real balance). A bet entirely or partly funded by the new grant engine's bonus money still advances the *old* rollover-bonus's wagering requirement by the full stake, as if the player wagered that much of their own real balance. A player holding both an old-style gift/rain bonus and a new grant can complete the old bonus's requirement - and convert it to withdrawable cash - using money that came out of the new grant, never their own. The comment above ("the two cannot double-count the same bet") only holds for the balance the two credit, not for the progress they advance.

Fix: pass `realPart` here instead of `amount`.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -248,7 +351,27 @@ export class WalletCommandsService implements WalletCommands {
...
+      (type === 'win' || type === 'bet_reversal')
+        ? await this.bonusWagering.settle(txn, {
+            userId,
+            currency: balanceKey(currency),
+            amount,
+            providerName: providerRef.providerName,
+            externalRoundId: providerRef.externalRoundId,
+            kind: type,
+          })
+        : { realShare: amount };
```
klaudia-blazyczek-blurify: [WARN] Unlike `debit()` (which fails closed at line 215-217 when a bonus-eligible bet is missing `context`/`externalRoundId`), this settlement gate fails open: if `providerRef.externalRoundId` is missing, or a reversal's `providerName` doesn't match what the original bet was recorded under, this falls through to `{ realShare: amount }` and credits the full amount as real money - even for a round that actually drew bonus funds. No in-core caller triggers this today (`gaming.service.ts` always sets both), but the first aggregator adapter whose void callback omits the round id will release an un-wagered bonus stake as cash while the wagering progress it bought stays on the grant.

Fix: when `bonusWagering` is bound and `type === 'bet_reversal'`, refuse rather than default to full real credit when the round id is missing, mirroring the debit-side guard.

packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -0,0 +1,404 @@
...
+      args.externalRoundId,
+    );
+    const reversal = args.kind === 'bet_reversal';
+
+    // A reversal can only give back what the round still holds. A second rollback callback for a
+    // round already returned in full finds nothing outstanding, and reporting its amount as the
+    // player's own money would pay an un-wagered bonus stake out as withdrawable cash.
+    const outstanding = stakes.map((stake) => ({
+      stake,
+      bonus: reversal ? moneySubtract(stake.bonusStake, stake.bonusReturned) : stake.bonusStake,
```
klaudia-blazyczek-blurify: [WARN] For a `win`, `outstanding.bonus`/`outstanding.real` use the round's gross `stake.bonusStake`/`stake.realStake` - only a `bet_reversal` nets against `bonusReturned`/`realReturned`. If a round was already voided (bonus and progress given back) and the provider later reports the same round as a win under a different `providerRefId` (the wallet's duplicate-reference guard only matches the exact same ref, and the advisory lock here serializes but doesn't reject this), the win is split and credited against the original 100% stake as if the earlier reversal never happened - re-funding a round the provider already took back.

Fix: net the win branch against earlier reversals too, the same way the reversal branch nets against itself.

packages/core/src/wallet/service/wallet-commands.service.ts
```diff
@@ -66,6 +77,7 @@ export class WalletCommandsService implements WalletCommands {
     private readonly audit: AuditWritePort,
     private readonly platformConfig?: PlatformConfig,
     private readonly rgLimits?: RgLimitsPort,
+    private readonly bonusWagering?: BonusWageringCommands,
```
klaudia-blazyczek-blurify: [WARN] `WalletCommandsService`'s constructor now takes 5 positional parameters. `docs/standards/functions.md`'s rule ("more than 3 params -> one named object") applies to constructors too, and the PR description acknowledges the violation but defers the fix. Since every call site is touched by this same PR anyway, folding the optional ports (`platformConfig`, `rgLimits`, `bonusWagering`) into one named object is a small, mechanical change worth landing here rather than carrying the debt forward.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Correct, and already how it was. Stake, win and reversal are per-bet movements recorded in promo_grant_entry, which is append-only and trigger-enforced to sum to the grant's bonus_balance since #189. Grant, conversion, forfeiture and expiry are the ones that get an audit row, matching LDG-05. Not adding one here.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Already handled. debit() only calls the engine when both context and providerRef.externalRoundId are present (see the guard right before the wager() call). A bet missing either draws no bonus funds and settles as a plain real debit, refusing on insufficient funds instead.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Already handled. The status check right after creditBonus/reverseRound refuses distinguishes completed (falls through to real, since its bonus already converted at completion) from expired or forfeited (excluded via the forfeited bucket). Covered by the existing "pays neither balance when expired/forfeited" tests.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Already handled. settle() wraps the whole call in an advisory transaction lock keyed on userId/providerName/currency/externalRoundId, so two concurrent settlement callbacks for the same round serialize instead of racing.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Already handled. roundStakes filters on userId, currency, providerName and externalRoundId together, so two providers or currencies minting the same round id can't settle against each other's stake.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Fixed in efc6d6af. Tracked what each row was assigned in a separate `allocated` accumulator, so the last row's remainder is computed against that instead of `settled`, and a failed row can no longer bleed into the next row's slice. Also clamped the returned realShare at zero as a backstop.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Fixed in efc6d6af. Reversal against a grant that's left active no longer throws: a completed grant's slice now falls through to the real balance the same way a win's does, and an expired or forfeited grant's slice is withheld and recorded, never released as cash. Updated the "refuses to return a stake" test to assert the new (correct) outcome, and added a case for the expired/forfeited path.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Fixed in efc6d6af. The early return now fires only when stakes.length is 0 (no grant attributed to the round at all) or the netted totalStake is 0 (already fully reversed), not on totalBonus alone. reverseRound's progress calculation is now proportional to the round's whole stake, bonus and real together, so a cash-funded bet under an active grant reverses the progress it bought. Added a test for this exact scenario.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Fixed in efc6d6af. Inserts a promoGrantEntry (type 'forfeit', bonusAmount 0 to respect the balance-matches-ledger invariant, realAmount carrying the withheld amount) before the continue, for both the win and reversal paths. Added a test asserting the forfeit entry.

packages/core/src/wallet/service/wallet-commands.service.ts
zaxovaiko: Fixed in efc6d6af. debit() and credit() now lock wallet_balance in the same order before either calls into the engine: credit() takes readWalletBalanceForUpdate right before settle(), matching debit()'s existing lock before wager().

packages/core/src/wallet/service/wallet-commands.service.ts
zaxovaiko: Fixed in efc6d6af. applyBonusRolloverProgress now takes realPart instead of the gross stake, so a bet the grant engine partly or fully funded no longer advances the old rollover bonus's requirement by money that wasn't the player's own.

packages/core/src/wallet/service/wallet-commands.service.ts
zaxovaiko: Fixed in efc6d6af. credit() now refuses (ok:false) a bet_reversal missing an external round id whenever bonusWagering is bound, before writing the ledger row, mirroring debit()'s fail-closed guard.

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Fixed in efc6d6af. The netting (bonusStake/realStake minus bonusReturned/realReturned) now applies unconditionally instead of only on the reversal branch, so a win reported after a round was already voided settles against what's actually still outstanding, not the original 100% stake.

packages/core/src/wallet/service/wallet-commands.service.ts
zaxovaiko: Fixed in efc6d6af. Folded platformConfig, rgLimits and bonusWagering into one WalletCommandsPorts object param, so the constructor takes three params instead of five. Updated the call sites and the test constructors that passed rgLimits positionally.

PR #192, opened 2026-09-18
feat(promo): grant expiry, forfeiture and the player's view of a bonus

zaxovaiko:
## Summary

Stacked on #191. Everything that ends a grant - expiry, forfeiture on self-exclusion or account closure - plus the two routes a player reads their bonuses through. BF-592, BF-595 (the read half).

## Why

**Expiry and forfeiture are one code path with two reasons.** Both leave the same shape behind: a terminal status, a zeroed bonus balance, a ledger entry for the funds that died with the grant, and an audit row. Writing them separately would have meant two chances to forget one of those four.

**The claim is the `status = 'active'` predicate on the update, not a read.** A second sweep running beside the first, or an admin forfeiting a grant the sweep is about to expire, updates zero rows and writes nothing - rather than a second ledger entry against a balance that is already gone. Each grant closes in its own transaction: one transaction for a batch of five hundred would hold locks across the whole sweep for no benefit.

**Self-exclusion and account closure forfeit immediately**, not at the next cron tick. A bonus is money a player may not keep once they have excluded themselves, and a five-minute window where they still can is not what the rule says. Both subscribe to events that already exist (`rg.self_exclusion.activated`, `player.account.closed`).

**The forfeited amount is read off the grant's own ledger**, not from the column the update just zeroed. The entries sum to the balance by construction, so what the grant still held is what the entries say it held - and if that identity is ever broken, this is one of the places it shows up rather than being papered over.

**The player routes return a grant as its holder sees it.** The `terms` snapshot stays internal: it carries the operator's weighting, which is not the player's business. A grant belonging to someone else is **404, not 403** - a 403 confirms the id exists.

## Why these land together

The cron that drives the sweep is scheduled from a router factory, which is the only place in a plugin with container access after boot. An expiry sweep without a router would have meant mounting an empty router to hold the schedule.

**A win on a grant this branch could now expire or forfeit was paid out as withdrawable cash.** Settlement treated "the grant refused this" as "this round drew no bonus", which credits the full amount to the real balance. That was correct while `completed` was the only closed state - those funds had already converted - and wrong the moment a grant could be expired or forfeited, where the funds were destroyed. A player could stake a bonus on a slow-settling round, self-exclude, and collect the win as cash with none of the wagering done. Settlement now asks *why* the grant refused and destroys the bonus share when the grant's own balance was destroyed. The same change unsticks the void path, which used to throw on a closed grant and leave a provider retrying a rollback forever.

**Forfeiture moved onto a durable job.** It moves money, and a best-effort event handler that died between two of a player's grants left the rest active with nothing to retry them - the sweep only looks at expiry dates, never at exclusion state. The `status = 'active'` claim already made the work idempotent, so the retry costs nothing.

**The actor is carried through.** Both subscribed events name who acted; an admin closing an account and a player excluding themselves were both being recorded as `system`.

**The forfeited amount is read under the row lock** the update is about to take, so it is the column the CHECK constraint protects rather than a sum derived from the ledger.

Also from the quality pass: event payloads are validated with `domainEventSchemas[...].safeParse` instead of a cast, `GET /promo/grants` paginates like every other list route in this repo, and a dead conditional type and its suppressive cast are gone.

## Alternatives considered

**A single transaction per sweep batch.** Simpler to write, holds row locks across up to five hundred grants, and one failure rolls back the whole batch. Per-grant transactions mean a bad row costs one grant.

**Reversing a completed grant's conversion on forfeit.** Out of scope here: a grant that completed converted its funds to real money, and clawing that back is a different decision with a different audit story.

**Deferring exclusion forfeiture to the sweep.** Would have been less code and a window in which an excluded player still holds spendable bonus funds.

## Risks

A grant that completed or expired before a forfeit arrives is left alone - the forfeit finds nothing active and writes nothing. That is the intended reading of "forfeit what they still hold", but it means a forfeit issued seconds after a conversion takes nothing.

The sweep takes five hundred grants per tick at a five-minute cadence. A backlog larger than six thousand an hour would grow; the cadence is a constant and the batch is ordered by expiry, so the oldest always move first.

This PR also drops the promo value sets that the module was still declaring alongside the copies in the isomorphic contracts zone. Same values, one declaration.


packages/core/src/promo/bonus/plugin.ts
```diff
@@ -1,12 +1,50 @@
...
   WAGER_TRACKING,
+  domainEventSchemas,
+  queue,
+  type JobQueueAdapter,
 } from '@openora/core/contracts';
+import {
+  DRIZZLE,
+  EVENT_BUS,
+  createLogger,
+  type CoreTokenCatalog,
```
klaudia-blazyczek-blurify, line 19: A self-exclusion provides the player ID as actorId, and this publishes it on promo.bonus.forfeited. That event defines actorId as the admin actor and requires null when the player or a rule caused the forfeit. Publish null for a self-exclusion so audit and notification consumers do not attribute it to an admin.

packages/core/src/promo/bonus/plugin.ts
```diff
@@ -1,12 +1,50 @@
...
+const EXPIRY_QUEUE = queue('promo-bonus-expiry');
+const FORFEIT_QUEUE = queue('promo-bonus-forfeit');
+const EXPIRY_CRON = '*/5 * * * *';
+
+const EmptyJobPayloadSchema = z.object({});
+
+/**
+ * Taking a player's bonuses away is a money movement, so it runs as a durable job rather than
+ * inside a best-effort event handler: a process that dies between two of a player's grants has
+ * to resume, and the `status = 'active'` claim makes the retry harmless.
```
klaudia-blazyczek-blurify, line 40: The forfeit worker is enqueued with the driver default of one attempt. A transient database failure after some grants close ends the job and leaves the remaining active grants without another forfeit. Configure retry and backoff; the active-status claim is already safe for redelivery.

packages/core/src/promo/bonus/plugin.ts
```diff
@@ -17,5 +55,112 @@ export default {
...
+          grantId: grant.grantId,
+          currency: grant.currency,
+          forfeitedAmount: grant.forfeitedAmount,
+          ...(reason === undefined ? {} : { reason, actorId: grant.actorId }),
+        });
+      }
+    };
+
+    ctx.jobs.worker({
+      queue: EXPIRY_QUEUE,
```
klaudia-blazyczek-blurify, line 80: RgInitiator permits system, but this condition treats every non-player self-exclusion as an admin action. A system-triggered exclusion therefore writes a regulator-facing audit record with actorType admin. Map system to its actual actor type instead.

packages/core/src/promo/bonus/plugin.ts
```diff
@@ -17,5 +55,112 @@ export default {
...
+      schema: EmptyJobPayloadSchema,
+      handler: async () => {
+        if (!lifecycle) {
+          logger.warn('promo-bonus-expiry sweep skipped - service not constructed');
+          return;
+        }
+        const closed = await lifecycle.expireDue();
+        announce('promo.bonus.expired', closed);
```
klaudia-blazyczek-blurify, line 88: This is the only handoff from exclusion or account closure to a money job, but it detaches enqueue and only logs a failure. A queue outage acknowledges the source event and leaves all active grants in place with no durable retry. Persist the work atomically with the source state change, or use an outbox.

packages/core/src/promo/bonus/service/grant-reader.service.ts
```diff
@@ -0,0 +1,88 @@
...
+  async list(
+    userId: Uuid,
+    query: PageQuery & { status?: PlayerGrant['status'] },
+  ): Promise<Paginated<PlayerGrant>> {
+    const { page, limit } = query;
+    const where =
+      query.status === undefined
+        ? eq(promoGrant.userId, userId)
+        : and(eq(promoGrant.userId, userId), eq(promoGrant.status, query.status));
+    const [rows, [total]] = await Promise.all([
```
klaudia-blazyczek-blurify, line 50: This query filters by userId and sorts by createdAt and id, but there is no index for that access pattern. The active-grant index starts with user_id, currency, expires_at and cannot serve terminal history, so the player history scans and sorts as it grows. Add an index on user_id, created_at, id.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Done in `1f4cf6c8` - `actorId` on `promo.bonus.forfeited` is now null unless the actor is an admin. The audit row still names the player, the event no longer does.

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Done in `1f4cf6c8` - the forfeit enqueue carries `attempts: 5` with exponential backoff, same as the other durable jobs.

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Done in `1f4cf6c8` - `initiatedBy` maps all three `RgInitiator` values to the real actor type, so a system exclusion records as `actorType: 'system'`.

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Hm, not doing this one here. Detached enqueue from an event handler is the existing convention (`enqueueEval` in `compliance/plugin.ts` does the same), and there is no event-to-job outbox anywhere yet - that is a platform capability, not a bonus fix. `1f4cf6c8` adds retries with backoff, so the remaining gap is a sustained queue outage, same as every other event-driven job. Maybe worth a separate ticket.

packages/core/src/promo/bonus/service/grant-reader.service.ts
zaxovaiko: Done in `b78f4a15` - added `promo_grant_user_id_created_at_idx` on `(user_id, created_at, id)`.

zaxovaiko:
- `1f4cf6c8` -> `f963edbc`
- `b78f4a15` -> `bc6bd8f7`

PR #193, opened 2026-09-18
feat(promo): deposit offers, eligibility and opt-in

zaxovaiko:
## Summary

Stacked on #192. `promo_offer` and `promo_opt_in`, the eligibility rules, the deposit that turns an opt-in into a bonus, and the admin CRUD behind it. BF-589, and the deposit-offer half of BF-594.

## Why

**An offer is one mutable row, not a versioned one.** A grant snapshots the terms it was created under, which already makes editing an offer unable to reach a bonus a player holds. A `version` column with a bump-on-edit rule, or a `promo_offer_version` table, would be a second mechanism enforcing a property the snapshot enforces by construction - and two mechanisms for one rule is how they drift. `status` covers draft, active, paused and archived.

**Eligibility is a pure function, and it is the same function on both sides.** "Which offers can this player see" and "does this deposit earn this bonus" are the same question asked with different facts. Two implementations would eventually disagree, and the one that granted would be the laxer one. It takes the facts from its caller and returns *which rule* closed the offer rather than a boolean, so the opt-in route can say why.

Absent facts fail the rule that needs them. An offer restricted to first depositors is not open to a caller who cannot say whether this is a first deposit - the alternative is a first-deposit-only offer granting on every deposit the moment a caller forgets to pass the flag.

**Deposits accumulate on the opt-in.** A player who deposits half the minimum twice qualifies; that is the progress bar the design asks for, and it is one column rather than a second table. The opt-in row is both the claim and the progress, and its unique index on `(user_id, offer_id)` is what makes taking an offer twice the same claim.

**The grant is created by a durable job**, keyed on the deposit's transaction id, not inside the event handler. A bonus a player was promised and did not receive because an event was dropped is the same class of defect as a forfeit that never ran - and the grant's own `(user, source, source_ref)` index means a retried job resolves to the grant it already made.

**`match_percent` is `numeric(5,2)`**, like a weight. A percentage was never money and does not need eighteen decimal places. The grant amount is the match, truncating, capped at the offer's ceiling - truncation favours the operator on the fraction and the cap is exact.

Six ways a deposit could pay more bonus than it earned, or none at all:

- **The accumulator had no durable guard.** Jobs are at-least-once and a queue idempotency key is not a correctness guard, so a redelivered deposit added itself to the running total twice - turning a below-minimum deposit into a full-cap bonus. `promo_opt_in_deposit` now records which deposits were counted toward which claim, and its unique index makes the second delivery a no-op.
- **The grant's source reference was the deposit id alone**, shared across every offer a player holds. One deposit satisfying two claims collided on the grant's idempotency index: either the amounts differed and the job threw and retried forever, or they matched and the second claim was silently consumed against the first offer's grant. It is per deposit *and* offer now.
- **A refusal banked the deposit whatever the reason.** Deposits made while an offer was paused or out of window accumulated, so re-activating it paid a match on the whole history. Only "short of the minimum" counts toward it now - the one refusal that means "not yet".
- A grant the port refuses no longer banks the total that would have earned it, and logs rather than failing silently.
- A match that truncates to zero skips instead of throwing and stranding every other claim in the batch.
- The row lock names the opt-in; locking the joined offer rows deadlocked unrelated players' deposits over a shared offer.

**`excludedCountries` is removed rather than shipped.** No port exposes a player's residence country, so the rule could never fire - and a jurisdictional exclusion an operator configures and a regulator asks about must not be one that silently does nothing. `firstDepositOnly` stays and now works: the wallet already answers lifetime deposits, and a total equal to this deposit means there was nothing before it.

Also: re-denominating an offer with live claims is refused, taking a claim is audited, and `requiresOptIn: false` now opens the claim itself instead of being a flag nothing read.

## Alternatives considered

**Calling a command port from inside the deposit transaction** instead of subscribing. The wallet publishes `wallet.deposit.completed` after its own commit, so there is no caller transaction left to join; a synchronous seam would mean a new port on the wallet's deposit path for one consumer. The durable job gets the same guarantee.

**A separate archive route.** Archiving is a status. A second endpoint writing one column is a second thing to keep in step with the first.

**`minRank` eligibility, and free spins.** Both are named in the ticket. The rank table belongs to the gamification module and does not exist yet, so the predicate would be written against a guess; free spins are unspecified beyond the number 50. Neither is in this PR, and `rules` is jsonb so neither costs a migration when it arrives.

## Risks

`applyDeposit` locks every un-granted opt-in a player holds (`FOR UPDATE`) for the duration of the job. A player with many open opt-ins serialises their own deposits against each other; nobody else is affected.

The offer list is filtered in the service rather than in SQL, because the eligibility predicate is pure and takes facts the query does not have. At the scale an operator runs offers - tens, not thousands - that is the right trade; if it ever is not, the window and status filters are already columns.

A player's residence country is not available to this module, so an offer cannot be restricted by jurisdiction yet. The rule is absent rather than present-and-inert; it returns when a port answers the question.

Lowering an offer's `minDeposit` instantly qualifies every banked accumulator across all players at once. That is arguably what an operator means by lowering it, but it is worth knowing before doing it.


packages/core/src/promo/bonus/plugin.ts
```diff
@@ -5,31 +5,38 @@ import {
...
   type BonusForfeitReason,
+  CurrencyTickerSchema,
   JOB_QUEUE,
+  MoneyAmountSchema,
   UuidSchema,
   WAGER_TRACKING,
+  WALLET_READER,
   domainEventSchemas,
   queue,
   type JobQueueAdapter,
```
klaudia-blazyczek-blurify, line 16: This job can create an active bonus after self-exclusion or account closure. If the forfeit job runs before this delayed job, it sees no grant; this job then creates one and no new forfeit is triggered. Reject excluded or closed players inside this transaction, or coordinate both jobs durably.

packages/core/src/promo/bonus/plugin.ts
```diff
@@ -5,31 +5,38 @@ import {
...
+import { OfferService } from './service/offer.service.js';
 import { WageringService } from './service/wagering.service.js';
 import { createBonusRouter } from './router/index.js';
 
 const logger = createLogger('promo-bonus');
 
 const EXPIRY_QUEUE = queue('promo-bonus-expiry');
 const FORFEIT_QUEUE = queue('promo-bonus-forfeit');
+const DEPOSIT_QUEUE = queue('promo-offer-deposit');
 const EXPIRY_CRON = '*/5 * * * *';
```
klaudia-blazyczek-blurify, line 41: This is the only handoff from a completed deposit to the money job, but the event and this enqueue are both best effort. A broker or Redis outage only logs the error and loses the promised grant. Persist a work item with the deposit transaction through an outbox and consume it from there.

packages/core/src/promo/bonus/plugin.ts
```diff
@@ -46,6 +53,19 @@ const ForfeitJobSchema = z.object({
   actorIsAdmin: z.boolean(),
 });
```
klaudia-blazyczek-blurify, line 54: The deposit money job uses the driver default of one attempt. A temporary database or dependency error permanently leaves an earned bonus ungranted. Set bounded retries and backoff with a dead-letter alert; the database guards already make retries safe.

packages/core/src/promo/bonus/service/offer.service.ts
```diff
@@ -0,0 +1,397 @@
...
+    });
+  }
+
+  async update(adminId: Uuid, input: UpdatePromoOfferInput): Promise<PromoOffer> {
+    const { id, ...patch } = input;
+    return this.drizzle.db.transaction(async (tx) => {
+      const before = await this.requireOffer(tx, id);
+      // Re-denominating an offer players have already banked deposits against would pay a grant
+      // in one currency backed by deposits made in another.
+      if (patch.currency !== undefined && patch.currency !== before.currency) {
```
klaudia-blazyczek-blurify, line 124: This read is not protected against a concurrent claim. An admin can see no opt-in, a player can create one using the old currency, then this update changes the offer currency. Later deposits are skipped because the claim and offer currencies no longer match. Lock or serialize offer changes and claim creation.

packages/core/src/promo/bonus/service/offer.service.ts
```diff
@@ -0,0 +1,397 @@
...
+  async applyDeposit(
+    tx: DrizzleTx,
+    deposit: { userId: Uuid; amount: string; currency: string; transactionId: Uuid },
+    context: OfferContext = {},
+  ): Promise<GrantedBonus[]> {
+    // The deposit has already committed by the time this job runs, so a lifetime total equal to
+    // this deposit means there was nothing before it. Asking the wallet keeps the fact where it
+    // is owned rather than snapshotting it onto an event.
+    const lifetime = await this.wallet.getLifetimeDeposit(deposit.userId);
+    const isFirstDeposit = moneyCompare(lifetime, deposit.amount) === 0;
```
klaudia-blazyczek-blurify, line 235: First-deposit eligibility is calculated from the lifetime total when this asynchronous job runs. If a second deposit commits before the first job runs, the total includes both deposits and both jobs reject a first-deposit-only offer. Use an immutable fact or ordering from the wallet transaction.

packages/core/src/promo/bonus/service/offer.service.ts
```diff
@@ -0,0 +1,397 @@
...
+  /**
+   * The claims this deposit could satisfy: the ones the player took, plus the offers that need no
+   * taking. An offer marked `requiresOptIn: false` applies to any qualifying deposit, so the claim
+   * is opened here rather than requiring the player to ask for something already theirs.
+   */
+  private async claimsFor(
+    tx: DrizzleTx,
+    deposit: { userId: Uuid; currency: string },
+  ): Promise<{ optIn: PromoOptInRow; offer: PromoOfferRow }[]> {
+    const automatic = await tx
```
klaudia-blazyczek-blurify, line 330: This creates automatic claims for every active offer before filtering by the deposit currency or eligibility. An unrelated-currency deposit can leave a claim that later blocks re-denomination although it could never earn a bonus. Filter candidates first, or only treat banked or granted claims as blockers.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Done in `2c7ceb45` - `promo/bonus` now requires `PLAY_ELIGIBILITY`, `optIn()` refuses a restricted player, and `applyDeposit()` re-checks `isRestricted` at credit time and skips the grant (the deposit still banks its accumulator). A delayed deposit job running after the forfeit sweep grants nothing now.

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Hm, leaving this one. It is the same best-effort event to job shape the forfeit path ships, and the job has its own DB guard. The durable version already exists as a platform switch - `OUTBOX_ENABLED` (or a broker) turns `emit()` into a transactional outbox per ADR-0017, so that is a deployment decision rather than code here.

packages/core/src/promo/bonus/plugin.ts
zaxovaiko: Done in `ef1d6735` - the deposit enqueue reuses the forfeit retry policy (renamed to `MONEY_JOB_RETRY` since it is shared now).

packages/core/src/promo/bonus/service/offer.service.ts
zaxovaiko: Done in `ef1d6735` - `requireOffer` takes the offer row `FOR UPDATE`, so the claim check in `update()` and a concurrent `optIn()` serialize on it.

packages/core/src/promo/bonus/service/offer.service.ts
zaxovaiko: Done in `ef1d6735` - added `WalletReader.isFirstDeposit(userId, transactionId)`, which compares the deposit's own committed `created_at` instead of a lifetime total, so out-of-order job processing cannot flip it.

packages/core/src/promo/bonus/service/offer.service.ts
zaxovaiko: Done in `ef1d6735` - the automatic-claim query in `claimsFor` is scoped to the deposit's currency, so a deposit never opens a claim on an offer it could not satisfy.

zaxovaiko:
- `ef1d6735` -> `0ecbab36`
- `2c7ceb45` -> `094cea37`

PR #194, opened 2026-09-18
feat(promo): admin bonus visibility and manual forfeit

zaxovaiko:
## Summary

Stacked on #193. A support manager can see a player's bonuses and take one away with a reason on the record. BF-595's API half.

## Why

**The admin view is the player's view plus the source.** Which deposit, which rain drop, which offer produced the bonus - a player has no need for that and does not get it, and support cannot answer "why do I have this bonus" without it. The terms snapshot still stays internal on both surfaces; it is the operator's weighting, not an answer to a dispute.

**The reason and the note are mandatory at the contract, not in the handler.** A handler check is one refactor away from being skipped, and a forfeiture with no recorded reason is exactly what a regulator asks about. The note has a minimum length and is trimmed, so "." does not satisfy it.

**Forfeiting a bonus that is already gone refuses rather than reporting success.** The claim-then-act update finds nothing to claim and the route returns 409. An admin who forfeits a bonus that converted an hour ago acted on stale information and needs to be told - a silent no-op leaves them believing they took money back that the player already has.

- **The admin forfeit emits `promo.bonus.forfeited`.** It was the only forfeit path that destroyed bonus funds silently - the sweep emits, self-exclusion emits - so every consumer of that event under-reported the hand-issued forfeits, which are the ones most likely to be disputed.
- **A missing grant answers 404, not 409.** The 409 asserted the bonus existed and was already closed, so a typo'd id read as a completed action and put a false statement in the dispute trail. The read-back's own not-found error is mapped too, instead of surfacing as a 500 *after* the forfeit had committed.
- **A refused forfeit leaves an audit row.** A money resource an admin can probe without trace is exactly where the outcome matters; only successful attempts were recorded before.
- **`pending` counts as live.** Nothing writes that status today, but this PR is what makes forfeiting admin-visible, and a pending grant that could be neither forfeited nor expired would sit there and pay out later.

## Alternatives considered

**Re-using the player grant route with an admin-visible flag.** Two audiences, two shapes, and a flag deciding which fields come back is the kind of thing that leaks the wider shape to the narrower audience the first time someone edits it.

**Letting the forfeit route return the grant it took as the service saw it.** It re-reads instead. The lifecycle service returns what it closed for the caller's own bookkeeping, not a DTO, and shaping that into the admin view inside the service would put a wire concern in the place that moves money.

## Risks

Forfeiting is per grant. Taking every bonus a player holds is several calls, and nothing in the API expresses "all of them" - the sweep does that on self-exclusion and account closure, where it is a rule rather than a decision.

The route reads the grant back after the forfeit, so the response is a second query rather than the row the update returned. Consistent within the transaction that just committed, but a grant forfeited and then immediately expired by the sweep would report whichever landed last.


packages/core/src/promo/bonus/drizzle/migrations/0006_square_argent.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX "promo_grant_expires_at_idx";--> statement-breakpoint
```
klaudia-blazyczek-blurify, line 1: This drops the index before its replacement exists. During a rolling deploy, the previous release can run expiry queries while no supporting index exists, and the drop takes an exclusive lock. Create a new index under a new name, deploy it, and remove the old index later.

packages/core/src/promo/bonus/service/grant-lifecycle.service.ts
```diff
@@ -38,7 +53,7 @@ export class GrantLifecycleService {
     const due = await this.drizzle.db
       .select({ id: promoGrant.id })
       .from(promoGrant)
```
klaudia-blazyczek-blurify, line 55: exists.status is stale if expiry, conversion, or another forfeit closes the grant between the earlier read and close. The API returns 409, but its immutable audit record can say active even when the grant is expired or completed. Read the current status after the failed claim, or return it from the transaction.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/router/index.ts
```diff
@@ -33,6 +48,37 @@ export function createBonusRouter({
...
+      grants: {
+        list: os.admin.grants.list.handler(async ({ input, context }) => {
+          await adminGuard.assert(context, 'bonus', 'view');
+          return grants.listForAdmin(input.userId, input);
+        }),
+
+        forfeit: os.admin.grants.forfeit.handler(async ({ input, context }) => {
+          const { userId } = await adminGuard.assert(context, 'bonus', 'cancel');
+          return mapErrors(
+            { CONFLICT: GrantNotForfeitableError, NOT_FOUND: GrantNotForfeitableNotFoundError },
```
marek-chmielowski-blurify, line 60: The forfeit handler's `mapErrors` only maps the lifecycle service's `GrantNotFoundError`, but the post-commit `grants.getForAdmin(input.id)` call can throw grant-reader.service.ts's own distinct `GrantNotFoundError` class (both built via separate `makeNotFoundError('Grant')` calls, so they aren't the same identifier), which isn't included in this map. If `grants.getForAdmin` ever throws its own `GrantNotFoundError`, `mapErrors` falls through to `throw err` unmapped, surfacing as a 500 to the admin even though the forfeit already committed — the exact failure mode this PR is meant to eliminate.

packages/core/src/promo/bonus/service/grant-reader.service.ts
```diff
@@ -34,6 +35,14 @@ const COLUMNS = {
   createdAt: promoGrant.createdAt,
 };
 
+const ADMIN_COLUMNS = {
+  ...COLUMNS,
+  userId: promoGrant.userId,
+  offerId: promoGrant.offerId,
+  source: promoGrant.source,
```
marek-chmielowski-blurify, line 42: `ADMIN_COLUMNS` redundantly redeclares `source: promoGrant.source`, which is already inherited from `...COLUMNS` above. No functional bug (same value overwrites itself), but it's dead/duplicate code that could mislead a reader into thinking `source` is admin-only when it's already part of the player-facing `COLUMNS` — worth removing.

packages/core/src/promo/bonus/drizzle/migrations/0007_fresh_squirrel_girl.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX "promo_grant_expires_at_idx";--> statement-breakpoint
```
zaxovaiko: Hm, not needed here - `promo/bonus` does not exist on `dev` yet, so migrations `0000` to `0007` all land together on the first merge of this stack. There is no previous release running expiry queries against the old index for the drop to race. Worth doing this way for a live table, though.

packages/core/src/promo/bonus/service/grant-lifecycle.service.ts
zaxovaiko: Done in `ec25b000` - `forfeit()` re-reads `promo_grant.status` after a failed `close()` instead of reusing the pre-check `exists.status`, so `forfeit_refused` records what the grant actually was. Added an E2E racing two admin forfeits on one grant: one wins, one ledger entry, and the refusal reads `forfeited`.

packages/core/src/promo/bonus/router/index.ts
zaxovaiko: Done in `ec25b000` - `NOT_FOUND` in the forfeit route's `mapErrors` now names both `GrantNotForfeitableNotFoundError` and the reader's `GrantNotFoundError`, so a miss on the read-back is a 404, not a 500.

packages/core/src/promo/bonus/service/grant-reader.service.ts
zaxovaiko: Done in `ec25b000` - dropped the redundant `source` in `ADMIN_COLUMNS`, it already comes from `...COLUMNS`.

zaxovaiko:
- `ec25b000` -> `463ea275`

PR #195, opened 2026-09-18
feat(gaming): extend admin game list filters

okapitula:
## Summary

Adds category (all-of, plus `uncategorized`), tag (all-of), game type (any-of) and geo-block filters to `GET /backoffice/gaming/games`. Stacked on #173: the geo filters read its `provider_geo_rule` table, so this merges after it.

## Why

The admin game list could only filter by search, provider, a single category, `isActive` and `isUnavailable`. Operators need to find games by category membership (including games in no category), by tag combinations, by type, and by geo-block status.

- `categoryIds[]`, `tagIds[]`, `geoBlockedCountries[]` are all-of: a game must match every listed value. `gameTypes[]` is any-of, since a game has exactly one type.
- A game counts as geo-blocked in a country by its own rule or its provider's, matching the play gate in #173. Country-wide blocks are not included.
- Geo rules are compliance data, so the geo filters require `compliance:view` on top of `game-config:view`, the same grant compliance's own geo-rule routes use.
- The geo filters return 400 when the compliance module is not loaded, and contradictory combinations (`uncategorized=true` with a category filter, `geoBlocked=false` with countries) return 400.
- Lists accept bracket notation (`?tagIds[]=a&tagIds[]=b`) or a single bare value; duplicates are dropped before the per-list cap applies.

## Alternatives considered

- Revenue and plays range filters: dropped for now. They aggregate `game_round`, need a date window and a currency to be meaningful, and fit the existing game performance report better than the catalog list.
- A compliance-owned read port for geo rules instead of reading its `/schema` subpath: the port would return unbounded id lists to feed back into the paginated query; the `/schema` read is the sanctioned pattern and compliance already reads gaming's tables the same way.

## Risks

- Merge order: depends on #173 (`provider_geo_rule`). Retarget to `dev` once #173 merges.
- "Compliance loaded" is detected from the `GAME_GEO_CHECK` binding. An overlay that binds its own geo check without the compliance module would get a 500 instead of a 400 on the geo filters.
- Additive contract change only; the public `GET /gaming/games` is unchanged.


zaxovaiko (dismissed):

packages/core/src/casino/gaming/service/gaming.service.ts
```diff
@@ -188,15 +213,112 @@ export class GamingService {
...
+      table: providerGeoRule,
+      column: providerGeoRule.providerId,
+      equals: game.providerId,
+    });
     return this.listGames({
       ...input,
       playableOnly: false,
       sort: 'admin',
       includeInvisibleTags: true,
+      filters: [
```
zaxovaiko, line 250: Checked the new filters on a seeded DB (30k and 300k games) and indexes look fine, category, tag and geo country filters all hit the existing indexes on the link and geo rule tables.

The one outlier is `geoBlocked`. The `or(exists(...), exists(...))` pair takes ~255ms at 300k games (~350ms for `false`), while the same check as `inArray(game.id, db.select({ gameId: gameGeoRule.gameId }).from(gameGeoRule))` plus `inArray(game.providerId, ...)` for `providerGeoRule` takes ~37ms (~64ms for `false`). Let's switch to `inArray`/`notInArray` here? Both columns are `notNull`, so `not in` is safe.

zaxovaiko:
Not from this PR, but while checking the filters I noticed `game.name` has no index, and the public `listGames` sorts by it. So the lobby's first page and every category page sort all active games on each request: 12.7ms at 30k games and ~40ms at 300k. With `index('game_name_idx').on(t.name)` it goes down to ~0.3ms, and the category filter drops from ~37ms to ~3ms at 300k.

Let's add it to the `game` table in `gaming/schema/index.ts`? It's one index plus a generated migration, so it can land here or in a small separate PR.

packages/core/src/casino/gaming/service/gaming.service.ts
okapitula: Reproduced and switched, with one tweak: the game-rule subquery is `group by game_id`, and `false` stays `notExists`. The plain `IN`/`NOT IN` forms stop finishing once the rule table grows.

300k games, page query:

| | 24k rules | 250k rules |
|---|---|---|
| true: `exists OR exists` (before) | 435 ms | 1100 ms |
| true: `IN OR IN` | 58 ms | timeout >60s |
| true: `IN (group by) OR IN` (now) | 59 ms | 109 ms |
| false: `NOT IN AND NOT IN` | 86 ms | timeout >60s |
| false: `not exists AND not exists` (kept) | 87 ms | 168 ms |

okapitula:
Confirmed, added here with the migration. 300k games:

| | before | after |
|---|---|---|
| public first page | 64 ms | 0.3 ms |
| category page | 60 ms | ~2-20 ms |

The parallel `count(*)` is now the floor at ~30ms. Also tried `game_provider(name, slug)` for the admin sort: first page 78 → 3ms, but a no-match `q` search 182 → 589ms, so left it out.

zaxovaiko (approved):

PR #197, opened 2026-09-18
fix(cms): guard default banner unset

damianrzepka:
## Summary

Prevents a placement default from being unset while an active or future banner schedule depends on it. Serializes the default-unset and schedule-create operations with the placement schedule lock.

## Why

A concurrent schedule creation could otherwise succeed after the default had been removed, leaving the schedule without its fallback banner.

## Alternatives considered

A client-side check cannot make this invariant safe because another administrator can change schedules after the check. The service-level transactional guard is the shared enforcement point.

## Risks

The migration adds an index on schedule end time. Deploy migrations before application instances that use the guard.


packages/core/src/cms/service/cms.service.ts
```diff
@@ -636,25 +640,45 @@ export class CmsService {
...
+          .select({ id: bannerScheduleTable.id })
+          .from(bannerScheduleTable)
+          .innerJoin(
+            bannerConfigurationTable,
+            eq(bannerScheduleTable.bannerConfigurationId, bannerConfigurationTable.id),
+          )
+          .where(
+            and(
+              eq(bannerConfigurationTable.placement, placement),
+              gt(bannerScheduleTable.endsAt, new Date()),
```
jakubfilinger-b, line 659: This guard can still be bypassed by reviving an expired schedule: once endsAt is not later than now, the default can be unset, then updateBannerScheduleEnd can move that schedule end date back into the future because it only checks that the end is after the start. The placement then has a queued or active schedule without a fallback default. Preserve the default invariant when extending a schedule into the future, or prohibit reviving an expired schedule, and add coverage for expire, unset default, then extend end date.

jakubfilinger-b (changes requested):
Blocker: an expired schedule can be revived after the default is cleared, so the end-date update path must preserve the fallback invariant.

packages/core/src/cms/service/cms.service.ts
damianrzepka: Fixed in 8e118ce1. Expired schedules now return a conflict when an end-date update would move them back into the future, preserving the placement fallback invariant. Added integration and HTTP E2E coverage for expiry, default unset, then attempted resumption.

jakubfilinger-b (approved):

PR #202, opened 2026-09-20
feat(promo): bonus limits, the weight admin surface and a player's own view

zaxovaiko:
## Summary

Stacked on #194. The pieces the engine stack left to configuration, plus the two reads a player needs to see what they hold, plus the realtime signal behind them. BF-592, BF-594.

## Why

**Limits the terms carry.** `maxBet` and `maxWinMultiplier` join the grant terms and are snapshotted with the rest, so editing an offer never reaches a bonus a player already holds.

`maxBet` is enforced inside the debit transaction against that snapshot rather than by a caller beforehand: it exists so a player cannot turn a wagering requirement into one high-variance spin, and a preflight check is one client away from being skipped. A stake over it raises `WalletMaxBetExceededError` rather than reporting a short balance - telling the player to deposit more would send them into the same wall with a bigger balance.

`maxWinMultiplier` bites at conversion, the single point where bonus funds become real money, rather than on each win along the way. What the operator caps is the payout, and winnings that are still bonus can still be lost back. What the cap takes leaves the ledger as its own `forfeit` entry, so a grant's entries keep summing to the balance they explain.

Both were decided rather than assumed: the maximum payout from bonus winnings is an admin setting per the payments and bonuses consultation, and a per-bonus maximum bet is a business rule on the engine's own specification page.

**Sportsbook and PvP are seeded at zero.** Weight resolution falls through to the profile default, so without those rows the headline rule - only casino play clears a requirement - was not in effect on a fresh install. A PvP bet counted in full.

**The weight admin surface.** `promo_weight_profile` and `promo_weight` were configurable only by hand in SQL, and `promo.weight_profile.set` was an audit action nothing could write. An operator could not create a profile, let alone exclude a product. `set` replaces the whole row set instead of merging: a half-saved profile scores bets against whatever happened to land.

**What a player can see.** `GET /promo/balance` sums their live grants per currency - there is no balance table, the grants are where the funds are. `GET /promo/grants/{id}/entries` is the only place bonus-only movements appear at all, since a stake paid from bonus funds and a win attributed back to the grant never reach the wallet ledger, and the criterion is that every credit, debit, conversion and forfeiture is visible in the player's history. Someone else's grant is 404, not 403.

**Realtime.** `GET /promo/balance/stream` publishes a change *signal*, never the figures, exactly like the wallet balance stream: a dropped or reordered message would leave a stale amount on screen with no self-correction, and the client refetches instead. It is subscribed centrally to the four topics rather than published beside each emit, so a fifth producer gets the signal without having to remember it.

Wagering progress moving on the player's own bet is deliberately not on the stream. The debit response already carries `bonusSpent` and `bonusBalance`, so the client that placed the bet already knows; the stream is for the changes a client cannot see coming - a deposit's bonus arriving from a job, the expiry sweep, an admin forfeiting, a requirement completing.

## A defect this found in itself

`GrantService` validated terms against its own inline Zod schema rather than the contract's, and Zod strips what it does not declare - so both new limits were dropped silently on the way to the snapshot the engine enforces against. Types compiled, the field existed on the port and the contract, and the limit did nothing.

That is the same shape as two other defects in this stack: one declaration of a thing, then a second one beside it. The test was written before the code was believed, which is the only reason it surfaced.

## Alternatives considered

**Capping the win rather than the conversion.** Rejected: it would mean deciding, on every credit, whether this win is the one that crosses the ceiling, and a player whose bonus balance later falls back below it would have been capped for nothing. Conversion is one site and one decision.

**Per-row CRUD on weights.** A profile is saved as a set or not at all. Per-row editing lets an operator save half of one, and a bet resolving against a half-saved profile scores at whatever is there.

**Publishing the balance on the stream.** The wallet stream already documents why not, and a bonus position is the same argument with more moving parts.

## Risks

`GET /promo/balance` sums per currency on every call. At the number of live grants one player holds that is a covered index read; it is not a leaderboard.

The cap kills the surplus at conversion with a ledger entry and no separate event, so a consumer wanting to tell a player "your win was capped" has to read the entries. Worth an event if the UI asks for it.

`maxBet` bounds the whole stake rather than the bonus part, which is what "maximum bet while a bonus is active" means - a player holding an active grant is limited even on a bet their real balance covers entirely.


packages/core/src/promo/bonus/service/wagering.service.ts
```diff
@@ -26,6 +26,23 @@ import { resolveContributionPercent, weightedStake } from '../shared/wagering-we
 
 const ZERO = '0';
 
+/**
+ * What a completing grant is allowed to turn into real money. Absent multiplier means uncapped;
```
klaudia-blazyczek-blurify, line 30: This checks only the grant selected for this wager. If a player holds another active grant in the same currency with a lower maxBet, they can place a larger stake while that bonus is still active. Enforce the strictest maxBet across all active grants in the transaction.

packages/core/src/promo/bonus/seed/index.ts
```diff
@@ -23,11 +23,14 @@ export async function seedDefaultWeightProfile(db: DrizzleDb): Promise<void> {
...
-      scopeRef: null,
-      contributionPercent: '100',
-    })
+    .values([
+      { profileId: profile.id, scope: 'default', scopeRef: null, contributionPercent: '100' },
+      // The headline rule from the specification: casino play clears a wagering requirement,
+      // sportsbook and PvP do not. Seeded rather than left to the default, because resolution
+      // falls through to the default and a bet the operator never weighted would otherwise
+      // count in full.
+      { profileId: profile.id, scope: 'product', scopeRef: 'sportsbook', contributionPercent: '0' },
```
klaudia-blazyczek-blurify, line 32: Existing default profiles never receive these rows. On an environment seeded before this change, the profile insert conflicts and returns before sportsbook and PvP are inserted, so both still use the 100 percent default. Upsert the new rows for the existing default profile without overwriting configured rows.

klaudia-blazyczek-blurify (changes requested):

packages/core/src/promo/bonus/service/wagering.service.ts
zaxovaiko: Hm, leaving this one for now. `maxBet` is checked against the grant the bet is attributed to, which is the single-grant model `lockAttributedGrant` has had since #190. Applying the strictest limit across every active grant only makes sense once a bet counts against all of them, and that is question 1 in the design doc, still open. Happy to revisit once it is answered.

packages/core/src/promo/bonus/seed/index.ts
zaxovaiko: Already handled by `c0ef72cb` in #191, which always upserts the default weight. Rebasing onto it exposed a stale `profile.id` reference in this branch that made the seed throw on boot, fixed in `074a5402`.

zaxovaiko:
- `074a5402` -> `df042f3e`
- `f7e7b10e` -> `05808418`

PR #203, opened 2026-09-20
feat(gaming): per-category game sort catalog with pinning

okapitula:
## Summary

Each game category now carries its own sort mode, and its effective order is materialized by a background job rather than computed per request. Core ships two sorts, `manual` (drag-and-drop) and `name`; operators add their own through the new `GAME_SORT_CATALOG` seam without forking. Games can be pinned to fixed slots that survive re-sorting, and dragging any game switches the category back to manual.

## Why

The public game list was always alphabetical, and nothing stored a per-category order. Operators merchandise categories differently — one wants a hand-curated "Staff picks", another wants slots ordered by revenue over the last 30 days — so the order has to be configurable per category, with room for manual curation on top.

Attribute and statistics sorts (revenue, plays, average bet, RTP, volatility, release date) are deliberately not in core. Each needs a read model or vendor catalog data core does not own, and they are exactly what the seam exists for. An operator ships them as a `GameSortDefinition` in an overlay, with no core change.

The shape that makes that possible: a sort definition answers "what order should these games be in?" and returns ordered ids. It never sees SQL, and it never runs while a player is loading a list.

## Alternatives considered

- **A fixed enum of sort modes** — rejected. Every operator-specific sort would need a core change plus a migration, and core would collect one-off variants that do not generalize. The lobby sections work (#148) turned this down for the same reason.
- **Definitions returning SQL order expressions** — rejected. Core would have to expose joins and a query builder, which limits definitions to data core already knows about, and couples every overlay to this module's table layout. An external ranking service could not be used at all.
- **Resolving the order per request** — rejected. It puts a full-category sort, and possibly an external call, on the public list path. Materializing into a `rank` column keeps the public read at one indexed join no matter what a definition does; the cost is a staleness window, normally under a second.
- **Letting definitions apply pins themselves** — rejected. Pinning would then only work for sorts whose authors implemented it. Core merges pins after a definition returns, so every current and future sort gets it for free.
- **Optimistic concurrency on the write routes** — rejected. An `expectedSortVersion` guard (as lobby layouts and compliance country rules use) was built and then removed: every sibling admin PATCH in this repo is last-write-wins, and a category's ordering does not carry the cost that justifies the guard on those paths.

## Risks

- **Migration** adds columns and two indexes, and drops `game_category_game_category_id_idx`, now redundant beside the new `(category_id, rank)` index. Additive for a previous release running against the new schema; the legacy `game_legacy_*` triggers from `0005` only insert `(game_id, category_id)` and are unaffected. Index builds are not `CONCURRENTLY`, so writes to `game_category_game` block for the build — a consideration for an operator with a large catalog.
- **`GAME_CATALOG_READER.listPlayableGamesInCategory` changes behaviour without changing its type**: it now returns the category's configured order instead of always alphabetical. An overlay that assumed alphabetical needs to know. Nothing in this repo calls it yet.
- **Ordering is eventually consistent.** A write enqueues a re-rank directly and normally lands within a second; `rankDirtyAt` plus a 60s sweep (200 categories per pass, oldest first) is the backstop if that enqueue is lost. Lobby sections add their own layout-cache TTL on top. A category with no completed run yet falls back to name order.
- **Requires a durable job queue in production.** With no worker running, categories keep their last materialized order and new ones stay name-ordered.
- **Event payloads changed.** `gaming.category.created` / `updated` carry the sort snapshot, `games_reordered` gains sort-config before/after (v3), and `pins_updated` is new. Older recorded events still parse via schema defaults.
- **Concurrent admin writes are last-write-wins**, serialized by the category row lock and recorded in the audit trail.
- **Deferred:** attribute and statistics sorts are not covered here — they belong in an overlay on this seam. Draft/Active category status, rule-based membership and homepage eligibility are untouched. The lobby module's own `lobby_category*` tables remain a separate concept.


packages/core/src/casino/gaming/service/game-category.service.ts
```diff
@@ -228,6 +309,289 @@ export class GameCategoryService {
...
+      ? GameSortParamsSchema.safeParse(definitionParams.data)
+      : definitionParams;
+    if (!parsedParams.success) {
+      throw new GameSortConfigInvalidError(`Invalid sortParams for sort '${nextKey}'`);
+    }
+
+    const changed =
+      nextKey !== existing.sortKey ||
+      storedDirection !== existing.sortDirection ||
+      !jsonEqual(parsedParams.data, existing.sortParams ?? {});
```
marek-chmielowski-blurify: `resolveSortPatch`'s change-detection (`jsonEqual`, line 81) uses `JSON.stringify` equality on `sortParams`, which is key-order sensitive. But `existing.sortParams` comes back from a Postgres `jsonb` column, whose key order is normalized on write/read rather than preserved from client insertion order.

Concretely: an admin PATCHes a category once, then resubmits the identical config later (e.g. a settings form that always sends the full object). `input.sortParams` (client key order) vs `existing.sortParams` (jsonb-normalized order) serialize differently even though the values are semantically identical, so `changed` is wrongly `true` — the category gets needlessly re-enqueued for ranking and a `gaming.category.updated` audit event records a no-op as a real config change.

Worth comparing with a structural/deep-equality check instead of stringify.

packages/core/src/casino/gaming/contract/index.ts
```diff
@@ -261,9 +274,126 @@ export const UpdateCategoryInputSchema = z.object({
...
+  id: UuidSchema,
+});
+export type ListCategoryGamesInput = z.infer<typeof ListCategoryGamesInputSchema>;
+
+export const CategoryGamesPageSchema = paginated(CategoryGameItemSchema);
+export type CategoryGamesPage = z.infer<typeof CategoryGamesPageSchema>;
+
+export const ReorderCategoryGamesInputSchema = z.object({
+  id: UuidSchema,
+  gameIds: z
```
marek-chmielowski-blurify, line 326: `ReorderCategoryGamesInputSchema.gameIds` has no `.min(1)`, so `PUT .../games/order` accepts `gameIds: []` and still forces the category into manual sort.

In `reorderCategoryGames` (game-category.service.ts:354-397), the empty-array branch skips the position update for dragged ids but still seeds every existing member's `position` from its current rank and unconditionally sets `sortKey: 'manual'`. So an empty payload (e.g. a drag-end handler firing on an empty/filtered view) silently and permanently drops the category out of its configured automatic sort — no visible symptom until an operator notices ordering has stopped updating.

packages/core/src/casino/gaming/plugin.ts
```diff
@@ -46,16 +81,157 @@ export default {
...
       ));
 
-    ctx.routers.add('gaming', (c) =>
-      createGamingRouter({
+    // svcRefs are null at registration (subscriptions wire before router factories run)
+    // but set before any real event/job arrives. See create-app.ts boot order. A ref
+    // still null when a trigger actually fires is a boot-order bug, not a quiet no-op -
+    // every call site below throws rather than silently skip the rank (mirrors mail's
+    // job handler in mail/plugin.ts).
+    let jobQueueRef: JobQueueAdapter | null = null;
```
marek-chmielowski-blurify: `jobQueueRef`/`drizzleRef`/`rankingRef` start `null` here and are only assigned inside the router factory (further down, where `c.get(JOB_QUEUE)` etc. run). The event handlers and job worker that close over these refs are wired up earlier, so correctness depends on the router factory resolving before any event fires or job runs.

If a future refactor of app boot (or an alternate boot path, e.g. a test harness that subscribes events/workers without invoking every router factory) changes that ordering, `enqueueRank`/`requireDb`/the job handlers will throw `'gaming: ... not resolved yet'` for every rank trigger until this router happens to resolve. Nothing in the type system catches this temporal coupling — worth considering whether these can be resolved lazily or asserted at boot instead.

marek-chmielowski-blurify (commented):
A few findings from reviewing this diff, most significant first.

packages/core/src/casino/gaming/service/game-sort-ranking.service.ts
```diff
@@ -0,0 +1,289 @@
...
+  gameProvider,
+  type GameCategory,
+} from '../schema/index.js';
+import { GAME_CATEGORY_RANK_QUEUE, RANK_SWEEP_BATCH_LIMIT } from '../contract/index.js';
+import { isGamePlayable } from '../../shared/game-catalog.js';
+
+const logger = createLogger('gaming');
+
+/**
+ * Enqueues a `gaming.category.rank` run for `categoryId`. No `idempotencyKey` - a key
```
marek-chmielowski-blurify: This JSDoc on `enqueueGameCategoryRank` documents design rationale ("No idempotencyKey - a key derived only from categoryId would dedupe...") rather than the function's contract. Per this repo's comment convention, reasoning like this belongs in the commit message / PR description / an ADR, not inline — otherwise it rots in place, uncoupled from review history, as the design evolves.

packages/core/src/casino/gaming/service/game-category.service.ts
okapitula: Confirmed, jsonb reorders keys so an identical resubmit counted as a change. Switched to `isDeepStrictEqual`, added a route test for it. Fixed in 42c97e92.

packages/core/src/casino/gaming/contract/index.ts
okapitula: Agreed. Added `.min(1)` and an e2e test (400, sort key unchanged). Fixed in 42c97e92.

packages/core/src/casino/gaming/service/game-sort-ranking.service.ts
okapitula: Dropped the rationale. Kept one line on the BullMQ jobId retention, since without it adding an `idempotencyKey` looks like an easy fix and would break ranking. Fixed in 42c97e92.

zaxovaiko (approved):
Self-review as a listed code owner to satisfy the stacked-PR merge requirement (async merge endpoint doesn't support --admin bypass). All prior review threads were addressed and fixed in 42c97e92; verify and commitlint are green.

